PR #8, opened 2015-11-18
Credo should not be a production dependency

johnhamelink:
Hey there,

I noticed that the readme recommends you install credo without specifying which environment it will run in - to me it seems that credo should only ever be run in development and test environments, and is not particularly useful in production.


rrrene:
:+1:

PR #12, opened 2015-11-19
Add app/ to default config's included files

joekain:
This way, if you add credo to the top level of an umbrella app and have it run
over all source files.


rrrene:
Thx Joseph, very good catch! Will release a new version later today!

joekain:
Thanks @rrrene!

rrrene:
It's in `v0.1.8`. :shipit:

PR #16, opened 2015-11-23
Problem: Compilation errors on Elixir 1.2.0-dev

c-rack:
Latest Elixir master introduces additional compilation warnings.
This patch fixes some of them:

---

lib/credo/check/design/duplicated_code.ex:122:
warning: you are piping into a function call without parentheses, which is ambiguous. Please wrap the function you are piping into in parentheses. For example:

    foo 1 |> bar 2 |> baz 3

Should be written as:

    foo(1) |> bar(2) |> baz(3)

---

lib/credo/check/readability/trailing_white_space.ex:20:
warning: an expression is always required on the right side of ->. Please provide a value after ->

---

lib/credo/cli/output/issues_short_list.ex:15:
warning: an expression is always required on the right side of ->. Please provide a value after ->

---

lib/credo/config.ex:87:
warning: default arguments in from_exs/3 are never used

### DIFF
--- a/lib/credo/check/design/duplicated_code.ex
+++ b/lib/credo/check/design/duplicated_code.ex
@@ -119,7 +119,7 @@ defmodule Credo.Check.Design.DuplicatedCode do
       first_node
       |> hashes(%{}, filename)
       |> Map.keys
-      |> List.delete my_hash # don't count self
+      |> List.delete(my_hash) # don't count self
 
     subhashes
   end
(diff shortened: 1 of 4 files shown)


rrrene:
Nice. Thank you! :+1:

c-rack:
@rrrene You are welcome!

Actually, there are still some more warnings, like the following one. Will work on these shortly.
```
lib/credo/cli/output/explain.ex:144:
warning: the variable "line" is unsafe as it has been defined in a conditional clause,
as part of a case/cond/receive/if/&&/||.
Please rewrite the clauses so the value is explicitly returned. For example:

    case int do
      1 -> atom = :one
      2 -> atom = :two
    end

Can be rewritten as:

    atom =
      case int do
        1 -> :one
        2 -> :two
      end
```

rrrene:
I read a comment by @josevalim about this behaviour and very much welcome the changes (meaning I will gladly accept any PR fixing warnings for 1.2)!

PR #17, opened 2015-11-23
Solution: move the issue.column if-block inside the issue.line_no if-block

c-rack:
Fixes Elixir 1.2 compilation warning.


rrrene:
:+1:

PR #18, opened 2015-11-23
Problem: Credo.Consistency.HelperTest was not executed

c-rack:
Running `mix test` produces the warning:

```
test/credo/check/consistency/helper_test.ex does not match "*_test.exs" and won't be loaded
```

Solution:

1. Rename `helper_test.ex` to `helper_test.exs`
2. Rename test to `Credo.Check.Consistency.HelperTest`
3. Fix alias to `Credo.Check.Consistency.Helper`
4. Fix asserts to reflect changed return format of `most_picked_prop_value/1`


rrrene:
:+1:

PR #21, opened 2015-12-03
Add note that predicates should not start with 'is'

sorentwo:
The message didn't fully explain what would flag a function as having poor predicate naming.


sorentwo:
There was a typo in the commit message, that has been fixed.

rrrene:
👍 thank you!

PR #25, opened 2015-12-20
Describe default behaviour for function definitions and strings.

c-rack:
Also, fix minor typo.


rrrene:
As always: Thank you sir! :neckbeard:

PR #31, opened 2016-01-04
Fox minor clause warning for a missing value

sorentwo:
Inserts an explicit `nil` value instead of an implicit `nil` value where a catch-all clause matches and only has a comment in place.


rrrene:
thx! :+1:

PR #33, opened 2016-01-23
Check for config files in the config directories at each path node

barruumrex:


barruumrex:
I've got an umbrella project where I wanted to apply the same style throughout. I put the config in the config directory at the root of the project. Made it through most of the implementation to add that directory before I realized I could just have moved the file out to root and everything would have worked. 

Filed the pull request because I'd still prefer to have the file in the config folder instead of cluttering up the root directory.

rrrene:
Very cool! Added a test for good measure ;-)

Since `v0.3.0` won't be out for a couple of weeks, I released `v0.3.0-dev2` so you can use it today: https://hex.pm/packages/credo/0.3.0-dev2

Thanks again :+1:

PR #37, opened 2016-02-03
Do not use `:white` as the color for filenames.

lucasmazza:
The fixed `:white` color makes credo a bit unhelpful when using light backgrounds:

. 

Using the `:default_color` leaves this up for your terminal to decide:


rrrene:
Great! Thanks for the heads up and PR! :+1:

PR #87, opened 2016-05-28
Introduce BoolOperationOnSameValues check

vladimir-tikhonov:
It's very similar to the `OperationOnSameValues` check, but `OperationOnSameValues` is about constant results, so I decided to create a separate check.

Not sure if I should use release-0.4.0 branch. Also my english and elixir skills are quite poor, so any feedback is highly appreciated :)


rrrene:
Looks good! :+1: Thanks for contributing, I appreciate it!

Choosing `release-0.4.0` was exactly right! :tada: :shipit:

PR #88, opened 2016-05-28
Ignore metadata in OperationOnSameValues and BoolOperationOnSameValue checks

vladimir-tikhonov:
I've noticed that `OperationOnSameValues` and `BoolOperationOnSameValue` checks don't work when `lhs` and `rhs` are on different lines and `metadata` values inside ast are different.

Not sure if `CodeHelper` is good place for `remove_metadata`. Feedback is highly appreciated!

### DIFF
--- a/lib/credo/check/code_helper.ex
+++ b/lib/credo/check/code_helper.ex
@@ -122,4 +122,48 @@ defmodule Credo.Check.CodeHelper do
     |> Strings.replace_with_spaces
   end
 
+
+  @doc """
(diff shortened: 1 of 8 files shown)


rrrene:
Looks good to me. :+1:

Could you add a test to both checks where the `lhs` and `rhs` are on different lines? Just to document this improvement and ensure that we don't break it in the future ...

vladimir-tikhonov:
@rrrene done

rrrene:
Perfect.

PR #90, opened 2016-05-29
Minor tweaks to do with summing

mikestok:
See individual commit messages, the pattern of

    Enum.reduce(0, &(&1+&2))

can usually be better expressed in my opinion.


rrrene:
love it 👍🏻

PR #91, opened 2016-05-30
Minor clean-ups on credo

mikestok:
This came about because I ran `credo` on itself as an experiment, and these seemed like harmless changes to eliminate a whole class of messages.


rrrene:
This looks really good! Thanks! :+1:

PR #92, opened 2016-05-31
allow module names when parsing config in safe mode

alemata:
Hi @rrrene, I was trying to load a config in safe mode, using default credo config and it was not working because of module names. 

What do you think about this change?


rrrene:
Looks great. Could you check if we can use `Module.safe_concat` in there? Seems "even safer" :wink: but I'm not sure if it works in this case without modifications ...

alemata:
@rrrene done! Thanks!

PR #94, opened 2016-05-31
V1.3 compile warning

barruumrex:
First pass at cleaning up Elixir v1.3.0-rc.0 compile warning vomit.


rrrene:
This seems like a really nice PR :+1: 

One thing: please rebase on `release-0.4.0` and target it there, as that is were the current release is build. (I will change this mode of operation for future releases as it is non-obvious for contributors, but for now we have to deal with it).

barruumrex:
Sure thing. Would you like me to squish the commits down to a single commit while I'm at it? I was committing as I went before just to keep the noise down, but there's nothing stopping a squish at this point. 

Do you have an feedback on any pieces of the implementation/testing? It compiles and passes all existing tests, but I didn't do a review of the coverage around these files. I did notice that cli.ex doesn't appear to be directly tested and I did a fairly major refactoring of option parsing in there, which has me a little worried.

rrrene:
Don't worry. The important bit is the rebasing onto the right branch. And you don't need to squish them, because this way I would be able to just revert a single refactoring while keeping the rest :shipit:

barruumrex:
Closing pull request an opening a new one against the correct branch.

barruumrex:
New pull request is https://github.com/rrrene/credo/pull/96

PR #106, opened 2016-06-05
Release 0.4.0 - reduce number of messages from `mix credo --strict` on itself

mikestok:
This contains changes which I think are uncontroversial to reduce the number of "violations" in the `credo` source itself.

### DIFF
--- a/lib/credo/check/refactor/abc_size.ex
+++ b/lib/credo/check/refactor/abc_size.ex
@@ -40,7 +40,7 @@ defmodule Credo.Check.Refactor.ABCSize do
   end
   for op <- @def_ops do
     defp traverse({unquote(op), meta, arguments} = ast, issues, issue_meta, max_abc_size) when is_list(arguments) do
-      abc_size = abc_size_for(ast) |> round
+      abc_size = ast |> abc_size_for |> round
       if abc_size > max_abc_size do
         fun_name = CodeHelper.def_name(ast)
         {ast, [issue_for(issue_meta, meta[:line], fun_name, max_abc_size, abc_size) | issues]}
--- a/lib/credo/cli/output/ui.ex
+++ b/lib/credo/cli/output/ui.ex
@@ -11,8 +11,15 @@ defmodule Credo.CLI.Output.UI do
   defdelegate puts(v), to: Bunt
   def puts(v, color), do: Bunt.puts([color, v])
 
+  def puts_edge(color, indent \\ 2) when is_integer(indent) do
+    [color, indent]
+    |> edge
+    |> puts
+  end
+
   def wrap_at(text, number) do
-    Regex.compile!("(?:((?>.{1,#{number}}(?:(?<=[^\\S\\r\\n])[^\\S\\r\\n]?|(?=\\r?\\n)|$|[^\\S\\r\\n]))|.{1,#{number}})(?:\\r?\\n)?|(?:\\r?\\n|$))")
+    "(?:((?>.{1,#{number}}(?:(?<=[^\\S\\r\\n])[^\\S\\r\\n]?|(?=\\r?\\n)|$|[^\\S\\r\\n]))|.{1,#{number}})(?:\\r?\\n)?|(?:\\r?\\n|$))"
+    |> Regex.compile!
     |> Regex.scan(text)
     |> Enum.map(&List.first/1)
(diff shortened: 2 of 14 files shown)


lib/credo/sources.ex
```diff
@@ -64,7 +64,8 @@ defmodule Credo.Sources do
     |> Enum.any?(&matches?(file, &1))
   end
   defp matches?(file, string) when is_binary(string) do
-    find(string)
+    string
+    |> find
```
rrrene: I think we have to add parens to this function call to avoid warnings.

lib/credo/cli/output/ui.ex
```diff
@@ -11,8 +11,15 @@ defmodule Credo.CLI.Output.UI do
   defdelegate puts(v), to: Bunt
   def puts(v, color), do: Bunt.puts([color, v])
 
+  def puts_edge(color, indent \\ 2) when is_integer(indent) do
+    [color, indent]
+    |> edge
+    |> puts
```
rrrene: I think we have to add parens to this function calls to avoid warnings.

lib/credo/check/refactor/abc_size.ex
```diff
@@ -40,7 +40,7 @@ defmodule Credo.Check.Refactor.ABCSize do
   end
   for op <- @def_ops do
     defp traverse({unquote(op), meta, arguments} = ast, issues, issue_meta, max_abc_size) when is_list(arguments) do
-      abc_size = abc_size_for(ast) |> round
+      abc_size = ast |> abc_size_for |> round
```
rrrene: I think we have to add parens to this function call `abc_size_for` to avoid warnings.

rrrene:
Great! I added some comments. Thoughts?

mikestok:
I don't see any warnings for missing parens during checking with credo (`mix credo --strict`) or any warnings during compilation (`mix clean; mix compile`).

Looking at discussions like https://github.com/niftyn8/elixir_style_guide/issues/50 I *think* that the parens only matter if we're passing more arguments than the "magic" first argument put in by `|>`.  For example

    foo |> bar |> baz # OK
    foo |> bar "x" |> baz # Ambiguous
    foo |> bar("x") |> bad # OK

Checking in `iex`:

```
iex(1)> defmodule Test do
...(1)> def foo, do: "foo output"
...(1)> def bar(_), do: "bar/1 output"
...(1)> def bar(_, _), do: "bar/2 output"
...(1)> def baz(s), do: s
...(1)> end
{:module, Test,
 <<70, 79, 82, 49, 0, 0, 6, 56, 66, 69, 65, 77, 69, 120, 68, 99, 0, 0, 1, 24, 131, 104, 2, 100, 0, 14, 101, 108, 105, 120, 105, 114, 95, 100, 111, 99, 115, 95, 118, 49, 108, 0, 0, 0, 4, 104, 2, ...>>,
 {:baz, 1}}
iex(2)> Test.foo |> Test.bar |> Test.baz
"bar/1 output"
iex(3)> Test.foo |> Test.bar "x" |> Test.baz
iex:3: warning: you are piping into a function call without parentheses, which may be ambiguous. Please wrap the function you are piping into in parentheses. For example:

    foo 1 |> bar 2 |> baz 3

Should be written as:

    foo(1) |> bar(2) |> baz(3)

"bar/2 output"
iex(4)> Test.foo |> Test.bar("x") |> Test.baz
"bar/2 output"
```

So I don't *think* we need the parens in the cases.

rrrene:
Mmmh. I could swear this was one of the features of 1.2 - it even says in the [release notes](https://github.com/elixir-lang/elixir/releases/tag/v1.2.0) 

> [Kernel] Warn if piping into an expression without parentheses

@josevalim Is the following code not supposed to trigger a warning?

    defmodule X do
      def run(y) do
        y |> convert
      end

      def convert(z) do
        # ...
      end
    end

I would have expected a warning saying something akin to "use parentheses when piping into a local function"?

josevalim:
It doesn't warn if you have no arguments.

rrrene:
@josevalim ah, so this would?
    defmodule X do
      def run(a, b) do
        a |> convert b
      end

      def convert(foo, bar) do
        # ...
      end
    end

because `convert` takes a second argument, which should be put in parens?

PR #109, opened 2016-06-05
Release 0.4.0 - cleanup lib/credo/priority.ex

mikestok:
I'm not sure what should go in `@moduledoc` but this eliminates all the other warnings for `priority.ex`.

I was puzzled by the long lines which snuck past `credo` mentioned in the commit message for eb49a29, but as my goal is to eliminate warning messages I'll not dig into that *now* 😀


rrrene:
Thank you! :+1:

rrrene:
btw: The "not reporting an issue" thing is due to a parameter in `MaxLineLength` which allows you to exclude function headers (and it is `true` by default).

PR #119, opened 2016-06-19
Checks for parentheses in if/unless conditions

### LINKED ISSUE #78: Doesn't catch parentheses around a condition, which seems to hide other issues
Saw the following during a code review and was surprised that Credo found no issues:

    if (!some_condition) do
      # ...
    else
      # ...
    end

Given the code above, I would expect:
1. An issue to be raised about parentheses around the condition;
2. "If/else blocks should not have a negated condition in `if`" to be raised.

Actual results were that Credo reported no issues.

If it matters, I'm using Credo 0.4.0-beta1 with Elixir 1.2.3.

stevelove:
Adds a check for parentheses in `if` or `unless` conditions.
Moves existing `Credo.Check.Refactor.ParenthesesInConditionTest` to `Credo.Check.Readability`.
Fixes #78

### DIFF
--- a/lib/credo/check/readability/parentheses_in_condition.ex
+++ b/lib/credo/check/readability/parentheses_in_condition.ex
@@ -1,0 +1,82 @@
+defmodule Credo.Check.Readability.ParenthesesInCondition do
+  @moduledoc """
+  Because `if` and `unless` are macros, the preferred style is to not use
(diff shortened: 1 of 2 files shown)


rrrene:
Looks good to me! :+1: will merge soon!

PR #121, opened 2016-06-24
Add `ignore_specs` option to max_line_length

### LINKED ISSUE #120: Possibly ignore @spec on line length checks
There's currently on option to ignore function definitions when checking line length, and I think `@spec` kind of goes along with a function definition, and there should maybe be an option to ignore those as well. That might be a terrible practice though, I have no idea.

As a side note, I noticed there's also an option to ignore line length for strings, though that hasn't been implemented yet. Are there plans to add it in any time soon?

sgtpepper43:
Add option to ignore `@spec` definitions when getting line length
errors.

Fix #120

### DIFF
--- a/lib/credo/check/readability/max_line_length.ex
+++ b/lib/credo/check/readability/max_line_length.ex
@@ -10,11 +10,13 @@ defmodule Credo.Check.Readability.MaxLineLength do
     params: [
       max_length: "The maximum number of characters a line may consist of.",
       ignore_definitions: "Set to `true` to ignore lines including function definitions.",
+      ignore_specs: "Set to `true` to ignore lines including `@spec`s.",
     ]
   ]
   @default_params [
     max_length: 80,
     ignore_definitions: true,
+    ignore_specs: false,
     ignore_strings: true
   ]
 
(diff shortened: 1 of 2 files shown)


sgtpepper43:
The current default is `false`, since I didn't want to break anyone's current linting options, and I don't know what everyone wants it to be going forward. Though if it was true, it wouldn't cause any lint errors, since it's just relaxes a current rule.

rrrene:
Great! Thx for implementing this :+1:

One edge case that is covered here is if the `@spec` continues in the next line and breaks the line length limit there:

``` 
    @spec some_fun(binary, binary, 
      binary, binary, binary, binary, binary, binary, binary, binary, binary, binary)
    def some_fun(arg0, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8) do
      assert 1 + 1 == 2
    end
```

But I'm unsure how we would cover this ... Ideas?

sgtpepper43:
It looks like the same thing happens for really long multi-line `def`'s too, but I'm thinking that if you're going to do multi-line anyway, you should just follow the line length limits at that point. In terms of readability, it makes sense to let a `@spec` or `def` be like 100, 120 characters long so you can keep it on one line, but once you get so long that you need to break it up, it makes sense to keep each line to 80, I think.

rrrene:
Absolutely. My comment wasn't a critique of your PR, just thinking out loud. Will merge tomorrow and make a release!

sgtpepper43:
Awesome! I figured you were just thinking through that edge case, I just have no idea how to even handle it, so figured I'd just try to convince you to ignore it :wink:

PR #130, opened 2016-07-17
add map assignment feature

slavaZim:


rrrene:
Wow, this looks great! :+1:

Let me look over the code and add some comments :)

lib/credo/check/consistency/pattern_matching_assignment.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Credo.Check.Consistency.PatternMatchingAssignment do
+  @moduledoc """
+  Both left and right pattern matching assigments of Map arguments on function call are valid -
```
rrrene: We should probably get this to match not only maps, but any kind of argument.

I.e. the following should be caught as inconsistent:

    def my_fun(:signup = action, user = %User{}) do

lib/credo/check/consistency/pattern_matching_assignment/left_side_assignment.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Credo.Check.Consistency.PatternMatchingAssignment.LeftSideAssignment do
+  use Credo.Check.CodePattern
+
+  alias Credo.Check.CodeHelper
+  alias Credo.Code
+
+  @regex_def ~r/def .+?\((.+?)\),? do:?/
+  @regex_map ~r/%{.+}\s*=\s*\S+/
```
rrrene: I am pretty sure we don't want to use a regex to achieve this, because it is less reliable than using `Credo.Code.traverse/2`.

rrrene:
The test suite looks good. Please put all module attributes at the top of the test and include these test cases to ensure corner cases are checked as well: 

    def my_fun(:signup = action, user = %User{}) do 
    end

    def my_fun(user = %User{}, %Project{active: true} = project) do 
    end

    def my_fun(user = %User{id: user_id}, %Project{active: true, 
                                                user_id: user_id} = project) do 
    end

    def my_fun(user = %User{id: user_id}, %Project{active: true, 
                                            # only match on projects for `user`
                                            user_id: user_id} = project) do 
    end

Thanks for working on this!

lib/credo/check/consistency/pattern_matching_assignment.ex
slavaZim: Hi. `def my_fun(:signup = action)` This one confuses me. Why do I assign variable to value I already know?

rrrene: To use it as a variable. For example in Phoenix you might do

    def show(conn, %{"id" => id} = params) do

because you need to pattern match `id` but also want the ability to look upo other things in `params`.

But there are also situations where a name is easier to grasp than a simple value (like in my example). Say you have a function that takes a number and then that number is used again in the functions body. If the number has semantic meaning it might be helpful to give it a name:

    def my_fun(10_000 = user_limit) do
      "Limit reached: #{user_limit} (users)"
    end

This can also be used to make meta-programming less confusing because you don't have to unquote as much:

    for value <- [1, 2, 3, 4] do
      defp some_fun(unquote(value), lhs, rhs) do
        if some_fun(lhs, rhs, unquote(value)) == some_fun(lhs, unquote(value), rhs) do
          {unquote(value), lhs}
        else
          {rhs, unquote(value)}
        end
      end
    end

    for v <- [1, 2, 3, 4] do
      defp some_fun(unquote(v) = value, lhs, rhs) do
        if some_fun(lhs, rhs, value) == some_fun(lhs, value, rhs) do
          {value, lhs}
        else
          {rhs, value}
        end
      end
    end

These examples might seem a bit contrieved (sorry, I'm out of coffee), but I think it illustrates the point that there might be some use in having this covered by the tests.

slavaZim: ok, got you

slavaZim:
@rrrene, hi, I updated code check to use `Credo.Code.traverse/2 ` instead of string regex filter as you suggested (also it now should be triggered by structs and non-hash values assignment), however travis check fails with  strange `(UndefinedFunctionError) undefined function Credo.Code.traverse/2` which is not happening locally on my machine. Also, I see

rrrene:
@slavaZim Credo has changed in the meantime and that's why Travis fails.

You will have to rebase your patch branch against the latest version of my master branch ([here is a StackOverflow answer](http://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository), just in case you've never done it).

devonestes:
@slavaZim Are you still working on this PR? It looks like it's very nearly there! I'd be happy to help get it over the finish line if you want - just shout if you need help. Otherwise, I think someone might take this branch over and finish it up so this check can get merged in.

slavaZim:
@devonestes,  hi, I'm not sure why it is not reviewed, may be it doesn't meet some expectations of @rrrene, but you are of cource wellcome if you want to help

devonestes:
@slavaZim The last CI run is like 4 months old, unfortunately, which at this point might be a hold up. Rebasing and re-pushing it up so we get a new CI run on top of the current `master` branch might be a good first step - that way we know it's still going to work with all the changes that have been merged in over the last four months.

rrrene:
@slavaZim So sorry I did fail to review your changes! The problem was that we included `Credo.Check.Consistency.ParameterPatternMatching` which does something very similar to what your feature seems to be doing.

If you want to check and maybe point out the differences, you can [check the source here](https://github.com/rrrene/credo/blob/1adc8082cbbdaf9b2c7dd23aefdeba13cb55e4bf/lib/credo/check/consistency/parameter_pattern_matching.ex).

Again sorry for missing this! 😞

PR #132, opened 2016-07-22
Improve source file loading

britto:
Making use of [`Enum.partition/2`](http://elixir-lang.org/docs/stable/elixir/Enum.html#partition/2) we avoid looping twice over the list of source files when filtering out invalid source files. Apart from improved legibility, in a sufficiently large list these changes may even mean a perceptible performance improvement.


rrrene:
:+1: Thank you, sir!

rrrene:
`v0.4.7` contains the fix. Thx again!

PR #147, opened 2016-08-08
Prevent Credo from processing the same file multiple times

britto:
Currently, a minor misconfiguration in the list of source files to be checked can lead to unwanted behavior. Intersecting glob patterns will make the engine process files **multiple times**, which in turn will trigger some issues of the `Credo.Check.Design.DuplicatedCode` type.

### DIFF
--- a/lib/credo/sources.ex
+++ b/lib/credo/sources.ex
@@ -23,12 +23,14 @@ defmodule Credo.Sources do
   def find(%Credo.Config{files: files}) do
     files.included
     |> Enum.flat_map(&find/1)
+    |> Enum.uniq
     |> exclude(files.excluded)
     |> to_source_files
   end
   def find(paths) when is_list(paths) do
     paths
     |> Enum.flat_map(&find/1)
+    |> Enum.uniq
   end
   def find(path) when is_binary(path) do
     path
--- a/lib/credo/sources.ex
+++ b/lib/credo/sources.ex
@@ -5,13 +5,15 @@ defmodule Credo.Sources do
   @stdin_filename "stdin"
 
   @doc """
-  Finds sources for a given Config, List or String.
+  Finds sources for a given `Credo.Config`.
 
-  iex> Sources.find(%Credo.Config{files: %{excluded: [], included: ["*.ex"]}})
+  Through the `files` key, configs may contain a list of `included` and `excluded`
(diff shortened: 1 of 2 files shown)


lib/credo/sources.ex
```diff
@@ -23,12 +23,14 @@ defmodule Credo.Sources do
   def find(%Credo.Config{files: files}) do
     files.included
     |> Enum.flat_map(&find/1)
+    |> Enum.uniq
     |> exclude(files.excluded)
     |> to_source_files
   end
   def find(paths) when is_list(paths) do
```
britto: I am not sure if this function should be actually public. IMO it works as a private helper for the top level `find(%Credo.Config)` function, which is the real public API we expect to use. The same applies to the other fuction below, `find(path) when is_binary(path)`. Even the return types are different:

```elixir
Credo.Sources.find(%Credo.Config{files: %{excluded: [], included: ["file.ex"]}})
# => %Credo.SourceFile{filename: "file.ex"}

Credo.Sources.find("file.ex")
# => ["file.ex"]
```

So I'd say it makes sense to `defp` them. WDYT?

rrrene: We can't `defp` them because we would lose the ability to test them. You are using tests on these functions yourself to ensure the results are uniq.

You might have a point however about renaming those two functions (to something like `find_via_glob/1`) since they have different returns.

If you would want to do that and patch the tests accordingly, I'd be fine with that!

britto: 👍 That was my point.

If we don't mean to expose this API of `find` with a _binary_ or a _list_ to the public, we should definitely make them private, then we only test the composed behavior of the public part of the `find` algorithm.

Ensuring that `find/1 when is_list` and `find/1 when is_binary` return unique results would be worhtless if the "parent" `find(%Credo.Config)` function - the one we intend to use - would not do the same. From the Credo Engine point of view what really matters is the output of `find(%Credo.Config)`, right? so, at the end of the day we only really need a single final call to `uniq`.

britto:
@rrrene I've added some tests to cover all the tricky edge cases, then worked a little bit on improving the code of `Credo.Sources` f265731. Feedback is welcome 😃.

rrrene:
Thx a lot!!

Issue #20, opened 2015-11-26
Warn when module contains function of same name but different arity of behaviour callback


Havvy:
Example:

```
defmodule Example
  use GenServer

  def handle_call(:x, state) do
    # Notice the missing "from" parameter
    {:reply, :y, state}
  end
end
```

Should warn on Example.handle_call/2 being similar to Example.handle_call/3 since the `@behaviour` added via `use GenServer` has a `@callback handle_call/3`.

rrrene:
It's not as trivial as the checks we have today, but would be *very* useful. We will have to enhance Credo's internal tooling a bit to make this happen, but this is a really great idea! :+1:

rrrene:
Closing this because the results from #148 strongly suggest that we won't be able to do this without compilation.

We will be able to provide functionality like as soon as we provide plugins (b/c those will be free to support checks which require prior compilation) :+1:

PR #148, opened 2016-08-11
issue/20: add new check CallbacksArity

DavydenkovM:
Hello, René!

This is a PR related to the issue #20 about checking arities of the callback functions.

A brief changelog:
- new check called `Credo.Check.Warning.CallbacksArity` implemented (with some initial tests)
- new functions `behaviours` and `callbacks` created in `module.ex` to fetch behaviours/callbacks for a given ast/module
- a few new `test_helper.ex` functions provided

There are some issues to discuss i think. Probably someone knows a better way of getting required callbacks or behaviours from ast/module. Please don't hesitate to correct my grammar, semantics or logical mistakes, I would be happy to fix it. 

Also please find my code-related comments attached. 

Thanks in advance for your time!

### DIFF
--- a/lib/credo/cli/command/explain.ex
+++ b/lib/credo/cli/command/explain.ex
@@ -7,7 +7,6 @@ defmodule Credo.CLI.Command.Explain do
   alias Credo.Config
   alias Credo.CLI.Output.Explain
   alias Credo.CLI.Output.UI
-  alias Credo.CLI.Output
   alias Credo.Sources
 
   # TODO: explain used config options
@@ -15,7 +14,7 @@ defmodule Credo.CLI.Command.Explain do
   def run(_args, %Config{help: true}), do: print_help()
   def run([], _), do: print_help()
   def run([file | _], config) do
-    {_, source_files} = load_and_validate_source_files(config)
+    {_, source_files} = Sources.load_and_validate_source_files(config)
     {_, {source_files, config}}  = run_checks(source_files, config)
 
     file
@@ -25,20 +24,6 @@ defmodule Credo.CLI.Command.Explain do
     # TODO: return :error if there are issues so the CLI can exit with a status
     #       code other than zero
     :ok
-  end
-
-  defp load_and_validate_source_files(config) do
-    {time_load, {valid_source_files, invalid_source_files}} =
-      :timer.tc fn ->
-        config
-        |> Sources.find
-        |> Enum.partition(&(&1.valid?))
-      end
-
-    invalid_source_files
-    |> Output.complain_about_invalid_source_files
-
-    {time_load, valid_source_files}
   end
 
   defp run_checks(source_files, config) do
--- a/lib/credo/code/module.ex
+++ b/lib/credo/code/module.ex
@@ -119,7 +119,7 @@ defmodule Credo.Code.Module do
       []
     end
   end
-  def behaviour(_), do: []
+  def behaviours(_), do: []
 
   @doc "Returns the keyword list of callbacks defined in a given module"
   def callbacks(behaviour) when is_atom(behaviour) do
--- a/lib/credo/code/module.ex
+++ b/lib/credo/code/module.ex
@@ -123,15 +123,10 @@ defmodule Credo.Code.Module do
 
   @doc "Returns the keyword list of callbacks defined in a given module"
   def callbacks(behaviour) when is_atom(behaviour) do
-
-    if Code.ensure_loaded?(behaviour) do
       [callbacks_from_module_attributes(behaviour) |
        callbacks_from_source_file(behaviour)]
       |> List.flatten
       |> Enum.uniq
-    else
-      []
-    end
   end
   def callbacks(_), do: []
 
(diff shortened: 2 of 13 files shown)


lib/credo/cli/command/explain.ex
```diff
@@ -7,15 +7,14 @@ defmodule Credo.CLI.Command.Explain do
...
   alias Credo.CLI.Output.UI
-  alias Credo.CLI.Output
   alias Credo.Sources
 
   # TODO: explain used config options
 
   def run(_args, %Config{help: true}), do: print_help()
   def run([], _), do: print_help()
   def run([file | _], config) do
-    {_, source_files} = load_and_validate_source_files(config)
```
DavydenkovM: This function was used without any changes in 3 different modules. I have moved it into `sources.ex`. Please let me know if there are a better home for this function

lib/credo/code/module.ex
```diff
@@ -4,7 +4,6 @@ defmodule Credo.Code.Module do
   funcions or module attributes.
   """
 
-  alias Credo.Code
```
DavydenkovM: I have removed this alias, because It is conflicts with original module `Code` (which is used in new functions)

lib/credo/code/module.ex
```diff
@@ -94,21 +93,79 @@ defmodule Credo.Code.Module do
...
+    if Code.ensure_loaded?(module) do
+      module.module_info[:attributes]
+      |> Keyword.get_values(:behaviour)
+      |> List.flatten
+      |> Enum.map(&camelize_behaviour/1)
+    else
+      []
+    end
+  end
+  def behaviours(_), do: []
```
DavydenkovM: Here I am using module_info to get the list of behaviours. 
This requires `module` to be loaded. That is why helpers `ensure_loaded` provided in tests.

I did't find a way of getting behaviours / callbacks using ordinary `ast traversal`. I have tried `Macro.expand`, but this gives me nothing.

DavydenkovM:
It seems that all checks are failing now because of specific compilation options on travis (there is no tuples in `module.module_info[:compile]`  - http://erlang.org/doc/reference_manual/modules.html#id77056). I will try to figure out what can be done to resolve this problem (all tests are passing locally since I am using `kerl` and `kiex`).

DavydenkovM:
Hmm, I have found that there are some strange issues with `travis CI` filenames.

I have done a few experiments with code instrumentation in my local branch (branch - https://github.com/DavydenkovM/credo/pull/2)

Here is the instrumented code sample (`Credo.Sources#find when is_atom(module)`):

```
  def find(module) when is_atom(module) do
    IO.inspect module.module_info[:compile]

    path =
    module.module_info[:compile][:source]
    |> to_string

    IO.inspect path

    file = %Credo.Config{files: %{excluded: [], included: [path]}}
    |> find
    |> List.first

    IO.inspect file
  end
```

And here are the results of the instrumented `travis-ci build`:

1) when test passes - https://travis-ci.org/DavydenkovM/credo/jobs/151817791#L495-L513
2) when test fails - https://travis-ci.org/DavydenkovM/credo/jobs/151817791#L405-L408

As you can see when test passing source path looks fine - 
`"/home/travis/build/DavydenkovM/credo/test/support/callbacks/test_behaviour.exs"`

And when test fails source path looks like this - `"/home/ubuntu/bob/tmp/715944860da4d7bb5af7006bd8198704/elixir/lib/elixir/lib/gen_server.ex"`
(pointing to strange home directory)

For some reasons

```
  defp to_source_file(filename) do
    filename
    |> File.read!
    |> SourceFile.parse(filename)
  end
```

returns nothing in the second case (probably permissions issue). 

How can we resolve aforementioned situation, what do you think?

rrrene:
Hi Mikhail,

thx for working on this. The tests and public API you designed for this are really good!

Unfortunately, I can't accept this in its current state. We can not use things like `module.module_info[:compile][:source]` in Credo, because Credo is supposed to be a static analysis tool that works without compiling the source code which it analyses.

I know that this makes analysing things like "behaviours applied to this module" much harder. Sorry about this :disappointed: 

If it is an option to you, you could submit your tests in a separate pull request. They seem quite good (just remove the `ensure_loaded` and other compilation related stuff) and can be a first step in the right direction. :+1:

DavydenkovM:
Thanks for your reply, René!

I will remove all the compilation-related stuff and will think a bit more how to resolve this issue without source compilation (if there are any way).
My pleasure to work on this project. This PR was just a proposal and I hope we can find solution next time!

DavydenkovM:
I have inspected this code once more and discuss this issue on forum -
https://elixirforum.com/t/fetch-all-callbacks-from-behaviour-module/1372

For now I have no vision how to resolve this issue and how to test it without compilation. René, please let me know if you have any suggestions, I would try to resolve it again using your algorithm.

Otherwise please feel free to close this PR. All the tests we discussed are related to the check `CallbacksArity`, so there is no useful things to be submitted without this check. 

Sorry about all inconveniences, it was an interesting experience anyway :)

PR #154, opened 2016-08-21
Don't enforce @moduledoc requirement for .Mixfile or Test modules.

scouten:


rrrene:
Good catch! :+1: Thank you!

scouten:
Thanks for the quick merge! :-)

PR #156, opened 2016-08-22
Checks no parentheses exists on zero arity functions

rodrigues:
This will trigger a warning on function heads that have no arguments, but do have parentheses.

Interesting to have that in credo? 😊


rrrene:
Thanks so much :+1:

PR #167, opened 2016-09-10
Stop moduledoc check for child nodes in ignored modules

### LINKED ISSUE #165: Don't enforce @moduledoc requirement modules declared within Test modules
The Executable.ControllerTest.ExampleCommand module is for testing only and shouldn't need a moduledoc

```elixir
defmodule Executable.ControllerTest do
  use ExUnit.Case, async: true

  defmodule ExampleCommand do
  end
end
```

ottolin:
Adding accumulator to keep track if we should continue the ast prewalk when an ignored module is found.
fix #165


rrrene:
Looks good. Will merge soon. :+1:

rrrene:
Just released `v0.4.12` which includes your contribution! :+1:

PR #178, opened 2016-09-29
Don't care about excluded files when reading stdin

deadtrickster:
"stdin" annotation can't be overridden since 0.3.13 with default config.

Today I spent quite some time trying to figure out why flycheck-credo doesn't work.
As it turned out this https://github.com/rrrene/credo/blob/master/.credo.exs#L23 doesn't work with this https://github.com/rrrene/credo/blob/master/lib/credo/sources.ex#L19


rrrene:
thx!

rrrene:
Just released `v0.4.12` which includes your contribution! :+1:

PR #179, opened 2016-10-02
Fix false positive with grouped logic in if statement

### LINKED ISSUE #164: False positive with grouped logic in if statement
When using parentheses to group logic in an if statement, I get a false positive regarding using parentheses.

```
echo "if (thing && other_thing) || better_thing, do: something" | mix credo --read-from-stdin
Checking 1 source file ...

  Code Readability                                                              
┃ 
┃ [R] ↗ The condition of `` should not be wrapped in parentheses.
┃       stdin:1 ()
```

EricWorkman:
Fixes #164 

I added some more tests to account for grouping on either side of the outer operation and for one lined statements.

This keeps track of the condition that had the grouped logic. 

Before 
```
echo "if (thing && other_thing), do: something" | mix credo --read-from-stdin
Checking 1 source file ...

  Code Readability                                                              
┃ 
┃ [R] ↗ The condition of `` should not be wrapped in parentheses.
┃       stdin:1 ()
```

After
```
echo "if (thing && other_thing), do: something" | mix credo --read-from-stdin
Checking 1 source file ...

  Code Readability                                                              
┃ 
┃ [R] ↗ The condition of `if` should not be wrapped in parentheses.
┃       stdin:1:1 ()
```

### DIFF
--- a/lib/credo/check/readability/parentheses_in_condition.ex
+++ b/lib/credo/check/readability/parentheses_in_condition.ex
@@ -43,18 +43,13 @@ defmodule Credo.Check.Readability.ParenthesesInCondition do
     collect_parenthetical_tokens(t, acc, head)
   end
 
-  defp parenthetical_condition?({:"(", _} = token) do
-    {elem(token, 0), elem(token, 1), nil}
+  defp parenthetical_condition?({:identifier, _, :if} = start, [next_token | t],
+                                prev_head) do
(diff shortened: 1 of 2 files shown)


rrrene:
Thx for your contribution.

EricWorkman:
Thank you for Credo! It's really helpful!

Issue #171, opened 2016-09-16
Failing ABC test on simple but large map


mayppong:
Hi there,

I couldn't find any ticket on this so hopefully it hasn't been documented somewhere. I am running into a suggestiong for "too complex" report even though the method isn't all that complex aside from the fact that it's really simple. Here is the output from `credo`.

```
  Refactoring opportunities
┃ 
┃ [F] → Function is too complex (ABC size is 34, max is 30).
┃       web/views/api/v1/address_view.ex:12:7 (Ephemeral.Api.V1.AddressView.render)
```
And here is the offending function.
```
def render("address.json", %{address: address}) do
    %{id: address.id,
      street_number: address.street_number,
      street_name: address.street_name,
      street_direction: address.street_direction,
      unit_number: address.unit_number,
      unit_type: address.unit_type,
      description: address.description,
      po_box: address.po_box,
      district: address.district,
      city: address.city,
      region: address.region,
      division: address.division,
      country: address.country,
      postal_code: address.postal_code,
      longitude: address.longitude,
      latitude: address.latitude,
      is_verified: address.is_verified}
  end
```

Does any one have suggestions for working around this? I'd like to not have to change the ABCsize or skip the ABC check for the function altogether. Perhaps a way to make the function better would be preferred.

Side note: this is actually a function generated for me by Phoenix untouched.

rrrene:
This looks like a bug in Credo. Thx for reporting! :+1:

sstern6:
@rrrene looking into this and trying to provide a fix.  Need a little guidance on execution though.  Looks like the reason it is throwing an error is because the value in the key value pair is being counted too many times.  ie,  if you change `street_number: address.street_number` to `street_number: address` for all values, then the rule passes.  So it seems like there is something wrong with the count based off of the AST.

Ill keep digging into this but let me know if you have any guidance or advice.

Thanks

PR #183, opened 2016-10-10
Add new function head for #traverse_abc in abc_size.ex(fixes issue #171)

sstern6:
@rrrene created a new head for `traverse_abc` in order to omit counting of Map/Struct record access using dot notation inside of a function.  Please let me know if there was an edge case I have missed and need to rework the solution or if the patch needs further explanation.

This issue fixes https://github.com/rrrene/credo/issues/171.


rrrene:
Thx for investigating this and working on a fix! :+1: Will release soon!

PR #186, opened 2016-10-11
Add support for string concatentation for pipe chain start

sstern6:
This PR is not reference to an issue, however, per your recommendation for list concatenation, I thought it would be appropriate to have string concatenation supported as well.

You can see a use case in the test file.

Please let me know if you have any questions.


sstern6:
Closing 👍

PR #188, opened 2016-10-12
fix compiler warnings for unused variables

austinsmorris:
This PR does not change functionality, so while the build is currently failing, this PR is independent of that result.  Cheers!


lib/credo/cli/output/explain.ex
```diff
@@ -127,7 +127,7 @@ defmodule Credo.CLI.Output.Explain do
     |> UI.puts
 
     if issue.line_no do
-      {_, line} = Enum.at(source_file.lines, issue.line_no - 1)
+      {_, _line} = Enum.at(source_file.lines, issue.line_no - 1)
```
rrrene: The whole tuple seems to be unused, so we should be able to remove it. Could you check?

austinsmorris:
Ha, you're right!  I didn't even notice I was left with a completely unused tuple.  It looks like this code became irrelevant with ebbacbff99997c5c30afb45fd5598d1653b7cf49.  It should be ready to go now.  Thanks!

rrrene:
:+1: Thx!

Issue #201, opened 2016-10-26
Bug in regex


andrewtimberlake:
Running credo (0.4.13, Elixir 1.3.3) against [https://github.com/andrewtimberlake/elixir-pdf] I get the following error,

```
==> credo
Compiling 120 files (.ex)
Generated credo app
Checking 19 source files ...
** (EXIT from #PID<0.70.0>) an exception was raised:
    ** (FunctionClauseError) no function clause matching in Regex.match?/2
        (elixir) lib/regex.ex:159: Regex.match?(~r/(\.\w+Controller|\.Endpoint|\.Mixfile|\.Repo|\.Router|\.\w+Socket|\w+Test|\.\w+View)$/, nil)
        (elixir) lib/enum.ex:2511: Enum.do_any?/2
        lib/credo/check/readability/module_doc.ex:42: Credo.Check.Readability.ModuleDoc.traverse/4
        (elixir) lib/macro.ex:296: anonymous fn/4 in Macro.do_traverse_args/4
        (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/macro.ex:262: Macro.do_traverse/4
        (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3

16:31:52.311 [error] Task #PID<0.906.0> started from #PID<0.70.0> terminating
** (FunctionClauseError) no function clause matching in Regex.match?/2
    (elixir) lib/regex.ex:159: Regex.match?(~r/(\.\w+Controller|\.Endpoint|\.Mixfile|\.Repo|\.Router|\.\w+Socket|\w+Test|\.\w+View)$/, nil)
    (elixir) lib/enum.ex:2511: Enum.do_any?/2
    lib/credo/check/readability/module_doc.ex:42: Credo.Check.Readability.ModuleDoc.traverse/4
    (elixir) lib/macro.ex:296: anonymous fn/4 in Macro.do_traverse_args/4
    (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (elixir) lib/macro.ex:262: Macro.do_traverse/4
[...]

PR #204, opened 2016-10-26
Replace return value of nil to return string in Module.name

sstern6:
This PR fixes: https://github.com/rrrene/credo/issues/201

### DIFF
--- a/lib/credo/code/module.ex
+++ b/lib/credo/code/module.ex
@@ -178,11 +178,10 @@ defmodule Credo.Code.Module do
     {ast, defs}
   end
 
-  # TODO: write unit test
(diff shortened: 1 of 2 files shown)


rrrene:
Superb. As always. :+1:

One thing: Can we have a test reproducing the old, broken behaviour to ensure it does not happen in the future?

sstern6:
Thanks for the feedback @rrrene, on it and will repush!

sstern6:
@rrrene let me know if you think the placeholder value `"<Unknown Module Name>"` should be changed to something else.

Thanks

rrrene:
`"<Unknown Module Name>"` works perfect for now!

PR #207, opened 2016-10-27
Add ability to specify `strict` in `.credo.exs` config file

### LINKED ISSUE #168: is there a way to specify `strict` in a config?
I have tinkered around and perused the source, but I'm not seeing a way to set `strict` in a config. It makes sense to me that any command line parameters should be configurable in a config. Is this possible currently? I also tried elevating the priority on the low priority checks, to no avail.

devonestes:
It was requested to be able to set `--strict` in the config file, so I've added
that ability here. I have also updated the documentation to make it apparent
that this is a setting that is available to the user.

I had a hard time finding tests for similar functionality, so I haven't added
any tests for this feature. I'd love to do that, though, if someone can point me
in the right direction as to how this should be tested.

I've tried this on my local machine and seen that it works, but I'm also fairly
certain that the implementation that I have isn't consistent with the rest of
the application. If folks have any other guidance on how I can make my
implementation better, I'd be happy to hear it!

Resolves #168

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -31,6 +31,10 @@
       # You can disable this behaviour below:
       check_for_updates: true,
       #
+      # If you want to enforce a style guide and need a more traditional linting
+      # experience, you can change `strict` to true below:
+      strict: false,
+      #
       # You can customize the parameters of any check by adding a second element
       # to the tuple.
       #
--- a/lib/credo/cli.ex
+++ b/lib/credo/cli.ex
@@ -171,7 +171,13 @@ defmodule Credo.CLI do
   defp set_strict(config, %{strict: true}) do
     %Config{config | all: true, min_priority: -99}
   end
-  defp set_strict(config, _), do: config
+  defp set_strict(config, _) do
+    if config.strict do
+      set_strict(config, %{strict: true})
+    else
+      config
+    end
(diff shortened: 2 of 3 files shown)


sstern6:
Thanks for getting this out!  Looks good, but have a few thoughts 😄 .

Did a few local tests and seems to be working properly, but youre right, I cant find any tests testing `--strict` for the cli, @rrrene any recommendations for testing the functionality of `strict`? Might want to add tests for the `cli` ` --strict` as well as for the `config`(that can be a separate PR/issue)?  As the codebase changes it would be nice to know if stuff like this breaks based on a new implementation etc.

From an organizational standpoint and for future development, it might get a little confusing that the function that sets `strict` from the args in config file is in `cli.ex`, it would organizationally make more sense to have it in the `config.ex` file, which sets the `%Config` module to have values`  all: true, min_priority: -99`(if possible???).  This is just my opinion.

Again, thanks for your PR!!!!

cc @rrrene, thoughts?

rrrene:
@devonestes Great work! One thing: @sstern6 is right in that `cli.ex` does not "feel" like the right place to set this. I will merge and provide a proposal! 

Thx from me as well! :+1:

PR #209, opened 2016-10-28
Implement no_case? and add it to variable_names

### LINKED ISSUE #208: Non alphanumeric variable names
We are using ... as a variable name in a macro parameter and I think it doesn't make sense that it would be stopped for not being snake case. I added a no_case check.

olafura:
Fixes #208

### DIFF
--- a/lib/credo/code/name.ex
+++ b/lib/credo/code/name.ex
@@ -65,6 +65,10 @@ defmodule Credo.Code.Name do
     name |> String.match?(~r/^[a-z0-9\_\?\!]+$/)
   end
 
+  def no_case?(name) do
+    name |> String.match?(~r/^[^a-zA-Z0-9]*$/)
+  end
+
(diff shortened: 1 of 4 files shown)


lib/credo/code/name.ex
```diff
@@ -65,6 +65,10 @@ defmodule Credo.Code.Name do
     name |> String.match?(~r/^[a-z0-9\_\?\!]+$/)
   end
 
+  def no_case?(name) do
+    name |> String.match?(~r/^[^a-zA-Z0-9]*$/)
```
rrrene: Is the `*$` intentional to match empty strings (I would have expected `+$`)?

olafura: Fixed

olafura:
I don't understand why the CI tests are failing seems like a bug. I can't find the code it's referring to.

rrrene:
Not your fault. Thx for your work on this!

PR #227, opened 2016-11-13
Always print where check runner errors happened to help with debugging

novaugust:
I made a large change to my codebase and credo's runner started crashing on me. While the debug stack told me where in credo I was crashing, I wasn't able to figure out what code of mine was causing the issue.

With this change, I was able to track down the file that was breaking things for me.

Obviously, I could have gotten the same debug line by turning off the "crash_on_error" setting, but I didn't know about it and don't really want to have to flip config settings to debug.

---------------
Sorry for the second PR. For some reason, force pushing my branch disabled my ability to reopen the original PR


rrrene:
Thx!

Issue #210, opened 2016-10-28
Credo crashes with hex 0.14.0 due to using Hex.Utils.ensure_registry!/0


davydog187:
### Precheck

* For proposing a new feature, be sure to name your issue "Proposal: ..."
* For bugs, please do a quick search and make sure the bug has not yet been reported here

### Environment

* Credo version (`mix credo -v`): 0.4.13
* Erlang/Elixir version (`elixir -v`): 1.3.3
* Operating system: OSx El Capitan

### What where you trying to do?
Updated local.hex to 0.14.0, which breaks credo because its relying on a Hex Util thats no longer available.

### Expected outcome
It not to crash

### Actual outcome

```bash
$ mix credo
** (UndefinedFunctionError) function Hex.Utils.ensure_registry!/0 is undefined or private
    (hex) Hex.Utils.ensure_registry!()
    lib/credo/check_for_updates.ex:4: Credo.CheckForUpdates.run/0
    lib/credo/cli.ex:92: Credo.CLI.run/1
    lib/credo/cli.ex:58: Credo.CLI.main/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2
```

davydog187:
Looks like it was removed in this PR https://github.com/hexpm/hex/pull/300/files

It appears credo is relying on an undocumented API that it should not be.

roehst:
Just happened to me.

** (UndefinedFunctionError) function Hex.Utils.ensure_registry!/0 is undefined or private
(hex) Hex.Utils.ensure_registry!()
lib/credo/check_for_updates.ex:4: Credo.CheckForUpdates.run/0
lib/credo/cli.ex:92: Credo.CLI.run/1
lib/credo/cli.ex:58: Credo.CLI.main/1
(mix) lib/mix/task.ex:296: Mix.Task.run_task/3
(mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2

davydog187:
@roehst to revert back, `mix archive.install https://repo.hex.pm/installs/1.3.0/hex-0.13.2.ez`, FYI

roehst:
@davydog187 tks a lot u just got my ass saved

davydog187:
It looks like you can disable it in your credo config as well: https://github.com/rrrene/credo/blob/2c51773767fb4438403c4a6fac7422bbeba99c51/.credo.exs#L32

[...]

PR #232, opened 2016-11-17
Update README installation instructions to v0.5

dkarter:
Since 0.4 does not work properly with the Hex v0.14.0. See #210.

_Update:_
I did see that v0.4.14 addresses this issue for v0.4, but still seems like a good idea to have new projects use 0.5.


rrrene:
:+1:

PR #237, opened 2016-11-23
Fix RedundantBlankLines raising when file doesn't have blank lines

mememori:
If you had a file without ANY blank line, RedundantBlankLines would raise because `consecutive_lines([], _)` was returning `:false` and that value was being piped into an `Enum.map/2`

An example of file that would trigger the error behaviour without this patch:
```elixir
defmodule MyApp do
end
````


rrrene:
:+1: Thx!

PR #238, opened 2016-11-23
Fix bug on ParameterPatternMatchingTest on private funs and funs with no binding

mememori:
`position_collector.ex` checked the parameter definitions of the files and flat_mapped this. But when the input ast node was either a `def` without bindings or a `defp` or `defmacro`, it would return `nil` .

Example of input file that would have triggered an error:
```elixir
defmodule Foo do
  def bar do
    :bar
  end
  defp baz(foo) do
    foo
  end
end
```


rrrene:
Thx for your contribution! 👍

Issue #221, opened 2016-11-07
Suggestion: NeverPipeOnce


josevalim:
We at the Elixir core team consider a smell to use the pipe just once: `value |> Some.fun`. The pipe was designed to express pipelines and you don't have a pipeline if you have only one call. :) Therefore it would be nice to have a check that guarantees you never pipe only once.

PS: We are adding Credo to Ecto via ebert: https://ebertapp.io/github/elixir-ecto/ecto/reviews/1. It is working quite well so far! Thank you!

PS: Read the discussion below for examples of where piping once is actually useful in case you want to disable the check.

rrrene:
It is a good idea to have a check for this. :+1: 

I did not know of this preference by the core team. Just as an anecdote: The "single pipe" helped me alot in the beginning because `list |> Enum.take(5)` almost felt like `list.take(5)` (coming from an OO background).

josevalim:
That's part of why we believe it is an important check. It helps to keep old habits at home. :)

rrrene:
> It helps to keep old habits at home. :)

Haha, I might just print & frame that. :grin:

sstern6:
@rrrene would you say this would be a `Refactor` check or a `Warning`?  or neither 😄 ?

svoynow:
I'm working on this one at the moment. I'm making it a `Refactor` for the moment, but I'm happy to change.

forki:
But isn't a single pipe the beginning of writing a longer pipeline? I'm working > 10 years in F# and I think this really artificial constraint. Are there any technical limitations that lead to this suggestion?

josevalim:
@forki I would consider this about enforcing a style rather than being technical limitation. For new developers, sometimes a single pipe is used to cope with the lack of OO methods:

```elixir
def my_function(arg, msg) when arg |> is_tuple do
  arg |> send(msg |> format)
end
```

which in turns makes the function much harder to read.

I cannot think of any situation where a single pipe is more readable than no pipes so I would like to enforce that in my codebases but your mileage may vary. :)

polytypic:
I have to second @forki, though I have only cursory interest in Elixir at this point. I also have a long history of programming and programming in functional languages.  I designed [A family of function application operators for Standard ML](http://comp.lang.functional.narkive.com/zZJZg20r/a-family-of-function-application-operators-for-standard-ml) some 11 years ago.

[...]

PR #240, opened 2016-11-23
refactor check for single pipes

svoynow:
fix for https://github.com/rrrene/credo/issues/221

### DIFF
--- a/lib/credo/check/refactor/single_pipe.ex
+++ b/lib/credo/check/refactor/single_pipe.ex
@@ -1,0 +1,36 @@
+defmodule Credo.Check.Refactor.SinglePipe do
+  @moduledoc """
+    "The pipe was designed to express pipelines and you don't have a pipeline if you have only one call"
+    So instead of foo |> bar, just use foo(bar)
+  """
+
+  @explanation [check: @moduledoc]
+
+  use Credo.Check
+
+  def run(%SourceFile{ast: ast} = source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    {_continue, issues} = Credo.Code.prewalk(ast, &traverse(&1, &2, issue_meta), {true, []})
+    issues
+  end
+
+  defp traverse({:|>, _, [{:|>, _, _} | _]} = ast, {_, issues} = state, issue_meta) do
+    {ast, {false, issues}}
+  end
+
+  defp traverse({:|>, meta, _} = ast, {true, issues}, issue_meta) do    
+    {ast, {false, issues ++ [issue_for(issue_meta, meta[:line], "single pipe")]}}
+  end
+  
+  defp traverse(ast, {_, issues}, _issue_meta) do    
+    {ast, {true, issues}}
+  end
+
+  def issue_for(issue_meta, line_no, trigger) do
+    format_issue issue_meta,
+      message: "There's no reason to use a single pipe character. Replace with a simple function call",
+      trigger: trigger,
+      line_no: line_no
+  end
+end
--- a/test/credo/check/refactor/single_pipe_test.exs
+++ b/test/credo/check/refactor/single_pipe_test.exs
@@ -1,0 +1,50 @@
+defmodule Credo.Check.Refactor.SinglePipeTest do
+  use Credo.TestHelper
+
+  @described_check Credo.Check.Refactor.SinglePipe
+
+  test "it should NOT report expected code" do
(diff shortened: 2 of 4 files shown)


brienw:
👍  nice!

test/credo/check/refactor/single_pipe_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test "it should report a violation for multiple violations" do
+"""
+defmodule CredoSampleModule do
+  use ExUnit.Case
+
+  def some_fun do
+    some_val |> do_something
+    some_other_val |> do_something
```
rrrene: Let's put `do_something` on the next line here (to cover the multi-line case). :+1:

lib/credo/check/refactor/single_pipe.ex
```diff
@@ -0,0 +1,36 @@
...
+    {ast, {false, issues ++ [issue_for(issue_meta, meta[:line], "single pipe")]}}
+  end
+  
+  defp traverse(ast, {_, issues}, _issue_meta) do    
+    {ast, {true, issues}}
+  end
+
+  def issue_for(issue_meta, line_no, trigger) do
+    format_issue issue_meta,
+      message: "There's no reason to use a single pipe character. Replace with a simple function call",
```
rrrene: We can do better here. Maybe "Pipes should not be used for single operations."? What do you think?

lib/credo/check/refactor/single_pipe.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Credo.Check.Refactor.SinglePipe do
+  @moduledoc """
+    "The pipe was designed to express pipelines and you don't have a pipeline if you have only one call"
```
rrrene: First draft for `@moduledoc`:

```
Pipes (`|>`) should only be used when piping data through multiple calls.

So while this is fine:

    list
    |> Enum.take(5)
    |> Enum.shuffle
    |> evaluate

The code in this example ...

    list
    |> evaluate

... should be refactored to look like this:

    evaluate(list)

A pipe should only ever be used to represent a pipeline of data. Therefore, when you only have a single `|>` you do not have a pipeline, it's just a single call.
```

lib/credo/check/refactor/single_pipe.ex
mikestok: "Use a function call when a pipeline is only one function long" maybe?

josevalim: "Use a function call when a pipeline is only one function long" - I like this. :+1:

lib/credo/check/refactor/single_pipe.ex
rrrene: I like it! (but you meant "last sentence", right?)

davydog187: Borrowing from the other comment.

> A pipeline should be used to represent a series of data transformations. Use a function call when a pipeline is only one function long.

josevalim: Given the discussion in the original issue about style/technical limitation, maybe we should make this a style check? This way we can be a bit more opinionated in the last paragraph:

> Using a single `|>` to invoke functions makes the code harder to read. Instead, write a function call when a pipeline is only one function long.

davydog187: `mix credo categories` defines refactoring as 

> The Refactor checks show you opportunities to avoid future problems and technical debt.

And I would have to agree with @josevalim that this falls in line with style better.

svoynow:
OK, I'll synthesize all the documentation changes.
So is the proposal that instead of the `Refactor` category we put it in `Design`? `Readability`?
We don't really have a category that's maps to *style* in the sense of "this is/isn't idiomatic elixir", do we?

lib/credo/check/refactor/single_pipe.ex
rrrene: I agree. Let's move this to Readability. :+1:

rrrene:
@svoynow The `Readability` category is fine. For example, the `Readability.ModuleAttributeNames` is a check that revolves around something being idiomatic. "Readable" is seen in a broad sense here atm.

svoynow:
I've moved the check and made the requested changes to the moduledoc / message

rrrene:
Thank you sir! :neckbeard:

PR #243, opened 2016-11-26
Add ability to use `--no-strict` command line flag

### LINKED ISSUE #214: Setting `strict` from Config instead of CLI
This continues a discussion from #207.

I added a function to manage the (implicit) settings of `strict` mode in eb546b9.

This should make setting `strict` more explicit while keeping the internal consequences more implicit.

Toughts?

/cc @sstern6 @devonestes

devonestes:
We implemented the ability to set a `strict: true` config in the `.credo.exs`
file for a project, but when we did that we didn't implement any way to
override that with a command line flag. In this commit we're adding that
ability to override that setting with a `--no-strict` flag so you don't need to
edit that file to turn off `strict` mode for just a single run.

Closes #214

### DIFF
--- a/lib/credo/config.ex
+++ b/lib/credo/config.ex
@@ -254,5 +254,8 @@ defmodule Credo.Config do
   def set_strict(%__MODULE__{strict: true} = config) do
     %__MODULE__{config | all: true, min_priority: -99}
   end
+  def set_strict(%__MODULE__{strict: false} = config) do
+    %__MODULE__{config | all: false, min_priority: 0}
+  end
   def set_strict(config), do: config
 end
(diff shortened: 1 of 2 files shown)


lib/credo/config.ex
```diff
@@ -255,4 +255,11 @@ defmodule Credo.Config do
     %__MODULE__{config | all: true, min_priority: -99}
   end
   def set_strict(config), do: config
+
+  @doc """
+  Sets the config values which `no-strict` implies (if applicable).
+  """
+  def set_no_strict(%__MODULE__{strict: false} = config) do
```
devonestes: Ya know, I didn't think of that, but now that you've brought it up I agree that it's a better name for that function. I'll make the change now!

rrrene:
Thx! :+1:

Issue #229, opened 2016-11-13
Readability.LargeNumbers false positives


novaugust:
### Environment

* Credo version (`mix credo -v`): 0.5.2
* Erlang/Elixir version (`elixir -v`): 1.3.0
* Operating system: osx

### What where you trying to do?

I have lists with lots of floats and large numbers. Sometimes credo is giving me false positives on needing underscores.

For this data, the first 7 lines are all fine, but it's compalning about the last 5.

```elixir
test_data = [
      {[14_434, 0.000, 0.250, -0.500, 4_200_000], {14_193.7059228487, -461.81222427595}},
      {[7_275, -0.013, 0.300, -0.400, 3_242_800], {6_705.06114827335, -686.923024527031}},
      {[2_656, 0.000, 0.320, -0.283, 1_070_000], {1_875.04749892584, -540.482846075626}},
      {[6_928, 0.011, 0.280, -0.283, 2_580_000], {6_238.09826567172, -633.308400721897}},
      {[6_928, -0.011, 0.280, -0.283, 2_580_000], {6_238.09625455701, -633.328209922203}},
      {[6_928, 0.011, 0.280, 0.283, 2_580_000], {6_238.09625455701, 633.328209922203}},
      {[6_928, -0.011, 0.280, 0.283, 2_580_000], {6_238.09826567172, 633.308400721897}},
      {[14_434, 0.000, 0.287, -0.526, 8_740_000], {13_922.7016262733, -872.190348982156}}, #bad
      {[7_270, 0.000, 0.287, -0.526, 68_000], {7_262.51042806277, -13.6958182877388}}, #bad
      {[7_270, 0.000, 0.287, -0.526, 70_000], {7_262.28975554153, -14.0984231358245}}, #bad
      {[7_270, 0.000, 0.200, -0.500, 70_000], {7_263.34062452151, -16.5934712666501}}, #bad
      {[7_270, 0.000, 0.200, 0.500, 70_000], {7_263.34062452151, 16.5934712666501}} #bad
    ]
```
In each of the last 5 lines, the 5th item in the line is being reported as not having underscores

novaugust:
I'm zeroing in on the 0.000 because changing it fixed the issue... but on lines further up I don't have that issue. So, here's the exact code that's mad at me

```
test_data = [
      {[14_434, 0.000, 0.250, -0.500, 4_200_000], {14_193.7059228487, -461.81222427595}}, #ok
[...]

PR #244, opened 2016-11-28
Add deprecation and exclusion of checks by Elixir version

devonestes:
We have some checks that don't behave entirely as expected for some older
versions of Elixir, and we want to deprecate (and eventually remove) thos
checks for those versions. This is the first very quick and _very_ dirty spike
on a way to do that so we can start a discussion around how we want to handle
this process.

This is related to #229 and will hopefully give us a starting point for discussion.

### DIFF
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -33,6 +33,11 @@ defmodule Credo.Check do
         default || :unknown
       end
 
+      def elixir_version do
+        default = unquote(opts[:elixir_version])
+        default || ">= 0.0.1"
+      end
+
       def run_on_all? do
         unquote(run_on_all_body(opts[:run_on_all]))
       end
--- a/lib/credo/check/readability/large_numbers.ex
+++ b/lib/credo/check/readability/large_numbers.ex
@@ -23,7 +23,7 @@ defmodule Credo.Check.Readability.LargeNumbers do
     only_greater_than: 9_999,
   ]
 
-  use Credo.Check, base_priority: :high
+  use Credo.Check, base_priority: :high, elixir_version: ">= 1.3.2"
 
   def run(source_file, params \\ []) do
     issue_meta = IssueMeta.for(source_file, params)
--- a/lib/credo/check/runner.ex
+++ b/lib/credo/check/runner.ex
@@ -69,12 +69,16 @@ defmodule Credo.Check.Runner do
 
   defp exclude_checks_based_on_elixir_version(config) do
     version = System.version()
+    skipped_checks = Enum.reject(config.checks, fn
+      ({check}) -> Version.match?(version, check.elixir_version)
+      ({check, _}) -> Version.match?(version, check.elixir_version)
+    end)
     checks = Enum.filter(config.checks, fn
       ({check}) -> Version.match?(version, check.elixir_version)
       ({check, _}) -> Version.match?(version, check.elixir_version)
     end)
 
-    %Config{config | checks: checks}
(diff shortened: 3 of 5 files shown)


devonestes:
So, in general, what I have here is a list of checks that we want to deprecate for certain versions of Elixir, and then another list of checks that we want to exclude (or maybe `disable` is a better name?) for certain versions. We'll need a much more robust way of keeping track of those two lists, but I didn't want to dive too deeply into that for this first run.

When something is deprecated, it prints that it's deprecated, and when it's excluded we remove it from the `config` as if it were excluded manually by the user. We might also want to have some optional printing to the console so users know that it's happening - otherwise for a new user it might appear to be magically excluding things for no reason. However, having that all the time might get really annoying, which is why I didn't go that route for now.

rrrene:
This is an interesting approach. I had something different in mind. I have a 4 hour train ride ahead of me, so maybe I'm able to write it up, so we can discuss it using real code :+1:

rrrene:
Here's the quick and dirty solution I came up with on the train: https://github.com/rrrene/credo/commit/86c69b18c371065f584b165d6a5ea4a8d9d6cd8c

It is more flexible, since it is not targeting single Elixir versions. But it is lacking the message to the user (that a configured check will not be run).

I think we can find a good solution somewhere in the middle :grin:

devonestes:
@rrrene Yeah, that's gonna be easier to manage in the long run! Would we have two different options for each check - one for `elixir_version_deprecate: ">= 1.3.0", elixir_version_exclude: ">= 1.0.0, < 1.3.0"` or something like that? I think we're definitely going to want to be able to both deprecate based on version and exclude based on version - although maybe not both at the same time.

rrrene:
The question here is if we really deprecate checks based on Elixir version? I would have thought that checks are deprecated based on Credo version (e.g. "This check is deprecated and will be removed in future versions of Credo") and excluded based on Elixir version ("This check is excluded for this version of Elixir").

WDYT?

devonestes:
So, let's imagine I'm on Elixir 1.2, and I upgrade Credo to some new version that excludes a check. I only see "This check is excluded for this version of Elixir", right? I wouldn't see any deprecation of that check first?

I think that could work, but I also imagine that some users might find it confusing. On the upside, though, it could encourage users to upgrade their Elixir versions, which is always good!

rrrene:
The problem is that i don't see it working any other way. I'm on Elixir 1.2 and upgrade to a newer version of Credo which includes a check that does not work on 1.2 - we can not deprecate stuff in this scenario, right?

Other scenario: we are on Elixir 1.3.0 where a check throws false positives, because `:elixir_tokenizer` behaves in an unexpected way. We can't deprecate (as opposed to exclude) a check here either, because we do not want people to get the false positives.

That's why I think that deprecation of individual checks is either a true/false thing or tight to a specific future version of Credo (like "this is deprecated now and will be excluded in Credo >= 0.6.0").

devonestes:
Yeah, I get what you're saying now. You're right, it does need to be done this way.

Well, I can take a crack at working on that a bit this week!

devonestes:
Ok, I've taken another swing at this based off of the version that @rrrene threw together. The big addition to that version that I made was some UI output for the user so they don't _think_ the check is running because they explicitly included it in their config file but since it's excluded anyway they might think it's running and just not finding any errors. that would be super confusing.

I'm not in love with the output that I have at the moment, but it might be good enough. Below is a screenshot of what it would look like for a user on 1.3.0 as of today.

credo_output

rrrene:
FYI: I moved the message to "above the analysis".

image

PR #251, opened 2016-11-29
Refactor `cli.ex` to extract module for setting config

devonestes:
We had a lot of private methods that all were essentially part of the same
operation - setting the `Config` struct - part of a module that had a lot going
on, so I extracted them all to a new module to group that behavior together and
hopefully make it a little easier to understand both modules now.

### DIFF
--- a/lib/credo/cli.ex
+++ b/lib/credo/cli.ex
@@ -12,7 +12,7 @@ defmodule Credo.CLI do
   alias Credo.Config
   alias Credo.Sources
   alias Credo.CLI.Filename
-  alias Credo.CLI.Output.UI
+  alias Credo.CLI.Switches
 
   @default_dir "."
   @default_command_name "suggest"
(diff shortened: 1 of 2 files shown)


lib/credo/cli.ex
```diff
@@ -146,99 +146,9 @@ defmodule Credo.CLI do
...
-    |> set_format(switches)
-    |> set_help(switches)
-    |> set_ignore(switches)
-    |> set_min_priority(switches)
-    |> set_only(switches)
-    |> set_read_from_stdin(switches)
-    |> set_strict(switches)
-    |> set_verbose(switches)
-    |> set_version(switches)
+    |> Options.parse_to_config(switches)
```
rrrene: This is a great idea. Thx! One minor nitpick: I think we should name the module `Switches` instead of `Options`. I am not a native speaker, but to me it feels closer to `CLI` (and more readable in this instance).

Thoughts?

rrrene: Or rename the other instances of `switches` into `options`. That would work for me as well. I just want to avoid ambiguous naming. :+1:

devonestes: Naming is so hard! `Switches` is indeed a much more consistent name - I'll go with that.

rrrene:
Thx! :+1:

Issue #247, opened 2016-11-28
Proposal: VariableRebinding


rrrene:
Although it is possible in Elixir, some people might want to shy away from variable re-binding, since it reminds them too much of matuability in their previous programming language.

A check could warn whenever a variable is re-binded. An optional parameter could allow for function parameters to be rebinded once (since we might want to allow normalization of input parameters while retaining the name).

narrowtux:
I'll try

rrrene:
Thx again! :+1:

PR #258, opened 2016-11-29
Implemented check for variable rebinding

narrowtux:
This will check for multiple declaration of the same variable within a do block. 

For example,

```elixir
  def some_function() do
    a = 1
    a = 2
  end
end
```

would trigger a warning because `a` is declared twice.

### DIFF
--- a/test/credo/check/refactor/variable_rebind_test.exs
+++ b/test/credo/check/refactor/variable_rebind_test.exs
@@ -78,4 +78,16 @@ end
 """ |> to_source_file
     |> assert_issue(@described_check)
   end
+
+  test "it should report violations when using destructuring maps" do
+"""
+defmodule CredoSampleModule do
+  def some_function(opts) do
+    %{a: foo, b: bar} = opts
+    bar = 3
+  end
+end
+""" |> to_source_file
+    |> assert_issue(@described_check)
+  end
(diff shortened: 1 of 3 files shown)


test/credo/check/refactor/variable_rebind_test.exs
```diff
@@ -0,0 +1,44 @@
...
+""" |> to_source_file
+    |> assert_issue(@described_check)
+  end
+
+  test "it should report two violations" do
+"""
+defmodule CredoSampleModule do
+  def some_function() do
+    var_1 = 1 + 3
+    var_b = var1 + 7
```
rrrene: 💩  greetings from hh.ex :-P

narrowtux:
Proper tagging now: Implements #247

narrowtux:
TODO: 
 -[ ] show last occurrence of same variable instead of first one
 -[ ] properly handle descructuring

narrowtux:
@rrrene all done except the `@moduledoc`. 

Also I think that redeclaring should be an extra check because I feel like it should have a higher priority than this check.

narrowtux:
Also don't want anyone to miss this gem:

image

rrrene:
Thx! :+1:

PR #260, opened 2016-11-30
Refactor to remove single pipes

devonestes:
I wanted to do a little refactoring to see how much of a difference the new
single pipe rule would make, and it turns out that a) we were using that
_everywhere_, and b) it actually makes the code a lot easier to reason about.
I think it's a great rule!

This commit doesn't remove _all_ of the single pipe instances since I figured
doing that in one huge PR might be overwhelming, but it's a step in the right
direction. If it would be helpful to have all of them removed in one PR then
I'm happy to do that!

### DIFF
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -105,12 +105,12 @@ defmodule Credo.Check do
     priority =
       case params[:priority] do
         nil -> issue_base_priority
-        val -> val |> Check.to_priority
+        val -> Check.to_priority(val)
       end
     exit_status =
       case params[:exit_status] do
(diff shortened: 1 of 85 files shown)


rrrene:
:+1: This is definitely a step in the right direction.

If you would like to add the "missing" single pipes to this PR, I would hold off on merging. 

Thx for all your contributions! They are highly appreaciated. :+1: :grin:

devonestes:
@rrrene Ok, all done! Hopefully it's not too much of a pain to review 😜

rrrene:
LGTM! :+1:

PR #264, opened 2016-12-05
Add check for prefering implicit `try`

### LINKED ISSUE #224: Proposal: prefer implicit try in functions
A useful, but little known feature of elixir is the implicit `try` that is placed around things like `def`, and also some other macros that create functions, like `test` from ExUnit.

For example, the two functions below are equivalent:
```elixir
def format(value) do
  try
    do_format(value)
  catch
    :exit, reason -> {:error, reason}
  end
end

def format(value) do
  do_format(value)
catch
  :exit, reason -> {:error, reason}
end
```

I believe this is a beneficial transformation simplifying code and decreasing nesting. Additionally this check should be fairly easy to detect - when the function's body is a single try expression.

devonestes:
We don't need to explicitly use `try` inside function definitions, and this
check will raise low priority issues when `try` is explicitly used.

Closes #224

### DIFF
--- a/lib/credo/check/readability/prefer_implicit_try.ex
+++ b/lib/credo/check/readability/prefer_implicit_try.ex
@@ -33,12 +33,34 @@ end
     Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
   end
 
-  defp traverse({:try, [line: line_no], _} = ast, issues, issue_meta) do
+  defp traverse({:def, _, [{_, _, _}, [do: {:try, [line: line_no], _}]]} = ast, issues, issue_meta) do
     {ast, issues ++ [issue_for(issue_meta, line_no)]}
+  end
+
+  defp traverse({:def, _, [{_, _, _}, [do: {:__block__, _, function_body}]]} = ast, issues, issue_meta) do
+    process_function_body(function_body, ast, issues, issue_meta)
+  end
+
+  defp traverse({:def, _, [{_, _, _}, function_body]} = ast, issues, issue_meta) do
+    process_function_body(function_body, ast, issues, issue_meta)
   end
 
   defp traverse(ast, issues, _issue_meta) do
     {ast, issues}
+  end
+
+  defp process_function_body(function_body, ast, issues, issue_meta) do
+    found = Enum.find(function_body, fn
+      ({:try, _, _}) -> true
+      (_) -> false
+    end)
+
+    if is_nil(found) do
+      {ast, issues }
+    else
+      {:try, [line: line_no], _} = found
+      {ast, issues ++ [issue_for(issue_meta, line_no)]}
+    end
   end
 
   defp issue_for(issue_meta, line_no) do
--- a/lib/credo/check/readability/prefer_implicit_try.ex
+++ b/lib/credo/check/readability/prefer_implicit_try.ex
@@ -1,0 +1,50 @@
+defmodule Credo.Check.Readability.PreferImplicitTry do
+  @moduledoc """
+We don't need to explicity use `try` in function definitions. For example, this:
+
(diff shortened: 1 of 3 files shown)


test/credo/check/readability/prefer_implicit_try_test.exs
```diff
@@ -0,0 +1,33 @@
...
+
+  test "it should NOT report expected code" do
+"""
+defmodule ModuleWithImplicitTry do
+  def failing_function(first) do
+    to_string(first)
+  rescue
+    _ -> :rescued
+  end
+end
```
rrrene: Please add another test covering this scenario:

```elixir

   def some_fun(first) do
      other_fun()

      string =
        try do
          to_string(first)
        rescue
          _ -> :rescued
       end

      is_atom(string)
   end
``` 

This should *not* yield an issue, as there is no option for an *implicit* try here.

rrrene: You misread my comment

> This should **not** yield an issue, as there is no option for an **implicit** try here.

You should be able to detect if a function body consists of *only* a `try` block and *only* yield an issue in that narrow use case. :+1:

lib/credo/check/readability/prefer_implicit_try.ex
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  defp process_function_body(function_body, ast, issues, issue_meta) do
+    found = Enum.find(function_body, fn
+      ({:try, _, _}) -> true
+      (_) -> false
+    end)
+
+    if is_nil(found) do
+      {ast, issues }
```
rrrene: Please remove the white-space after `issues`.

lib/credo/check/readability/prefer_implicit_try.ex
```diff
@@ -0,0 +1,72 @@
...
+  defp traverse({:def, _, [{_, _, _}, function_body]} = ast, issues, issue_meta) do
+    process_function_body(function_body, ast, issues, issue_meta)
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp process_function_body(function_body, ast, issues, issue_meta) do
+    found = Enum.find(function_body, fn
```
rrrene: Please format multi-line assignments like this:

```elixir
      str =
        try do
          to_string(first)
        rescue
          _ -> :rescued
        end
```

test/credo/check/readability/prefer_implicit_try_test.exs
devonestes: Ah, ok, I misunderstood. I thought the reason we needed the explicit `try` in your example was because that was basically a function and we cared about the return value from that function. Technically in the second test that I added we don't need the `try` there - it will still rescue if there is a failure, but I see now that we need the `try` to ensure that we're only rescuing errors in that scope and _that's_ why we need the explicit `try` there. I'll give it another whirl!

rrrene: Perfect! :grin:

rrrene:
Thx! I merged this on the command line, as I had to do a couple of quick fixes.

A word of advice: You might want to check your text editor, because the indentation was not fixed, but worse than in the previous commit (also: trailing white-space).

image

I really like your contributions. :+1: But from my experience there will be people not merging your otherwise fine pull requests because of "hygenic" things like these.

devonestes:
@rrrene Happy to help, and thanks for the input!

I'm definitely still working on my tooling for Elixir. I have a lot of my Ruby stuff automated (like stripped whitespace and in-editor notices for strange indentation), but I'm still figuring some of that stuff out for Elixir. I guess a good first step is getting that trailing whitespace automated!

PR #269, opened 2016-12-08
Fix `prefer_implicit_try` check

devonestes:
I realized that the implementation that I had for the
`PreferImplicitTry` check was incomplete and wouldn't raise an issue
when used in private functions. That's been added here, as well as a
test to cover that additional case.

### DIFF
--- a/lib/credo/check/readability/prefer_implicit_try.ex
+++ b/lib/credo/check/readability/prefer_implicit_try.ex
@@ -27,6 +27,8 @@ defmodule Credo.Check.Readability.PreferImplicitTry do
 
   @explanation [check: @moduledoc]
 
+  @def_ops [:def, :defp]
+
   use Credo.Check, base_priority: :low
 
   def run(%SourceFile{} = source_file, params \\ []) do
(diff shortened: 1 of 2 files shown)


rrrene:
Now that you're mentioning it ... 😅 

Should we add another case for `defmacro`?

devonestes:
Frankly I have no idea if `try` is implicit in `defmacro` as well as `def` and `defp`, or how that would behave since it's rather different than good ol' `def`, so that's why I didn't go there just yet. I can investigate, though, and see if we want to extend this check there as well if you think it's worth it.

rrrene:
I have to admit my Erlang Fu is nowhere near where it should be, but I read this

https://github.com/elixir-lang/elixir/blob/8cd972ed173d78c1b561c2aa8e9b1bfab52f0d57/lib/elixir/src/elixir_bootstrap.erl#L18

like the same behaviour should appy to `defmacro` (which would make sense to me, as macros are just functions altering things at compile-time).

But: I could be wrong 😰

devonestes:
Cool - well I'll play around and see if I can get to the bottom of it!

NobbZ:
Also in the called function `define/4` there seems to be no special casing for `defmacro`(`p`). But, the info which `def*` we have, is deeply burried in its return value. So in theory, there could still be special casing on the caller site, which might be much uglier to identify!

rrrene:
Thx! :+1:

PR #270, opened 2016-12-09
Implement `ignore_strings` param for `LineLength`

### LINKED ISSUE #122: Implement the ignore_strings option for line length checks
The code to ignore strings when checking line length is currently unimplemented. This is just a gentle reminder from #120

devonestes:
We had an unimplemented config key for the `LineLength` check, which
I've gone ahead and implemented in this PR. When set to true, it will
ignore any line that consists of only a single string literal, or any
line that is in a heredoc that is longer than the set maximum line
length.

Resolves #122


rrrene:
Thx for this! We really need to implement all these "teased at" optional features before going 1.0 :grin:

But I guess here we should use `CodeHelper.clean_strings_and_sigils/1` before the analysis in case the user wants to ignore strings (instead of re-implenting the string removal).

rrrene:
You are right. Thx! :+1:

PR #278, opened 2016-12-15
Remove non existing modules from example config

### LINKED ISSUE #277: Credo.Check.Readability.StringSigils.base_priority/0 is undefined
### Environment

* Credo version (`mix credo -v`): 0.5.3
* Erlang/Elixir version (`elixir -v`): Elixir 1.4.0-rc.1 (16a14c1f6)
* Operating system: macOS 10.12.1 (16B2555)

Add .credo.exs to umbrella app config folder, using the provided full configuration example

### Expected outcome
successful execution of `mix credo --strict`

### Actual outcome
** (UndefinedFunctionError) function Credo.Check.Readability.StringSigils.base_priority/0 is undefined (module Credo.Check.Readability.StringSigils is not available)
    Credo.Check.Readability.StringSigils.base_priority()
    lib/credo/check/runner.ex:65: anonymous fn/2 in Credo.Check.Runner.exclude_low_priority_checks/2
    (elixir) lib/enum.ex:1833: anonymous fn/3 in Enum.reject/2
    (elixir) lib/enum.ex:1755: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:1833: Enum.reject/2
    lib/credo/check/runner.ex:63: Credo.Check.Runner.exclude_low_priority_checks/2
    lib/credo/check/runner.ex:10: Credo.Check.Runner.run/2

So it looks like there is a module missing that the current config expects. Simple fix would be to remove that one line from the config for the time being.

mkrogemann:
Fixes #277

I guess these modules may be coming in future releases. But I feel that it would be better to not have them included at this time as the example configuration is currently not runnable.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -80,12 +80,12 @@
         {Credo.Check.Readability.ParenthesesInCondition},
         {Credo.Check.Readability.PredicateFunctionNames},
         {Credo.Check.Readability.RedundantBlankLines},
-        {Credo.Check.Readability.StringSigils},
+        # {Credo.Check.Readability.StringSigils}, # module does not exist in 0.5.3
         {Credo.Check.Readability.TrailingBlankLine},
         {Credo.Check.Readability.TrailingWhiteSpace},
         {Credo.Check.Readability.VariableNames},
-        {Credo.Check.Readability.PreferImplicitTry},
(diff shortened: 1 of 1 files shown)


rrrene:
See #277

Issue #205, opened 2016-10-27
Proposal: Missing @spec check for functions


asummers:
Rationale: We love dialyzer. Being able to have our linter yell at us when we don't provide `@spec` annotations would be awesome. They serve as function documentation, as well as helping Dialyzer do its job better. Additionally, I think this rule would serve as amazing documentation on how one might check that all functions have a certain annotation (say Ecto with `@primary_key` and `@fields` or something). To me, these are both basically the same conceptual feature: given a module and it's functions, do all the `def`s and `defp`s have `@spec` annotations and for `@doc`, just `def`. 

Possible solution: Looking through some of the existing rules, I believe it can be accomplished relatively simply by traversing the AST by finding all `:def` and `:defp` function names/arity, making a set of those, and doing the same thing with `@spec`s, and then comparing those two sets. If they're not equal, you missed one. I tried extending the `@moduledoc` one but was unable to at first stab, however I think the approach above should get what is necessary for this.

NobbZ:
At least the `@doc`-part is covered by [rrrene/inch_ex](https://github.com/rrrene/inch_ex) pretty well I think, but for the `@spec`-part of your proposal you get a very fat 👍. I was going to write just such a proposal ;)

rrrene:
So basically require `@spec` on all functions?

asummers:
Yep! Once that machinery is in place, there are more interesting things one can do like enforce consistency on use of parentheses after `map` or `atom` or use or `℅Struct{}` vs `Struct.t`, though I don't know if mainline Credo necessarily wants to be in the business of checking those.

asummers:
Yep! Once that machinery is in place, there are more interesting things one
can do like enforce consistency on use of parentheses after map or atom or
use or ℅Struct{} vs Struct.t, though I don't know if mainline Credo
necessarily wants to be in the business of checking those.

sorentwo:
I recently began working on @doc support (as I thought it would be easier than @spec). I managed to get a basic check working for `def`, `defmacro`, and `defcallback`, but it wasn't entirely accurate. After quite a bit of time looking at ex_doc and h I discovered they both use `Code.get_docs`. Unfortunately, `Code.get_docs` relies on a compiled .beam file for doc extraction, which isn't how Credo tests work currently.

[...]

PR #279, opened 2016-12-16
Add missing type specification check for functions

sorentwo:
The `MissingTypespecs` readability check compares specs against definitions using the name and the arity. This implementation only checks for specs on `def` and ignores `defp`. Based on a brief survey of core elixir projects it seems rather common to skip typespecs for `defp`, though it could be configurable very easily.

Due to the presence of the function name right in the spec this was _vastly_ easier to work with than `@doc` tags.

This addresses #205

### DIFF
--- a/lib/credo/check/readability/missing_typespecs.ex
+++ b/lib/credo/check/readability/missing_typespecs.ex
@@ -1,0 +1,58 @@
+defmodule Credo.Check.Readability.MissingTypespecs do
+  @moduledoc """
+  Functions, callbacks and macros need typespecs.
+
+  Adding typespecs allows tools like dialyzer to perform success typing on
+  functions. Without a spec functions and macros are ignored by the type
+  checker.
+
+      @spec add(integer, integer) :: integer
+      def add(a, b), do: a + b
+
+  Functions with multiple arities need to have a spec defined for each arity:
+
+      @spec foo(integer) :: boolean
+      @spec foo(integer, integer) :: boolean
+      def foo(a), do: a > 0
+      def foo(a, b), do: a > b
+
+  The check only considers whether the specification is present, it doesn't
+  perform any actual type checking.
+  """
+
+  @explanation [check: @moduledoc]
+
+  use Credo.Check
+
+  def run(%SourceFile{} = source_file, params \\ []) do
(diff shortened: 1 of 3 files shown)


lib/credo/check/readability/missing_typespecs.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  The check only considers whether the specification is present, it doesn't
+  perform any actual type checking.
+  """
+
+  @explanation [check: @moduledoc]
+
+  use Credo.Check
+
+  def run(%SourceFile{} = source_file, params \\ []) do
```
sorentwo: Note the ironic lack of a `@spec` within the typespec checker.

rrrene:
Thx! :+1:

josevalim:
For what is worth, dialyzer has an option for doing exactly this. :)

mgartner:
@josevalim what is the name of the option? I don't see it documented here: http://erlang.org/doc/man/dialyzer.html

josevalim:
You are right @mgartner. Maybe it is a compiler option?

mgartner:
@josevalim Maybe you are thinking of the `+warn_missing_spec` option of `erlc`? It seems undocumented, but I found a few references to it, such as [this](https://stackoverflow.com/questions/12092682/warnings-for-functions-without-type-specs).

How would that option be used from an Elixir project?

mgartner:
Looks like there is also a `warn_missing_spec_all ` option: http://erlang.2086793.n4.nabble.com/warn-missing-spec-td4689325.html

shouston3:
@mgartner did you manage to find a way to pass this option to the dialyzer mix command?

mgartner:
@shouston3 I have not.

Stratus3D:
For anyone who is curious, this seems to be the only way to turn on the `+warn_missing_spec` option on: 

```
ERL_COMPILER_OPTIONS="warn_missing_spec" mix compile --force
```

See: https://stackoverflow.com/questions/23600564/elixir-lint-for-confirming-that-every-function-has-type-sepcification

mgartner:
Is there anyway to remove the unnecessary warnings it produces for certain macros, like `warning: missing specification for function '__struct__'/0`.

PR #283, opened 2016-12-24
Fix regex matching tags and improvements.

nicolafiorillo:
I have found a regex bug matching tags: not indented lines were not correctly matched.
I took the opportunity to improve the same regex with more flexible patterns (spaces between # and tag, spaces between tag and the rest of the line).

Thanks.

### DIFF
--- a/lib/credo/check/design/tag_helper.ex
+++ b/lib/credo/check/design/tag_helper.ex
@@ -4,7 +4,7 @@ defmodule Credo.Check.Design.TagHelper do
   alias Credo.Check.CodeHelper
 
   def tags(source, tag_name) do
-    {:ok, regex} = Regex.compile("[^\?]?#\s*#{tag_name}:?\s*.+", "i")
+    {:ok, regex} = Regex.compile("(\A|[^\?])#\s*#{tag_name}:?\s*.+", "i")
 
     source
     |> CodeHelper.clean_strings_and_sigils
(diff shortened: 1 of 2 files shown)


lib/credo/check/design/tag_helper.ex
```diff
@@ -4,7 +4,7 @@ defmodule Credo.Check.Design.TagHelper do
   alias Credo.Check.CodeHelper
 
   def tags(source, tag_name) do
-    {:ok, regex} = Regex.compile("[^\?]# #{tag_name}:? .+", "i")
+    {:ok, regex} = Regex.compile("#\s*#{tag_name}:?\s*.+", "i")
```
rrrene: Why did you remove the leading `[^\?]`? It was meant as a safeguard against mistaking `?#` for a comment ...

nicolafiorillo: Match on (possible) `?#` restored.
But I did not get the point: why just `?#` and not any other char?
And writing ?# is not a syntax error, catch by compiler before credo? Trying to make a test about this, credo exits with a `** (RuntimeError) Source could not be parsed!`

rrrene: `?` can be used to describe a character (`?#` is equal to writing `'#'`, `?A` is equal to writing `'A'`).

We just want to make sure that the `#` we match is acutally indicating a comment.

Issue #241, opened 2016-11-24
MaxLineLength not running when priority set to normal


kelvinst:
### Environment

* Credo version (`mix credo -v`): 0.5.3
* Erlang/Elixir version (`elixir -v`): 1.3.4
* Operating system: Mac OSX Sierra

### What where you trying to do?

I have the following in my `.credo.exs`:

```
# ...
{Credo.Check.Readability.MaxLineLength, priority: :normal},
# ...
```

### Expected outcome

I was expecting `mix  credo` to execute the max line length check, without the need of `--strict`

### Actual outcome

It didn't run the max line length check. 😢

rrrene:
You are absolutly right. Thx for reporting! :+1:

rrrene:
I was not able to reproduce this, could you provide a sample where this breaks?

kelvinst:
Sorry, forgot the version up there, edited my original post.

And about more information, the only thing I did was to generate the config with `mix credo gen.config`, then I changed the line:

```
{Credo.Check.Readability.MaxLineLength, priority: :low, max_length: 80},
```

To:

```
{Credo.Check.Readability.MaxLineLength, priority: :normal},
```

Another important detail is that **my project is an umbrella project**.

kelvinst:
I'll try to attack this problem here. A PR coming soon. :wink:

rrrene:
That would be highly appreciated! :+1:

PR #289, opened 2016-12-30
WIP - Fixing the low priority exclusion

kelvinst:
Before this, the decision to execute or not the check was based only on the
`base_priority`, but we should read from the user config to decide this kind of
thing.


kelvinst:
I marked as WIP because I made no tests for this situation yet. @rrrene can you help me on where to put this tests, is there an existent test case for customized priority?

kelvinst:
BTW, this fixes #241.

rrrene:
Thx! :+1: `0.6.0-rc1` contains your fix!

PR #294, opened 2017-01-11
Add doctests for Credo.CLI.Filename

devonestes:
I noticed a while back that there are some functions that were untested
in the CLI part of the app, and I wanted to double back to add some
tests there. I noticed that some doctests had already been written for
`Credo.CLI.Filename`, so I decided to just run with that and turn the
doctests on for that module as part of the test suite.


rrrene:
@devonestes Great! Thx!

PR #301, opened 2017-01-14
Allow exclusion of file patterns per check

tsubery:
Not sure if this is something you are interested in. I find it useful to have different criteria for test files and the rest of the code. This comes from my usage of credo as a linter during development and not just before code commits.


zachdaniel:
:+1: This could be a temporary solution to our issues with https://github.com/rrrene/credo/issues/310.

evuez:
@rrrene Any plan on merging this? This'd be really useful!

PR #303, opened 2017-01-19
Add tests for `Credo.CLI.Switches` module

devonestes:
This module was untested, and even though it only has one public
function, and many of those functions are rather simple, there are a
couple things in there that I felt would be helped by having some test
coverage, so I went ahead and added that!


rrrene:
:+1: Thx @devonestes!

PR #307, opened 2017-01-24
Do not apply StringSigils rule to heredocs

### LINKED ISSUE #305: Proposal: StringSigils should not apply to heredocs
`StringSigils` should not apply to heredocs. There may be multiple quotes inside a heredoc, but there is no need to escape them. Therefore, `StringSigils` should not apply to heredocs.

joshprice:
Happy to try and implement a fix for this, adding a failing test for now.

Fixes #305


joshprice:
Since Elixir AST doesn't distinguish between a normal string literal and a heredoc this ended up being trickier than expected.

I ended up Regex matching the line the String literal is reported on, but I doubt this is an ideal fix. Improvements appreciated!

rrrene:
@joshprice Thx! :+1:

PR #314, opened 2017-02-02
Feature: Warn on unsafe `Map.get/2` calls inside pipes

narck:
Coming from Python this happens every now and then, especially when deserializing. You rarely know (or think!) whether the following function is able to handle `nil` values, and in most cases you run into a runtime at error forcing at minimum a recompile in dev and potentially dumb errors in production.

Mostly the first and consequent `n-1` pipe functions matter, nested expressions are also caught.

### DIFF
--- a/lib/credo/check/warning/map_get_unsafe_pass.ex
+++ b/lib/credo/check/warning/map_get_unsafe_pass.ex
@@ -1,0 +1,81 @@
+defmodule Credo.Check.Warning.MapGetUnsafePass do
+  @moduledoc """
+  `Map.get/2` can lead into runtime errors if the result is passed into a pipe without a proper default value.
+  This happens when the next function in the pipe cannot handle `nil` values correctly.
+
+  Example:
+
+        %{foo: [1, 2 ,3], bar: [4, 5, 6]}
+        |> Map.get(:baz)
+        |> Enum.each(&IO.puts/1)
+
+
+  This will cause a `Protocol.UndefinedError`, since `nil` isn't `Enumerable`. Often times while
+  iterating over enumerables zero iterations is preferrable to being forced to deal with an exception.
+  Had there been a `[]` default parameter this could have been averted.
+
+  Encourage use of `Map.get/3` instead to provide safer default values.
+  """
+
+  @explanation [check: @moduledoc]
+  @call_string "Map.get/2"
+
+  use Credo.Check, base_priority: :high
+
+  @doc false
+  def run(%SourceFile{} = source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:|>, meta , [left | right]} = ast, [] = issues, issue_meta) do
+    order = pipewalk(left) ++ pipewalk(List.first(right))
+
+    {ast, case is_unsafe?(order) do
+            true -> issues_for_call(meta, issues, issue_meta)
+            false -> issues
+          end}
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp pipewalk({:|>, _meta, [left | right]} = _ast) do
+    pipewalk(left) ++ pipewalk(List.first(right))
+  end
+
+  defp pipewalk({{:., _, [{:__aliases__, _, [namespace]}, callable]}, _, args}) do
+    [{"#{namespace}.#{callable}", args}]
+  end
+
+  defp pipewalk({value, _meta, args} = _ast) when is_atom(value),  do: [{value, args}]
+  defp pipewalk({value, _meta, []} = _ast),  do: [{value}]
+  defp pipewalk({value, _meta, nil} = _ast), do: [{value}]
+
+  defp is_unsafe?(expression) do
+    len = Enum.count(expression) - 1
+    {_, expression} = List.pop_at(expression, len)
+
+    true in Enum.map(expression, fn x ->
+                                   case x do
+                                     {"Map.get", [head | _]} when is_atom(head) -> true
+                                     {"Map.get", [head | tail]} when is_tuple(head) -> Enum.count(tail) < 2
+                                     {"Map.get", [_]} -> true
+                                     _ -> false
+                                   end
+                                 end)
+  end
+
+  def issues_for_call(meta, issues, issue_meta) do
+    [issue_for(issue_meta, meta[:line], @call_string) | issues]
+  end
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue issue_meta,
+      message: "Map.get/2 is unsafe in pipes, use Map.get/3",
+      trigger: trigger,
+      line_no: line_no
+  end
+end
--- a/lib/credo/check/warning/map_get_unsafe_pass.ex
+++ b/lib/credo/check/warning/map_get_unsafe_pass.ex
@@ -20,7 +20,7 @@ defmodule Credo.Check.Warning.MapGetUnsafePass do
   @explanation [check: @moduledoc]
   @call_string "Map.get/2"
 
-  use Credo.Check, base_priority: :high
+  use Credo.Check, base_priority: :normal
 
   @doc false
   def run(%SourceFile{} = source_file, params \\ []) do
--- a/lib/credo/check/warning/map_get_unsafe_pass.ex
+++ b/lib/credo/check/warning/map_get_unsafe_pass.ex
@@ -46,8 +46,8 @@ defmodule Credo.Check.Warning.MapGetUnsafePass do
     pipewalk(left) ++ pipewalk(List.first(right))
   end
 
-  defp pipewalk({{:., _, [{:__aliases__, _, [namespace]}, callable]}, _, args}) do
-    [{"#{namespace}.#{callable}", args}]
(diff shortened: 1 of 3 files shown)


lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  This will cause a `Protocol.UndefinedError`, since `nil` isn't `Enumerable`. Often times while
+  iterating over enumerables zero iterations is preferrable to being forced to deal with an exception.
+  Had there been a `[]` default parameter this could have been averted.
+
+  Encourage use of `Map.get/3` instead to provide safer default values.
+  """
+
+  @explanation [check: @moduledoc]
+  @call_string "Map.get/2"
```
rrrene: This is not the trigger. The trigger is the substring of code which triggers the issue (`Map.get` in this case).

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Credo.Check.Warning.MapGetUnsafePass do
+  @moduledoc """
+  `Map.get/2` can lead into runtime errors if the result is passed into a pipe without a proper default value.
```
rrrene: Please keep lines to 80 chars.

https://github.com/rrrene/elixir-style-guide#character-per-line-limit

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+  def run(%SourceFile{} = source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:|>, meta , [left | right]} = ast, [] = issues, issue_meta) do
+    order = pipewalk(left) ++ pipewalk(List.first(right))
+
+    pipe_issues = order
```
rrrene: Please format multi-line assignments like this: 

https://github.com/rrrene/elixir-style-guide#multi-line-call

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+                  |> unsafe_lines
+                  |> Enum.map(fn x -> issues_for_call([line: x || meta[:line]], issues, issue_meta) end)
```
rrrene: Please use parens for `fn`:

https://github.com/rrrene/elixir-style-guide#function-calling-parens

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp pipewalk({:|>, _meta, [left | right]} = _ast) do
+    pipewalk(left) ++ pipewalk(List.first(right))
+  end
+
+  defp pipewalk({{:., meta, [{:__aliases__, _, [_namespace]}, _callable]} = ast, _, args}) do
```
rrrene: I am not sure what `pipewalk` does. It seems to me that it collects relevant data in this one case, and all other cases below are basically meaningless. Am I wrong?

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+                  |> List.flatten
+
+    {ast, issues ++ pipe_issues}
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp pipewalk({:|>, _meta, [left | right]} = _ast) do
```
rrrene: Could `pipewalk` be renamed to something more descriptive?

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  defp pipewalk({{:., meta, [{:__aliases__, _, [_namespace]}, _callable]} = ast, _, args}) do
+    [{to_function_call_name(ast), args, meta[:line]}]
+  end
+
+  defp pipewalk({value, meta, args} = _ast) when is_atom(value),  do: [{value, args, meta[:line]}]
+  defp pipewalk({value, meta, []} = _ast),  do: [{value, [], meta[:line]}]
+  defp pipewalk({value, meta, nil} = _ast), do: [{value, [], meta[:line]}]
+  defp pipewalk(any_literal),               do: [{any_literal, [], nil}]
```
rrrene: Please do not align `do:` like this (but I think we will get rid of this section anyway, see above).

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+  defp unsafe_lines(expression) do
+    # The last expression inside the pipe doesn't really matter
+    [head | pipe] = Enum.drop(expression, -1)
+
+    unsafe_head = case head do
+                    {"Map.get", args, line} when length(args) != 3 -> [line]
+                    _ -> []
+                  end
+
+    tail_mapper = fn x, acc ->
```
rrrene: Why is this an anonymous function instead of a "real" private function?

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,96 @@
...
+                      _ -> []
+                    end
+                  end
+
+    unsafe_tail = Enum.reduce(pipe, [], tail_mapper)
+    unsafe_head ++ unsafe_tail
+  end
+
+  def issues_for_call(meta, issues, issue_meta) do
+    [issue_for(issue_meta, meta[:line], @call_string) | issues]
```
rrrene: `@call_string` will never match (see above).

test/credo/check/warning/map_get_unsafe_pass_test.exs
```diff
@@ -0,0 +1,95 @@
...
+
+    %{}
+    |> Map.get(:foo, [])
+    |> Enum.each(&IO.puts/1)
+
+end
+end
+""" |> to_source_file
+    |> refute_issues(@described_check)
+    end
```
rrrene: We are missing a very essential test case here. What about:

```
    %{}
    |> Map.get(:foo)
    |> some_function_which_might_handle_nil()
```

Is this really supposed to trigger an issue?

narck:
Thanks for the notes! I'll look into them ASAP and get back to you.

test/credo/check/warning/map_get_unsafe_pass_test.exs
narck: Well, no. At the very least I understand that this should not really be a warning. It would even be rather annoying if the developer is aware of how `Map.get/2` works and has prepared as such. 

Still, I think I'd much prefer or would use `Map.get/3` for being explicit. It shows very clearly to a reader the form of input which the next function expects, be it `nil` or else. This might be obvious to the developer working on their code, but you really can't know if a reader is going to understand that. That still would absolutely not count as a warning though.

So, following technically, I'd say it probably the simplest & cleanest fix for this would be to check if the next function in the pipe is a known `nil` accepting function. For example, this check could warn if the next function belonged to, say, `Enum` which cannot work with `nil`s (not 100% sure though, would have to verify).

I had some other thoughts too, but right now this made the most sense to me and I'd like to hear your ideas.

rrrene: I am more in favor of ensuring that the next call goes to `Enum` instead of trying to verify whether or not the receiving function accepts `nil`. The latter is not-so-easy(tm) if we want to account for functions created through `use` and `quote`.

Let's use a "blacklist" here.

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -34,32 +35,40 @@ defmodule Credo.Check.Warning.MapGetUnsafePass do
...
-        _ -> 2
-      end
+  defp find_pipe_issues(pipe, issue_meta) do
+    pipe
+    |> Enum.reduce([], fn ({expr, idx}, acc) ->
+         required_length = required_argument_length(idx)
+         {next_expr, _} = Enum.at(pipe, idx + 1, {nil, nil})
+
+         case {expr, nil_safe?(next_expr)} do
+           {{{{:., meta, [{_, _, [:Map]}, :get]}, _, args}, _}, false} when length(args) != required_length ->
```
narck: Felt this was clearer at a bit over 80 so I left it for now...

narck:
Changed the implementation as discussed.

lib/credo/check/warning/map_get_unsafe_pass.ex
```diff
@@ -0,0 +1,82 @@
...
+      {{{:., _, [{_, _, [module]}, _]}, _, _}, _} ->
+        !(module in @unsafe_modules)
+      _ ->
+        true
+    end
+  end
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue issue_meta,
+      message: "Map.get with no default return value is potentially unsafe
```
josevalim: The error message should also say: "if you are sure the value exists and can't be nil, please use `Map.fetch!/2`".

Issue #299, opened 2017-01-13
Proposal: detect expensive `Logger` calls that should be refactored


jmitchell:
### Environment

* Credo version: 0.5.3
* Erlang/Elixir version: Erlang/OTP 19; Elixir 1.4.0
* Operating system: Ubuntu 16.04.1 LTS

### What were you trying to do?

Detect any expensive calls to `Logger` that aren't wrapped in a zero-argument function (aka a thunk), and be advised to refactor.

### Expected outcome

This code

```elixir
Logger.debug "my_fun/1: input: #{inspect input}"
```

should be refactored to:

```elixir
Logger.debug fn ->
  "my_fun/1: input: #{inspect input}"
end
```

### Actual outcome

[This issue](https://github.com/jmitchell/backtrex/pull/1/files) wasn't detected during development.

jmitchell:
For more context on what prompted this, see ["Logging: a silent performance killer"](https://elixirforum.com/t/logging-a-silent-performance-killer/3258).

jmitchell:
After more reflection I realized there's room for improvement in Elixir itself. I submitted a [proposal](https://groups.google.com/d/msg/elixir-lang-core/LNZNyUGTdno/9Y0YOQ5tCQAJ) to elixir-lang-core. If accepted and implemented it may obviate the need for any work here.

josevalim:
Given the Elixir behaviour, if accepted, would be opt-in, I believe this suggestion has enough merit on its own. :+1:

rrrene:
@jmitchell I like this idea! If anybody wants to submit a PR on this, I would accept it :+1:

rrrene:
This has been implemented in #326. Thx for the initial suggestion! :+1:

PR #326, opened 2017-02-22
Ensures laziness of Logger calls

msoedov:
#299

### DIFF
--- a/lib/credo/check/warning/lazy_logging.ex
+++ b/lib/credo/check/warning/lazy_logging.ex
@@ -1,4 +1,4 @@
-defmodule Credo.Check.Warning.Logging do
+defmodule Credo.Check.Warning.LazyLogging do
   @moduledoc """
   Ensures laiziness of Logger calls.
   The best practice is to wrap an expensive logger expression into a zero argument function (fn -> "input" end)
--- a/lib/credo/check/warning/lazy_logging.ex
+++ b/lib/credo/check/warning/lazy_logging.ex
@@ -38,17 +38,17 @@ defmodule Credo.Check.Warning.LazyLogging do
     params = IssueMeta.params(issue_meta)
     levels = Params.get(params, :levels, @default_params)
     if Enum.member?(levels, level) do
-        {ast, issues_for_call(arguments, meta, state, issue_meta)}
+      {ast, issues_for_call(arguments, meta, state, issue_meta)}
     else
-        {ast, state}
+      {ast, state}
     end
   end
   defp traverse({:import, _meta, arguments} = ast, state, _issue_meta) do
     if logger_import?(arguments) do
-        {_, issue_list} = state
-        {ast, {true, issue_list}}
+      {_, issue_list} = state
+      {ast, {true, issue_list}}
     else
-        {ast, state}
+      {ast, state}
     end
   end
   defp traverse(ast, state, _issue_meta) do
@@ -64,8 +64,8 @@ defmodule Credo.Check.Warning.LazyLogging do
 
   defp issue_for(issue_meta, line_no) do
     format_issue issue_meta,
-        message: "Logger usage is not lazzy",
-        line_no: line_no
+      message: "Logger usage is not lazzy",
+      line_no: line_no
   end
 
   defp logger_import?([{:__aliases__, _meta, [:Logger]}]) do
--- a/lib/credo/check/warning/lazy_logging.ex
+++ b/lib/credo/check/warning/lazy_logging.ex
@@ -50,7 +50,7 @@ defmodule Credo.Check.Warning.LazyLogging do
     {ast, issues}
   end
 
-  def issues_for_call([{:fn, _, __}] = _args, _meta, issues, _issue_meta) do
+  def issues_for_call([{:fn, _, __}] = _args, _meta, issues, _issue_meta, _) do
     issues
   end
   def issues_for_call(_args, meta, {true, issues}, issue_meta) do
--- a/lib/credo/check/warning/lazy_logging.ex
+++ b/lib/credo/check/warning/lazy_logging.ex
@@ -1,7 +1,7 @@
 defmodule Credo.Check.Warning.LazyLogging do
   @moduledoc """
-  Ensures laiziness of Logger calls.
-  The best practice is to wrap an expensive logger expression into a zero argument function (fn -> "input" end)
+  Ensures laziness of Logger calls.
+  The best practice is to wrap an expensive logger calls into a zero argument function (fn -> "input" end)
   Example:
     Logger.debug fn ->
         "my_fun/1: input: #{inspect :expensive_to_compute}"
@@ -49,7 +49,7 @@ defmodule Credo.Check.Warning.LazyLogging do
 
   defp issue_for(issue_meta, line_no) do
     format_issue issue_meta,
-      message: "Logger input is not lazzy",
+      message: "Logger call is not lazzy",
       line_no: line_no
   end
 end
--- a/lib/credo/check/warning/lazy_logging.ex
+++ b/lib/credo/check/warning/lazy_logging.ex
@@ -18,11 +18,13 @@ defmodule Credo.Check.Warning.LazyLogging do
   @doc false
   def run(%SourceFile{} = source_file, params \\ []) do
     issue_meta = IssueMeta.for(source_file, params)
-    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+    state = {false, []} # Logger seen ?, list of issues
+    {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)
+    issues
   end
 
   defp traverse({{:., _, [{:__aliases__, _, [:Logger]}, _]}, meta, arguments} = ast, issues, issue_meta) do
-    {ast, issues_for_call(arguments, meta, issues, issue_meta)}
+    {ast, issues_for_call(arguments, meta, issues, issue_meta, logger_call: true)}
   end
   defp traverse({:debug, meta, arguments} = ast, issues, issue_meta) do
     {ast, issues_for_call(arguments, meta, issues, issue_meta)}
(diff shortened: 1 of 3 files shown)


rrrene:
@msoedov Thanks for taking the time to put this together! Unfortunately, this can't be merged in its current form. 😢 

The code, as I understand it, warns on *all* functions named `debug`, `info`, `warn` and `error`, regardless of whether or not they actually belong to the `Logger` module. I am not comfortable with Credo telling its users that *any* function named e.g. `warn` has to take an anonymous function as parameter.

Let's focus on the `debug` statement (and only if it is actually a call to `Logger.debug`) as suggested in #299.

msoedov:
Hi @rrrene do you think it would be worth to warn `debug` function call only if there is `import Logging` statement?

rrrene:
@msoedov Yep. If we see an `import Logger` statement, we could warn for `debug/1` :+1:

msoedov:
Done

josevalim:
I believe this will still warn on cases like this:

    Logger.warn automatically_generated_lazy_function(arg1, arg2)

Since Elixir is not a statically compiled language, it is impossible to know if that function returns an anonymous function or a string, therefore I don't believe we should warn to avoid false warnings.

This also seems to warn on cases like this:

    Logger.warn "foo bar"

which is completely fine and would actually be made worse if moved inside a function. Similarly, this case is fine:

    Logger.warn ["foo bar", another_argument]

while this may be less fine:

    Logger.warn "foo bar #{another_argument}"

My suggestion: warn if the argument being given to Logger is a `{:<<>>, _, [_ | _]}` at the AST level, which likely means it is a binary with interpolation inside.

rrrene:
@msoedov I see Josés point. Let's only raise an issue if there is interpolation happening :+1:

Let's also make sure we can configure which `Logger.<debug|info|warn|...>` calls should be reported. Default should be `[:debug, :info, :warn, :error]`.

msoedov:
@rrrene @josevalim 

That's a valid point, I'll remake this pr :)

msoedov:
@rrrene done, great suggestions

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+    state = {false, []} # Logger import seen ?, list of issues
+    {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)
+    issues
+  end
+
+  defp traverse({{:., _, [{:__aliases__, _, [:Logger]}, _]}, meta, arguments} = ast, state, issue_meta) do
+    {ast, issues_for_call(arguments, meta, state, issue_meta)}
+  end
+  defp traverse({level, meta, arguments} = ast, {true, _issues} = state, issue_meta) when level in @levels do
+    levels = Params.get(@default_params, :levels)
```
rrrene: You're retrieving the levels from `@default_params` instead of the configured params.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+    case Enum.member?(levels, level) do
```
rrrene: Please rewrite this as an `if`.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+     else
+        {ast, state}
+     end
+  end
+  defp traverse(ast, state, _issue_meta) do
+    {ast, state}
+  end
+
+  defp issues_for_call([{:<<>>, _, [_ | _]} | _] = _args, meta, {import?, issues}, issue_meta) do
+      {import?, [issue_for(issue_meta, meta[:line]) | issues]}
```
rrrene: Please fix indent here.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  #
+  # cases NOT raising issues
+  #
+
+  test "it should NOT report lazzy logging" do
+"""
+defmodule CredoSampleModule do
+  def some_function(parameter1, parameter2) do
+    Logger.debug fn ->
+        "A debug message: #{inspect 1}"
```
rrrene: Please fix indent.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+    |> refute_issues(@described_check)
+  end
+
+  test "it should NOT report imported :debug from Logger" do
+"""
+defmodule CredoSampleModule do
+  import Logger, only: debug
+  def some_function(parameter1, parameter2) do
+    debug fn ->
+        "Some message: #\{CredoSampleModule\}"
```
rrrene: Please fix indent.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+""" |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should NOT report for function call" do
+"""
+defmodule CredoSampleModule do
+
+  def message do
+      "Imma message"
```
rrrene: Please fix indent.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  #
+  # cases raising issues
+  #
+
+  test "it should report a violation for interpolated strings" do
+"""
+defmodule CredoSampleModule do
+
+  def some_function(parameter1, parameter2) do
+      var_1 = "Hello world"
```
rrrene: Please fix indent.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      Logger.warn "The module: #\{var1\}"
```
rrrene: Please fix indent.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  end
+end
+""" |> to_source_file
+    |> assert_issue(@described_check)
+  end
+
+  test "it should report a violation with imported :debug from Logger" do
+"""
+defmodule CredoSampleModule do
+  import Logger
```
rrrene: Please add empty line after this.

test/credo/check/warning/logging_test.exs
```diff
@@ -0,0 +1,109 @@
...
+  end
+end
+""" |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should NOT report imported :debug from Logger" do
+"""
+defmodule CredoSampleModule do
+  import Logger, only: debug
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+  defp traverse({level, meta, arguments} = ast, {true, _issues} = state, issue_meta) when level in @levels do
+    levels = Params.get(@default_params, :levels)
+    case Enum.member?(levels, level) do
+        true ->
+            {ast, issues_for_call(arguments, meta, state, issue_meta)}
+        _ ->
+            {ast, state}
+    end
+  end
+  defp traverse({:import, _meta, arguments} = ast, state, _issue_meta) do
```
rrrene: Please fix indent in this function.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
+defmodule Credo.Check.Warning.LazyLogging do
+  @moduledoc """
+  Ensures laziness of Logger calls.
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+  The best practice is to wrap an expensive logger calls into a zero argument function (fn -> "input" end)
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+  Example:
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+    Logger.info fn -> "expensive to calculate info" end
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+  Instead of:
```
rrrene: Please add empty line after this.

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -0,0 +1,74 @@
...
+      Logger.info "mission accomplished"
+
```
rrrene: Please remove empty line.

msoedov:
Hi @rrrene, could you take another look to my updated changes?

lib/credo/check/warning/lazy_logging.ex
```diff
@@ -31,37 +35,37 @@ defmodule Credo.Check.Warning.LazyLogging do
...
-    levels = Params.get(@default_params, :levels)
-    case Enum.member?(levels, level) do
-        true ->
-            {ast, issues_for_call(arguments, meta, state, issue_meta)}
-        _ ->
-            {ast, state}
+    params = IssueMeta.params(issue_meta)
+    levels = Params.get(params, :levels, @default_params)
+    if Enum.member?(levels, level) do
+        {ast, issues_for_call(arguments, meta, state, issue_meta)}
```
rrrene: The indent should be two spaces. Please check for the whole file again.

rrrene:
@msoedov Thx! :+1:

tc-nicholas:
Hi all. Is this still needed?
https://github.com/elixir-lang/elixir/commit/f730125b443515adfbe894b26e6bb6d97de6f73c
It seems it no longer matters whether you pass a zero-arg function or not.

Issue #250, opened 2016-11-28
Proposal: AppendSingleItemList


rrrene:
When building up large lists, it is faster to prepend than append. Therefore: It is sometimes best to prepend to the list during iteration and call Enum.reverse/1 at the end, as it is quite fast. 

```elixir
# we want to detect code like this:
list = list_so_far ++ [new_item]

# and suggest something like this:
list = [new_item] ++ list_so_far
...
Enum.reverse(list)
```

h4cc:
The "sometimes" seems to too unspecific: https://pragtob.wordpress.com/2016/06/16/tail-call-optimization-in-elixir-erlang-not-as-efficient-and-important-as-you-probably-think/

optikfluffel:
@PragTob maybe you could help with that one ^^

PragTob:
Heyho, thanks for the mention :)

So, prepend and reverse should be faster imo/afaik (although I'd prepend using `[item | rest]`seems more idiomatic). 

In the blog post, all "serious" suggestions use prepend, the one that uses append is by FAR the slowest. So I'm :+1: on the check, although I'd add that sometimes the order may not matter so one could even omit the `Enum.reverse/1` call.

It's just body recursion that may be faster/more memory efficient than tail-recursion, but that's a whole other topic :)

PS: still need to update that blog post with the new Benchee Interface :)

arjan:
I'm also willing to try to take this on. What should the module be called? Should this go in the "warnings" category? Of should we create a new one specifically focused on performance?

josevalim:
Sorry for being too late but I don't think this should be merged. `++` is harmful when done inside a loop, however, for one case scenarios, it is completely fine. The Erlang Performance Myths even has a section exactly about this, called "Operator ++ is always bad": http://erlang.org/doc/efficiency_guide/myths.html

Unless Credo can detect the ++ operator is being used in an recursive environment, I would avoid such check.

What is Credo policy for false positives and false negatives? Is a check worth adding if it has false positives and false negatives? If so, at what rate?

arjan:
Hmm yeah come to think of this.. It is of course possible to mute the check via `@lint`.
Detecting recursion is undoable, I think. (especially with modules calling each other (GenServers etc)).

arjan:
Maybe we can tailor the check to be very specific, e.g. just looking at `for` blocks and the `Enum` functions, to check if the last call of those blocks use tail appends?
[...]

PR #330, opened 2017-03-03
Emit warning when appending single item to a list

arjan:
Initial implementation of #250

### DIFF
--- a/lib/credo/check/warning/append_single_item.ex
+++ b/lib/credo/check/warning/append_single_item.ex
@@ -1,0 +1,42 @@
+defmodule Credo.Check.Warning.AppendSingleItem do
+  @moduledoc """
+  When building up large lists, it is faster to prepend than
+  append. Therefore: It is sometimes best to prepend to the list
+  during iteration and call Enum.reverse/1 at the end, as it is quite
+  fast.
+  """
+
+  @explanation [check: @moduledoc]
+
+  use Credo.Check, base_priority: :high
+
+  @doc false
+  def run(%SourceFile{} = source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  # [a] ++ b is OK
+  defp traverse(ast = {:++, _, [[_], _]}, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  # a ++ [b] is not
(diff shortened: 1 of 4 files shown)


test/credo/check/warning/append_single_item_test.exs
```diff
@@ -0,0 +1,51 @@
+defmodule Credo.Check.Warning.AppendSingleItemTest do
+  use Credo.TestHelper
+
+  @described_check Credo.Check.Warning.AppendSingleItem
+
+  test "it shoult NOT report appending 2 items" do
+"""
+defmodule CredoSampleModule do
+  def some_function(parameter1, parameter2) do
+    [parameter1] ++ [parameter2]
```
rrrene: Let's add this line `list ++ [parameter1, parameter2]` here for completeness sake (to make sure it does not report appending 2 items on a pre-existing list).

lib/credo/check/warning/append_single_item.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Credo.Check.Warning.AppendSingleItem do
```
rrrene: Let's put this check in the `Refactor` category.

lib/credo/check/warning/append_single_item.ex
```diff
@@ -0,0 +1,42 @@
...
+  @moduledoc """
+  When building up large lists, it is faster to prepend than
+  append. Therefore: It is sometimes best to prepend to the list
+  during iteration and call Enum.reverse/1 at the end, as it is quite
+  fast.
+  """
```
rrrene: Let's add an example:

```
When building up large lists, it is faster to prepend than append. Therefore: It is sometimes best to prepend to the list during iteration and call Enum.reverse/1 at the end, as it is quite fast.

Example:

    list = list_so_far ++ [new_item]

    # refactoring this can make the code faster

    list = [new_item] ++ list_so_far
    # ...
    Enum.reverse(list)
```

arjan: Done

test/credo/check/warning/append_single_item_test.exs
arjan: Done

lib/credo/check/warning/append_single_item.ex
arjan: Done

arjan:
I think it's ready like this.

rrrene:
@arjan Thx!

Issue #332, opened 2017-03-06
Proposal: Allow passing a list of files to `mix credo`


BachirC:
While trying to add credo analysis in my `pre-commit` git hook on staged files, I noticed that actually only the first file of the list was anaylzed. (Credo : `0.6.1`, Elixir : `1.4.1`, default `.credo.exs`)

So being able to do something like : `mix credo [options] file1.exs file2.ex file3.ex` would be really nice imo ! :)

PR #334, opened 2017-03-08
Allow passing a list of files to `mix credo`

tzzzoz:
Initial implementation for #332


rrrene:
@tzzzoz Thanks for putting this together. :+1:

Unfortunately, this PR has far reaching consequences, as it allows me to analyse multiple directories, which could contain multiple `.credo.exs` files, which might have consequences I can not foresee in the moment.

I will have to think about this, but this is not likely to be included before `v1.0`. But your PR is very helpful in answering these questions with a present implementation. Thx!

tzzzoz:
Thanks for your feedback, I understand your concern, let me know if I can help.

jacknoble:
@rrrene Perhaps there should be a cli option to specify a path to a `.credo.exs` file (`mix credo --config_path=~/.credo.exs`). This makes sense as a feature by itself but it could also be used to support running credo on a list of files. And running on a list of files could be prohibited without supplying a `config_path`.

rrrene:
@jacknoble Not sure about that. So it would use the `:files` from that `.credo.exs`, but execute in the current directory?

jacknoble:
That would be the idea. The :files thing could be confusing, I guess.

rrrene:
Hi, thx again for this. Unfortunately, I am closing it due to "it being obsolete" IMHO. 

PR #336, opened 2017-03-08
Prevent false positive when initial value to pipe is for loop

### LINKED ISSUE #323: Pipeline raw value check fails on for loops
### Environment

* Credo version (`mix credo -v`): "0.5.3", "0c405b36e7651245a8ed63c09e2d52c2e2b89b6d02b1570c4d611e0fcbecf4a2"
* Erlang/Elixir version (`elixir -v`): 1.3.4
* Operating system: OSX

```elixir
for i <- 1..4 do
  i
end
|> Enum.map(fn(i) -> i + 2 end)
```
or (out of desperation)
```elixir
(i <- 1..4)
|> for do
  i
end
|> Enum.map(fn(i) -> i + 2 end)
```
### Expected outcome

No warning

### Actual outcome

Warning of pipeline starting without raw value

mitchellhenke:
Fixes #323

### DIFF
--- a/lib/credo/check/refactor/pipe_chain_start.ex
+++ b/lib/credo/check/refactor/pipe_chain_start.ex
@@ -31,7 +31,7 @@ defmodule Credo.Check.Refactor.PipeChainStart do
   end
 
 
-  for atom <- [:%, :%{}, :.., :<<>>, :@, :__aliases__, :unquote, :{}, :&, :<>, :++, :--, :&&, :||, :-] do
+  for atom <- [:%, :%{}, :.., :<<>>, :@, :__aliases__, :unquote, :{}, :&, :<>, :++, :--, :&&, :||, :-, :for, :<-] do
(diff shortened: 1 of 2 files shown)


rrrene:
@mitchellhenke :+1:

lib/credo/check/refactor/pipe_chain_start.ex
```diff
@@ -31,7 +31,7 @@ defmodule Credo.Check.Refactor.PipeChainStart do
   end
 
 
-  for atom <- [:%, :%{}, :.., :<<>>, :@, :__aliases__, :unquote, :{}, :&, :<>, :++, :--, :&&, :||, :-] do
+  for atom <- [:%, :%{}, :.., :<<>>, :@, :__aliases__, :unquote, :{}, :&, :<>, :++, :--, :&&, :||, :-, :for, :<-] do
```
josevalim: We should probably add `with` to the list then (for the same reason we added `for`).

Issue #337, opened 2017-03-14
Credo --strict throws exception with macro-created zero arity methods


mcelaney:
### Environment

* Credo 0.6.1
* Erlang/OTP 19 [erts-8.1] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false]
* Elixir 1.3.4
* OS/X 10.12.3

### What were you trying to do?
Consider this hello_world project: https://github.com/mcelaney/CredoBugExample

I have a macro (lib/hello_world/module_attribute.ex) that creates a zero arity method.  I import it in web.ex and then use it in my User modeule.

### Expected outcome
Method I want is created. Tests pass. Credo passes. Great success.

### Actual outcome
Problem is that if I run Credo in Strict mode I get an exception:

```
$ mix credo --strict
Checking 16 source files ...
Error while running Elixir.Credo.Check.Readability.NoParenthesesWhenZeroArity on lib/hello_world/module_attribute.ex
** (EXIT from #PID<0.70.0>) an exception was raised:
    ** (Protocol.UndefinedError) protocol String.Chars not implemented for {:unquote, [line: 33], [{:attribute, [line: 33], nil}]}
        (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
        (elixir) lib/string/chars.ex:17: String.Chars.to_string/1
        lib/credo/check/readability/no_parentheses_when_zero_arity.ex:46: Credo.Check.Readability.NoParenthesesWhenZeroArity.remaining_line_after/3
        lib/credo/check/readability/no_parentheses_when_zero_arity.ex:34: Credo.Check.Readability.NoParenthesesWhenZeroArity.issues_for_definition/3
        lib/credo/check/readability/no_parentheses_when_zero_arity.ex:22: Credo.Check.Readability.NoParenthesesWhenZeroArity.traverse/3
        (elixir) lib/macro.ex:296: anonymous fn/4 in Macro.do_traverse_args/4
        (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
        (elixir) lib/enum.ex:1247: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3

18:19:56.736 [error] Task #PID<0.181.0> started from #PID<0.70.0> terminating
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:unquote, [line: 33], [{:attribute, [line: 33], nil}]}
[...]

PR #338, opened 2017-03-16
BugFix: Crashes on zero arity function in macro

mcelaney:
PR related to https://github.com/rrrene/credo/issues/337 

Credo.Check.Readability.ParenthesesOnZeroArityDefs crashes if a macro defines a zero-arity function. 

- [x] Add a test to demonstrate the issue.
- [x] Fix the issue.

Credo.Check.Readability.ParenthesesOnZeroArityDefs checks zero arity methods and flags if they have parenthesis.  When generating a zero arity method in a macro we are required to add the parens.  This causes an issue because issues_for_definition/3 sees that args are == 0 but there is no name defined for the method yet (we just have something like ({:unquote, [line: 4], [{:attribute, [line: 4], nil}]}`)... so we blow up.

In theory this check needs to be skipped in this case - as the parens are required to generate the method.  As such I just skip the issue anytime this comes up.

### DIFF
--- a/lib/credo/check/readability/parentheses_on_zero_arity_defs.ex
+++ b/lib/credo/check/readability/parentheses_on_zero_arity_defs.ex
@@ -42,6 +42,10 @@ defmodule Credo.Check.Readability.ParenthesesOnZeroArityDefs do
     {ast, issues}
   end
 
+  # skip the false positive for a metaprogrammed definition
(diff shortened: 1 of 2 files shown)


rrrene:
@mcelaney LGTM, thx! Will merge soon.

PR #342, opened 2017-03-19
Set unicode option for regex at UI.wrap_at

mururu:
The regular expression used in `UI.wrap_at` can break unicode characters into invalid binaries as unicode. Some functions using the result like IO.puts raise exception.

```
iex(1)> Credo.CLI.Output.UI.wrap_at("あ", 1)
[<<227>>, <<129>>, <<130>>]
```

This patch doesn't care about display width of unicode characters.


rrrene:
@mururu LGTM, thx! :+1:

Issue #324, opened 2017-02-16
Proposal: Warn when missing spaces after commas (lists, params, etc)


christiannelson:
### Environment

* Credo version (`mix credo -v`): 0.6.1
* Erlang/Elixir version (`elixir -v`): 1.4.1
* Operating system: macOS 10.2.4

### What were you trying to do?

Add a new check that will report when spaces are missing after parameters in a variety of contexts.

Example:

```
defmodule Module do
  alias Project.{A,B} # WARN: missing space after comma

  @attr [1,2,3,4,5] # WARN:  missing space after comma

  def a_func(first_name,last_name) do # WARN: missing space after comma
    Map.merge(%{a: first_name,b: last_name},%{a: 3, d: 4}) # WARN: missing space after comma
  end
end
```

FWIW, I currently can't think of any exceptions where a commas separates values or params that the check would not trigger the code style check.

rrrene:
Yep, this seems like a good idea. Not sure if this *has* to be a consistency check. WDYT?

christiannelson:
> Not sure if this has to be a consistency check. WDYT?

Sorry, I'm not exactly sure what you mean with the question.

I would describe this check as something teams might only care about if they're detailed oriented, or pedantic about consistency. Perhaps it's disabled at first and folks can opt-in?

I hope to take a shot at implementing this -- and possibly the other proposal I sent -- one of these days when I have some free time. TBH, it might be a while though. If you (or anyone else) feels the inspiration to do it, I certainly won't be upset about it. :)

rrrene:
> Sorry, I'm not exactly sure what you mean with the question.

In Credo, there is a concept called "concistency checks", which are not checking a specific behaviour (use tabs for indentation) but rather look at your project and examine it for consistency ("you seem to be using tabs for the majority of your code, but in these 5 lines use spaces").

What I wanted to say is that your proposal *might* fit that scheme, but I'm not sure.

christiannelson:
@rrrene Oh yeah, I totally know what you mean. I'll think about it some more, I could see it going either way.

Do consistency checks work across files, or do they look for consistency within the current file only?

rrrene:
They work on all files. But - after some introspection - I think this check might just be a regular one, which is also easier to implement ;-)

PR #344, opened 2017-03-22
Add check for spaces after commas

jacknoble:
Creates a new readability check SpaceAfterCommas and resolves https://github.com/rrrene/credo/issues/324 (as long as we're good with a readability check).

Let me know if I've missed something or my Elixir is rusty and I'll fix it :-)

### DIFF
--- a/lib/credo/check/readability/space_after_commas.ex
+++ b/lib/credo/check/readability/space_after_commas.ex
@@ -1,0 +1,56 @@
+defmodule Credo.Check.Readability.SpaceAfterCommas do
+  @moduledoc """
+  Don't use spaces after commas.
+
+      # good
+      alias Project.{Alpha, Beta}
+      arr = [1, 2, 3, 4, 5]
+      def some_func(first, second, third) do
+        ...
+      end
+
+      # hard to read
+      alias Project.{Alpha,Beta}
+      arr = [1,2,3,4,5]
+      def some_func(first,second,third) do
+        ...
+      end
+
+  To make your code more readable put spaces or newlines after your commas.
+  """
+
+  @explanation [check: @moduledoc]
+  @unspaced_commas ~r/(\,\S)/
+
+  use Credo.Check, base_priority: :low
+  alias Credo.Check.CodeHelper
+  alias Credo.Code
+
+  @doc false
+  def run(source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    source_file
+    |> CodeHelper.clean_strings_sigils_and_comments
+    |> Code.to_lines
+    |> Enum.flat_map(&find_issues(issue_meta, &1))
+  end
+
+  defp issue_for(issue_meta, trigger, line_no, column) do
+    format_issue issue_meta,
+      message: "Space missing after comma",
+      trigger: trigger,
+      line_no: line_no,
+      column: column
+  end
+
+  defp find_issues(issue_meta, {line_no, line}) do
+    @unspaced_commas
+    |> Regex.scan(line, capture: :all_but_first, return: :index)
+    |> List.flatten
+    |> Enum.map(fn({idx, len}) ->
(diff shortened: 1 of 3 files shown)


test/credo/check/readability/space_after_commas_test.exs
```diff
@@ -0,0 +1,57 @@
...
+end
+""" |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  #
+  # cases raising issues
+  #
+
+  test "it should report when commas are not followed by spaces" do
```
rrrene: Good test suite. :+1:

Two cases are missing:

```elixir
# should NOT trigger an issue
[?,, ?:]

# should trigger an issue
[?,,?:]
```

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Credo.Check.Readability.SpaceAfterCommas do
+  @moduledoc """
+  Don't use spaces after commas.
```
rrrene: Let's write:

> You can use white-space after commas to make items of lists, tuples and other enumerations easier to separate from one another.

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
...
+
+      # good
+      alias Project.{Alpha, Beta}
+      arr = [1, 2, 3, 4, 5]
+      def some_func(first, second, third) do
+        ...
+      end
+
+      # hard to read
```
rrrene: Let's write:

> NOT preferred - items are harder to separate

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Credo.Check.Readability.SpaceAfterCommas do
+  @moduledoc """
+  Don't use spaces after commas.
+
+      # good
```
rrrene: Let's write:

> preferred

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      # hard to read
+      alias Project.{Alpha,Beta}
+      arr = [1,2,3,4,5]
+      def some_func(first,second,third) do
+        ...
+      end
+
+  To make your code more readable put spaces or newlines after your commas.
```
rrrene: There a "standard footer" for Readability issues, which we should include here:

>  Like all `Readability` issues, this one is not a technical concern.
>  But you can improve the odds of others reading and liking your code by making
>  it easier to follow.

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
...
+  """
+
+  @explanation [check: @moduledoc]
+  @unspaced_commas ~r/(\,\S)/
+
+  use Credo.Check, base_priority: :low
```
rrrene: Let's remove the `, base_priority: :low` part and make this a first-class citizen. :+1:

lib/credo/check/readability/space_after_commas.ex
```diff
@@ -0,0 +1,56 @@
...
+      arr = [1,2,3,4,5]
+      def some_func(first,second,third) do
+        ...
+      end
+
+  To make your code more readable put spaces or newlines after your commas.
+  """
+
+  @explanation [check: @moduledoc]
+  @unspaced_commas ~r/(\,\S)/
```
rrrene: I'm pretty sure this has to be `~r/[^\?](\,\S)/`. See additional test cases below!

jacknoble:
@rrrene Updated. Good call on `?` -- modified further because we should still warn when the comma comes after a question mark in a variable or function name and because it needs to use look behind to capture each violation.

rrrene:
@jacknoble Great! Thx for this! :+1:

Issue #349, opened 2017-03-24
read-from-stdin with an annotation does not work


dvzubarev:
### Environment

* Credo version `0.7.0`:
* Erlang/Elixir version `1.40`:
* Operating system:
Ubuntu 16.04
### What were you trying to do?
cat lib/test.ex |  mix credo --format flycheck --read-from-stdin lib/test.ex

### Expected outcome
lib/test.ex:1:11: R: Modules should have a @moduledoc tag.
The filename is at the beginning.

### Actual outcome
stdin:1:11: R: Modules should have a @moduledoc tag
Stdin is at the beginning.

I found that it worked at version 0.6.1. 
There was only one commit from 0.6.1: b8d5ec34ff3a10f8958c99561e36a2a30cd25361
When I reverted it, I got expected outcome again.

dvzubarev:
After reverting and testing I found another case when annotating does not work. 
### What were you trying to do?
cat /full/path/to/lib/test.ex |  mix credo --format flycheck --read-from-stdin /full/path/to/lib/test.ex


### Expected outcome
lib/test.ex:1:11: R: Modules should have a @moduledoc tag.
The filename is at the beginning. It works like this in version 0.6.1

### Actual outcome
stdin:1:11: R: Modules should have a @moduledoc tag
Stdin is at the beginning.

PR #350, opened 2017-03-25
Fixed a pattern match that was preventing flycheck from being formatted correctly

benbotwin:
This should fix issue #349


rrrene:
Ouch, thx!

Issue #347, opened 2017-03-23
Credo unable to find files in windows


nikolauska:
### Environment

* Credo version (`mix credo -v`): 0.7.1
* Erlang/Elixir version (`elixir -v`): 1.4.1
* Operating system: Windows 10

### What were you trying to do?
Run `mix credo --strict` on umbrella project.

.credo.exs files
```
files: %{
  included: ["apps/"],
  excluded: [~r"/_build/", ~r"/deps/"]
},
```

### Expected outcome
Credo starts checking files same way as in linux machines.

### Actual outcome
```
No files found!
Please report incorrect results: https://github.com/rrrene/credo/issues
Analysis took 0.01 seconds (0.01s to load, 0.00s running checks)
0 mods/funs, found no issues.
```

GeorgeKaraszi:
I too am seeing the same exact issue.

rrrene:
I've never used Elixir on Windows. Is this possibly due to the difference of `/` vs. `\` on Linux and Windows?

kaiyote:
Super new to Elixir, and would love to have this work on windows, so I went snooping around w/ the debugger (first time for that too) and I might have found something?

It looks like you get a path (which seems to be correct: running `mix credo` in `c:\git\my_project` gets you a path of `c:\git\my_project`), and then pass it to `Filename.remove_line_no_and_column`, which splits on `:` and returns the first chunk of the path.

Problem is, windows paths (at least full paths), have colons in them, so in the above example you end up with an `include` in the config it generates of just `c`.

No idea how to fix it, just thought I might help you track this down.

nvanmeurs:
Same issue here, looks like it's a path problem?

rrrene:
@kaiyote Thx for the analysis! Could you please retry with Credo `v0.7.2`?

nvanmeurs:
@rrrene Just tried to run mix credo using v0.7.2, still `No files found!`
Elixir v1.4.1
Erlang/OTP 19

GeorgeKaraszi:
@nvanmeurs In case you haven't figured it out yet. They'll need to cut a new release with the merged bug fix.

Until then do
`{:credo, only: [:dev, :test], git: "https://github.com/rrrene/credo.git"}`

nikolauska:
I've also verified that merge #352 does fix this issue.

rrrene:
@nikolauska @GeorgeKaraszi @nvanmeurs Excellent. Will release a new version soon!

PR #352, opened 2017-03-28
turns out in Elixir windows uses '/' instead of '\'

kaiyote:
Pretty sure this solves #347 .  After making the change, fired up `iex` and ran `Credo.CLI.main([])`. It found 141 source files, and this being windows, I feel like that's an improvement.

I also used this branch as the resolver for another Elixir project on my box, and ran that one with `mix credo` and it found the files.....


rrrene:
@kaiyote Thx! :+1:

Issue #144, opened 2016-08-07
- in macros are treated as operations


AndrewDryga:
This code produces Consistency errors, but there are no operations in it:

```
defmodule BSON.BinaryUtils do
  @moduledoc false

  defmacro int64 do
    quote do: signed-little-64
  end

  defmacro int32 do
    quote do: signed-little-32
  end

  defmacro int16 do
    quote do: signed-little-16
  end

  defmacro uint16 do
    quote do: unsigned-little-16
  end

  defmacro int8 do
    quote do: signed-little-8
  end

  defmacro float64 do
    quote do: float-little-64
  end

  defmacro float32 do
    quote do: float-little-32
  end

  defmacro binary(size) do
    quote do: binary-size(unquote(size))
  end
end
```

Result:

```
~/Projects/www/bsoneach(features/perfomance ✗) mix credo -a
Checking 8 source files ...

  Consistency                                                                   
┃ 
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:33:21 (BSON.BinaryUtils.binary)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:5:21 (BSON.BinaryUtils.int64)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:5:28 (BSON.BinaryUtils.int64)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:9:21 (BSON.BinaryUtils.int32)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:9:28 (BSON.BinaryUtils.int32)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:13:21 (BSON.BinaryUtils.int16)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:13:28 (BSON.BinaryUtils.int16)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:17:23 (BSON.BinaryUtils.uint16)
┃ [C] ↗ There are spaces around operators most of the time, but not here.
┃       lib/bson/binary_utils.ex:17:30 (BSON.BinaryUtils.uint16)
[...]

PR #362, opened 2017-04-17
Fix false positives in SpacesAroundOperators

sevenseacat:
From #144, I was still having issues with false positives being reported due to my use of other keywords not listed in the check.


rrrene:
@sevenseacat Thx :+1:

PR #364, opened 2017-04-26
fix(sources): support command line filename args when reading from stdin

russmatney:
The behavior as described in #349 is still exactly as described before this [fix](https://github.com/rrrene/credo/commit/947efb05fdf57b352068fb0d8e7164d761eea051).

To reiterate #349, commands like:

```
mix credo --strict --format flycheck --read-from-stdin /path/to/lib/file.ex < /path/to/lib/file.ex
```

Return: 

```
stdin:20:1: R: There should be no trailing white-space at the end of a line.
```

Instead of:

```
/path/to/lib/file.ex:20:1: R: There should be no trailing white-space at the end of a line.
```

Which causes things like flycheck to be unable to parse and show the errors inline.

### DIFF
--- a/lib/credo/sources.ex
+++ b/lib/credo/sources.ex
@@ -20,6 +20,11 @@ defmodule Credo.Sources do
     |> source_file_from_stdin()
     |> List.wrap
   end
+  def find(%Credo.Config{read_from_stdin: true, args: [filename]}) do
+    filename
+    |> source_file_from_stdin()
+    |> List.wrap
+  end
   def find(%Credo.Config{read_from_stdin: true}) do
     @stdin_filename
     |> source_file_from_stdin()


rrrene:
```bash
$ mix credo --strict --format flycheck --read-from-stdin lib/credo.ex < lib/credo.ex
Compiling 1 file (.ex)
lib/credo.ex:1:11: R: Modules should have a @moduledoc tag.
```

I wonder why this is different on my machines (both Linux and MacOS)?

russmatney:
@rrrene After some more digging, i've found that it does indeed work when you use paths relative to the project, but fails if you use an absolute path.

I created a quick project here if it helps you reproduce it: [repo with strict credo errors](https://github.com/russmatney/credo_error_reproduction).

This does work: 

```
mix credo --strict --format flycheck --read-from-stdin ./lib/test_app.ex < ./lib/test_app.ex
```

This returns `stdin` as where the file path should be: 

```
mix credo --strict --format flycheck --read-from-stdin ~/projects/test_app/lib/test_app.ex < ~/projects/test_app/lib/test_app.ex
```

The patch here fixes this, but now it seems like this is a different issue – I can dig into what the root might be if you'd prefer a different solution.

rrrene:
Thx for the analysis. :+1: 

Could you retest with master (setting the dep to `{:credo, github: "rrrene/credo"}`)?

russmatney:
Nice dude, working great from master for both normal and umbrella apps. Thanks for the fix!

Issue #373, opened 2017-05-09
Missing space after comma warning triggered by comma inside single quotes


mikestok:
### Environment

```
$ mix credo -v
0.7.4
$ elixir -v
Erlang/OTP 19 [erts-8.3] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.4.2
$ uname -a
Darwin ratdog.local 16.5.0 Darwin Kernel Version 16.5.0: Fri Mar  3 16:52:33 PST 2017; root:xnu-3789.51.2~3/RELEASE_X86_64 x86_64
```

### What were you trying to do?

I am trying to use the new spaces after comma check to check our codebase, and this simplified example triggers the _Space missing after comma error_ when I run mix `credo --strict`.

```
defmodule CredoComma do
  @moduledoc false

  @doc false
  def hello do
    # The ' here causes a "Space missing after comma" warning when
    # you run `mix credo --strict`.
    ','
  end
end
```

Little repo at https://github.com/mikestok/credo_comma

I have not had time to dig into this yet, reporting it so that it doesn't get forgotten.

### Expected outcome

No issues

### Actual outcome

```
$ mix credo --strict
Checking 1 source file ...

  Code Readability
┃
┃ [R] → Space missing after comma
┃       lib/credo_comma.ex:8:6 #(CredoComma.hello)

Please report incorrect results: https://github.com/rrrene/credo/issues

Analysis took 0.03 seconds (0.00s to load, 0.03s running checks)
2 mods/funs, found 1 code readability issue.
```

rrrene:
Hi Mike, thx for this bug report. 

Could you retest this against `master` (setting the dep to `{:credo, github: "rrrene/github"}`)?

mikestok:
It's still producing the same error at 852be8a, current head of master. I will have a look at it today.

```
✔ ~/Projects/credo_comma [master|✚ 2]
08:58 $ mix credo --strict
Checking 1 source file ...

  Code Readability
┃
┃ [R] → Space missing after comma
┃       lib/credo_comma.ex:8:6 #(CredoComma.hello)

Please report incorrect results: https://github.com/rrrene/credo/issues

Analysis took 0.02 seconds (0.00s to load, 0.02s running checks)
2 mods/funs, found 1 code readability issue.

✘-4 ~/Projects/credo_comma [master|✚ 2]
08:58 $ grep credo mix.*
[...]

PR #374, opened 2017-05-10
Make space after comma check use tokens to analyse code #373

mikestok:
The regular expression version of this check was "confused" by code like
that added in the one of the new tests in this commit. The new code catches some occurrences of missing space which were missed by the old code (for example `,[`)

This uses the "real" tokenizer to scan the source code, and checks that
any token following a comma on the same line starts after the next
character (which would have been whitespace dropped by the tokenizing.)

The build failure seems to be the same as master, so no known new problems are introduced to the code by this commit.

On our code base at work this seems to provide a *small* speed up when running `mix credo --strict`.

### DIFF
--- a/lib/credo/check/readability/space_after_commas.ex
+++ b/lib/credo/check/readability/space_after_commas.ex
@@ -30,23 +30,43 @@ defmodule Credo.Check.Readability.SpaceAfterCommas do
 
   # Matches commas followed by non-whitespace unless preceded by
   # a question mark that is not part of a variable or function name
-  @unspaced_commas ~r/(?<!\W\?)(\,\S)/
 
   use Credo.Check
-  alias Credo.Check.CodeHelper
-  alias Credo.Code
+  alias Credo.IssueMeta
+  alias Credo.SourceFile
 
   @doc false
   def run(source_file, params \\ []) do
     issue_meta = IssueMeta.for(source_file, params)
 
-    source_file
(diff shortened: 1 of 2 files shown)


lib/credo/check/readability/space_after_commas.ex
```diff
@@ -30,37 +30,43 @@ defmodule Credo.Check.Readability.SpaceAfterCommas do
...
   def run(source_file, params \\ []) do
     issue_meta = IssueMeta.for(source_file, params)
 
-    source_file
-    |> CodeHelper.clean_charlists_strings_sigils_and_comments
-    |> Code.to_lines
-    |> Enum.flat_map(&find_issues(issue_meta, &1))
+    source_file.source
+    |> Credo.Code.to_tokens
+    |> Enum.chunk(2, 1)
```
mikestok: This "drops" the last token we find, in this case I don't think a valid Elixir source can end with a `,`. If we wanted to guarantee that we examined the last token we'd have to pad the chunking, maybe:

    |> Enum.chunk(2, 1, [{nil, {nil, nil, nil}}]) # dummy is right shape for collect_issues

I don't think it's worth doing this, what does @rrrene think?

rrrene:
@mikestok I can't merge this, but took your new test case to fix the underlying problem of the existing check.

Thank you very much for your steady contributions and support! :+1:

Issue #359, opened 2017-04-08
excluded list not working as expected


nathanhornby:
### Precheck

Can't seem to find anything on this issue, I assume it must be user error but I can't see where…

Here is my config:

```elixir
%{
  configs: [
    %{
      files: %{
        included: ["lib/"],
        excluded: [~r"/_build/", ~r"/deps/", ~r"/doc/"]
      },
      checks: [
        {Credo.Check.Readability.MaxLineLength, priority: :low, max_length: 80, ignore_string: true, ignore_specs: true}
      ]
    }
  ]
}
```

But on running `mix credo` all files are checked, included those in the 'excluded' directories.

### Environment

* Credo version (`mix credo -v`): 0.7
* Erlang/Elixir version (`elixir -v`): 1.4
* Operating system: OSX 10.11.6

### What were you trying to do?

Exclude directories.

### Expected outcome

Directories are excluded from the code analysis.

### Actual outcome

The directories are included in the code analysis.

rrrene:
Thx for the bug report! :+1: 

Your config is missing the required "name" field. I created an issue for that.

rrrene:
@nathanhornby To elaborate:

Just put `name: "default",` above the `files: %{` line and you're fine!

nathanhornby:
@rrrene

Ah thank you! Nice and simple. I must have misunderstood this line from the packages own .credo.exs:

> If no config name is given "default" is used.

I assumed this to mean it was an optional attribute.

rrrene:
@nathanhornby I thought about making it take the only config present (if there is only one), but then I remembered that I like Elixir's motto of "explicitness over implicit behaviour" very much and decided against it. So you're not crazy 😆  👍

PR #379, opened 2017-05-18
Add check for malformed .credo.exs files

### LINKED ISSUE #353: Proposal: Error when `.credo.exs` contains non-existent checks
Running Credo with a `.credo.exs` listing non-existent checks (either due to misspelling or versioning. 

### Expected outcome
A clear error message with the location of the config file, the name of the check that does not exist and a line number.

### Actual outcome
UndefinedFunctionError https://github.com/rrrene/credo/issues/351

### LINKED ISSUE #360: Proposal: Warn about malformed `.credo.exs` files

I investigated a bug report (#359) which contained a sample config.

### Expected outcome

Credo could have warned me that the config in question was missing the required "name" field.

### Actual outcome

Credo ignored the config and showed seemingly weird behavior.

devonestes:
Resolves #360
Resolves #353

### DIFF
--- a/lib/credo/exs_loader.ex
+++ b/lib/credo/exs_loader.ex
@@ -1,33 +1,4 @@
 defmodule Credo.ExsLoader do
-  def validate(parsed_file) do
-    if !Enum.all?(parsed_file[:configs], &do_validate/1) do
-      IO.puts("One or more of your .credo.exs files are invalid")
-    end
-    parsed_file
-  end
-
-  defp do_validate(%{name: name} = config) when is_binary(name) do
-    validate_requires(config) && validate_checks(config)
-  end
-  defp do_validate(_), do: false
-
-  defp validate_requires(%{requires: requires}) when is_list(requires) do
-    if Enum.all?(requires, &File.exists?/1) do
-      Enum.each(requires, &Code.require_file/2)
-      true
-    else
-      false
-    end
-  end
-  defp validate_requires(_), do: true
-
-  defp validate_checks(%{checks: checks}) when is_list(checks) do
-    Enum.all?(checks, fn(tuple) ->
-      tuple |> elem(0) |> Code.ensure_compiled?
-    end)
-  end
-  defp validate_checks(_), do: true
-
   def parse(exs_string, safe \\ false)
   def parse(exs_string, true) do
     case Code.string_to_quoted(exs_string) do
--- a/lib/credo/exs_loader.ex
+++ b/lib/credo/exs_loader.ex
@@ -1,4 +1,14 @@
 defmodule Credo.ExsLoader do
+  def validate(parsed_file) do
+    if !Enum.all?(parsed_file[:configs], &do_validate/1) do
+      IO.puts("One or more of your .credo.exs files are invalid")
+    end
+    parsed_file
+  end
+
+  defp do_validate(%{check_for_updates: _, checks: _, color: _, files: _, name: _, requires: _, strict: _}), do: true
+  defp do_validate(_), do: false
+
   def parse(exs_string, safe \\ false)
   def parse(exs_string, true) do
     case Code.string_to_quoted(exs_string) do
--- a/lib/credo/exs_loader.ex
+++ b/lib/credo/exs_loader.ex
@@ -6,8 +6,27 @@ defmodule Credo.ExsLoader do
     parsed_file
   end
 
-  defp do_validate(%{check_for_updates: _, checks: _, color: _, files: _, name: _, requires: _, strict: _}), do: true
+  defp do_validate(%{name: name} = config) when is_binary(name) do
+    validate_requires(config) && validate_checks(config)
+  end
   defp do_validate(_), do: false
+
+  defp validate_requires(%{requires: requires}) when is_list(requires) do
+    if Enum.all?(requires, &File.exists?/1) do
+      Enum.each(requires, &Code.require_file/2)
+      true
+    else
+      false
+    end
+  end
+  defp validate_requires(_), do: true
+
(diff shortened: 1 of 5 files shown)


lib/credo/exs_loader.ex
```diff
@@ -1,4 +1,14 @@
 defmodule Credo.ExsLoader do
+  def validate(parsed_file) do
+    if !Enum.all?(parsed_file[:configs], &do_validate/1) do
+      IO.puts("One or more of your .credo.exs files are invalid")
```
devonestes: This is pretty much just a placeholder for now. I'd love some input on a preferred way to give users this information. Should we reserve it until the final output? In cases where there are more than one `.credo.exs` file, should we save all of these warnings and bundle them together with more information as to _which_ files are invalid? If we do go down that path, it would require a slightly different implementation, though.

test/credo/exs_loader_test.exs
```diff
@@ -29,4 +30,25 @@ defmodule Credo.ExsLoaderTest do
...
+          check_for_updates: true,
+          strict: false,
+          color: true,
+          checks: [{Credo.Check.Consistency.ExceptionNames}]
+        }
+      ]
+    }
+    test_func = fn -> Credo.ExsLoader.validate(invalid_config) end
+
+    refute capture_io(test_func) == ""
```
devonestes: This test is also sort of a placeholder until I know how we want to handle the output. I'll do a more specific test once I feel confident that I know what the final output will be. For now, though, at least we're testing that we're capturing something that's going to `:stdout`.

rrrene:
@devonestes Thx for this!

A couple of notes:

1) We should not use `IO.puts` in the loader module as I think it violates a seperation of concerns.
2) We unfortunately cannot ensure validity through the mere presence of certain fields in the map.

```elixir
%{
  configs: [
    %{
      name: "default",
      requires: ["lib/my_app/credo_checks/no_secrets_outside_config_check.ex"],
      checks: [
        {MyApp.CredoChecks.NoSecretsOutsideConfig},
      ]
    }
  ]
}
```

This is a perfectly valid `.credo.exs` file although some keys are missing.

What we would need to do here is to check that the required file actually exists and that the added check `NoSecretsOutsideConfig` is actually present *after* the requiring (this was the original problem in #353).

devonestes:
@rrrene Ah, I think I was just looking at #359 and assumed that because the missing `name` key was the problem there, that those keys needed to be there.

So, if I understand correctly, the rules for a valid `.credo.exs` file are:

1) Must have a `name` key with a string as a value
2) Any files listed in `requires` must exist
3) Any check added in `checks` are present after requiring all files listed in `requires`

Does that sound more like the correct validation logic? Is there anything else I'm missing?

Also, about the `IO.puts`, that's totally just a placeholder for now until I figure out how we want to actually display these warnings. I just wanted to try and get the validation logic settled before I moved on to that part of this feature.

rrrene:
@devonestes Correct. :+1:

In the future, we could also think about 

* validating types of all config values (e.g. `strict` has to be a `:bool`),
* validating check params (every check has a function called `params_names/0` containing all valid param names) and
* validating check params types (we are not specifying those yet).

devonestes:
@rrrene Ok, I think the validation logic is solid now. Take a peek and let me know if you have any other ideas on that.

For ensuring all the checks listed in `checks` were valid, I couldn't think of any better way to validate that other than actually requiring all the files in `requires`, and then checking to see if all the modules listed in `checks` had been compiled. I think that by adding this I'm making some other work done later redundant, so maybe we can or should remove that?

Also, if the logic looks good, I'd love some input on how you want any of these validation errors to be displayed. I'm thinking that we might want to print something to the console and `exit(1)`, since running credo with a malformed `.credo.exs` file is going to probably be a waste of the user's time since it won't behave as expected.

rrrene:
I think catching the "validation errors" per file is mandatory to provide a rich and helpful error description.

Thinking out loud ...

1. rather than calling the validate functions with the map, we will have to call it with the map and filename of the particular config and
2. we probably want to return something like `{filename, line_no, :atom_key_of_the_problem}` so that we can keep track of these validations and
3. warn at some later point in the execution.

Which also begs the philosophical question: What happens if a config file that gets merged contains an invalid field, but it gets overwritten with something valid. Should we prevent execution of Credo in this case?

devonestes:
Yeah, I think my plan is going to be:

1. Catch as much info as possible about what's invalid and store it throughout the validation procedure, and
2. If anything is invalid, then we halt execution and output the information to the user to correct.

I personally think that requiring that each individual file be valid isn't too much of a burden on the user. I think that requirement is actually probably a helpful thing in the end, since if I were working on a project with an invalid `.credo.exs` file, I'd probably get really confused by that - even if somewhere else there was some other file that made it valid.

I'll get working on this today- yay!

rrrene:
Thx for working on this!

A final thought: Maybe we shouldn't validate the map returned my ExsLoader, but rather validate the AST of the config file (which is a map) directly by traversing it, so we can use pattern matching in function heads (like you started to do) *and* use the `meta[:line]` information from `{op, meta, arguments}` to locate the "error" for the user.

devonestes:
@rrrene Ok, I think this is more along the lines of what you were thinking now. Funny enough, the only validation that I could get a line number for was the one to make sure everything in the `checks` is valid, but frankly that's probably the most important one since there are often a lot of those.

I'm not in love with the naming of the module, so let me know if you think some of this code should be moved around a bit.

devonestes:
@rrrene Just checking in on this since it's been open for a while 😉

rrrene:
@devonestes Super sorry for not checking in. We're switching houses and the last weeks have been ... challenging 😉  I will try to get back as soon as I can!

PR #388, opened 2017-05-30
Replace deprecated functions for Elixir 1.5

leifg:
I came across a lot of compiler warning when testing out credo with Elixir 1.5

There is an easy fix, so I applied it

### DIFF
--- a/lib/credo/check/design/alias_usage.ex
+++ b/lib/credo/check/design/alias_usage.ex
@@ -159,13 +159,12 @@ defmodule Credo.Check.Design.AliasUsage do
 
         Map.put(memo, issue.trigger, [issue | list])
       end)
-    |> Enum.filter_map(
-      fn({_, value}) ->
+    |> Enum.filter(fn({_, value}) ->
         length(value) > count
-      end,
-      fn({_, value}) ->
-        value
       end)
+    |> Enum.map(fn({_, value}) ->
+      value
+    end)
     |> List.flatten
   end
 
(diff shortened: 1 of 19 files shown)


leifg:
Oh just saw, this clashes with Elixir versions `< 1.3` I don't know what the guideline is. Shall old versions be supported? If yes, then I guess this cannot be merged.

rrrene:
> There is an easy fix, so I applied it

I am afraid it's not that easy and I would really like to ensure backwards compatibility: https://travis-ci.org/rrrene/credo/builds/237500191?utm_source=github_status&utm_medium=notification

The only solution I see atm would be a "backports" module which uses the right functions at compile time. WDYT?

P.S. you already noticed this yourself. :wink:

leifg:
Mmmh interesting, does something like that already exist?

leifg:
OK I implemented a [backports library](https://github.com/leifg/backports) over the weekend. Tests run and the code is up to date.

What do you think?

rrrene:
Hi, sorry for the delayed answer.

Unfortunately, when I wrote that I'd like to use a module here, I did mean to use a module. I would not want to have such a critical dependency in Credo and not have write access to the source code.

Furthermore, I think your Hex package tries to do too much. I quickly looked over the source code and there is some dark meta programming magic going on, which can easily break on future versions of Elixir, just like `v0.1.0` of `backports` demonstrated.

That said, I merged most of your commits by hand because your contribution is very valuable to this project, even without the last commit. :+1:

P.S. I also referenced two commits in this PR which show how I implemented the Backports module.

leifg:
I see your point. I'm fine with the current implementation.

PR #391, opened 2017-06-03
Emit a warning for raise inside rescue

### LINKED ISSUE #381: Proposal: warn on raise inside rescue
An error that I see happening many times is people doing `raise` inside a caught exception, because they are unaware of the `reraise` function:

```elixir
try do
  # .. some stuff
rescue
  e ->
    Logger.warn("Something bad happened!")
    raise e
end
```

Pattern like this swallows stack trace of where the original exception occurred and replaces it with the stacktrace of the new `raise`, leaving you in the dark. Would it be an idea to emit a warning when a pattern like this is discovered?

little-bobby-tables:
Resolves #381.


rrrene:
Th! :+1:

PR #401, opened 2017-07-09
Handle 'raise inside rescue' for implicit try

### LINKED ISSUE #398: Does not warn me on `raise` with implicit `try`
### Environment

* Credo version (`mix credo -v`): 0.8.1

> Erlang/OTP 19 [erts-8.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
> 
> Elixir 1.4.4

* Operating system: macOS

This block raises two warnings:

```elixir
defp load_values do
   try do
     :petrovich_elixir
      |> Config.get_env(@setting)
      |> File.read!
      |> Poison.decode!
  rescue
    e in File.Error ->
      raise RulesFileException, message: Exception.message(e)
    end
end
```

First one is about code readability:

>   Code Readability                                                              
> ┃ 
> ┃ [R] ↓ Prefer using an implicit `try` rather than explicit `try`.
> ┃       lib/petrovich/stores/base.ex:41:9 #(Petrovich.Store.__using__.load_values)
> 

It also raises this warning about `reraise`: 

>   Warnings - please take a look                                                 
> ┃ 
> ┃ [W] → Use `reraise` inside a rescue block to preserve the original stacktrace.
> ┃       lib/petrovich/stores/base.ex:48:13 #(Petrovich.Store.__using__.load_values)

But when I try to change it to the implicit version:

```elixir
      defp load_values do
        :petrovich_elixir
        |> Config.get_env(@setting)
        |> File.read!
        |> Poison.decode!
      rescue
        e in File.Error ->
          raise RulesFileException, message: Exception.message(e)
      end
```

It does not raise any warnings at all.

### Expected outcome

One warning with `reraise` note.

### Actual outcome

No warnings.

Full source: https://github.com/petrovich/petrovich_elixir/blob/053f38851b605c9016c8252247a73ee2612957e6/lib/petrovich/stores/base.ex#L40

philipdexter:
Fix #398

### DIFF
--- a/lib/credo/check/warning/raise_inside_rescue.ex
+++ b/lib/credo/check/warning/raise_inside_rescue.ex
@@ -48,6 +48,11 @@ defmodule Credo.Check.Warning.RaiseInsideRescue do
         {ast, issues}
     end
   end
+  defp traverse({df, _meta, [_def, [do: _do, rescue: rescue_block]]}, issues, issue_meta)
+  when df == :def or df == :defp do
+    issues_found = Credo.Code.prewalk(rescue_block, &find_issues(&1, &2, issue_meta))
+    {rescue_block, issues ++ issues_found}
+  end
   defp traverse(ast, issues, _issue_meta), do: {ast, issues}
 
   defp find_issues({:raise, meta, _arguments} = ast, issues, issue_meta) do
--- a/test/credo/check/warning/raise_inside_rescue_test.exs
+++ b/test/credo/check/warning/raise_inside_rescue_test.exs
@@ -60,6 +60,26 @@ end
       end)
   end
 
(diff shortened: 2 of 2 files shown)


rrrene:
Thx for submitting a PR!

Could you please add at least one test, which fails with the previous implementation and is fixed by your code? :+1:

philipdexter:
Absolutely, updated.

rrrene:
Great. Now we just need to make sure that we do not accidentally trigger the check for unlikely, but valid code like this:

```elixir
"""
  def catcher do
    x = [do: 2, rescue: 6]
  end
end
"""
```

philipdexter:
Of course, this new version should fix that

rrrene:
@philipdexter Thx!

PR #419, opened 2017-08-05
Allow anonymous functions to be piped as raw values

### LINKED ISSUE #417: False positive when piping functions into HOFs
### Precheck

* For proposing a new feature, be sure to name your issue "Proposal: ..."

There is a _possibly_ similar closed issue (#48), but I'm pretty sure that this is a different case, since they cite an example...

`do_something() |> then_something_else() |> and_last_step()`

...and I'm trying for...

`fn x -> stuff end |> hof_a() |> hof_b() |> ...`

### Environment

* Credo version (`mix credo -v`): 0.8.4
* Erlang/Elixir version (`elixir -v`): Elixir 1.5.0
* Operating system: macOS 10.12.6 

Pipe a function into a higher order function

### Expected outcome

No warning

### Actual outcome

Credo complains with `Credo.Check.Refactor.PipeChainStart`.

The library that I'm working on is full of higher order functions. Unapplied functions _are_ values. It's a bit of "stuck between a rock and a hard place", since applying the HOF directly causes the exact same issue, because then the form is `hof(fun) |> more() |> ...` (yuck).

I've been [peppering my code](https://github.com/expede/witchcraft/search?utf8=✓&q=Credo.Check.Refactor.PipeChainStart&type=) with `# credo:disable-for-lines:<LINES> Credo.Check.Refactor.PipeChainStart` because I _like_ the rule when it actually applies (ex. `foo(val) |> bar()`), so I don't want to disable it globally. Credo is just being a bit overzealous in this case IMHO 😉 

### Letter of Appreciation

Long time user, first time writer. Thanks for creating and maintaining such a great library! I use it in _every_ project, recommend it to companies that I train, promote it at meet up, &c.`credo --strict` has saved many codebases 🙏 Thanks for all the effort!

philipdexter:
Fix #417

### DIFF
--- a/lib/credo/check/refactor/pipe_chain_start.ex
+++ b/lib/credo/check/refactor/pipe_chain_start.ex
@@ -55,6 +55,10 @@ defmodule Credo.Check.Refactor.PipeChainStart do
       true
     end
   end
+  # anonymous function
+  defp valid_chain_start?({:fn, _, [{:->, _, [_args, _body]}]}, _excluded_functions) do
+    true
+  end
   # function_call()
   defp valid_chain_start?({atom, _, []}, _excluded_functions) when is_atom(atom) do
     true
--- a/test/credo/check/refactor/pipe_chain_start_test.exs
+++ b/test/credo/check/refactor/pipe_chain_start_test.exs
@@ -115,6 +115,15 @@ defmodule CredoSampleModule do
       i
     end
     |> Enum.map(fn(i) -> i + 2 end)
+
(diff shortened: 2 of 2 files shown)


OvermindDL1:
Might also want to allow the AST of `{:&, _, [_]}` too for anonymous functions.

philipdexter:
That seems to already be supported, but I've added test cases for that in my PR to make it certain going forward.

rrrene:
:+1:

PR #423, opened 2017-08-11
[Fix #409] Fix explicit `explain` command usage

### LINKED ISSUE #409: mix credo explain path_line_no_column printing command help instead of the explanation
### Environment

* Credo version (`mix credo -v`): 0.8.3
* Erlang/Elixir version (`elixir -v`): 1.4.4
* Operating system: Darwin 16.6.0 x86_64

`mix credo explain lib/core/users/users.ex:71:17`

### Expected outcome

Same as `mix credo lib/core/users/users.ex:71:17`, because it works as expected without the `explain` subcommand.

### Actual outcome

```
Usage: mix credo explain path_line_no_column [options]

Explain the given issue.

Example: $ mix credo explain lib/foo/bar.ex:13:6

General options:
  -v, --version       Show version
  -h, --help          Show this help
```

ivanovaleksey:
Hi guys,

I am very new to Elixir and this is my first try to contribute to any Elixir project.

I have found the issue #409 which is marked as good for beginners.
@philipdexter even managed to provide code snippet with solution.

I debug the issue to understand how it works (should I say _how it doesn't work_ 😄) and came to almost the same solution.

The difference is that I didn't touch `Credo.CLI.Command.Explain` itself and just prepared proper `exec` structure in `Credo.Execution.Task.ConvertCLIOptionsToConfig`.

It works fine for me now but maybe I have missed something?

Any comments are welcome.

Thank you.

### DIFF
--- a/lib/credo/execution/task/convert_cli_options_to_config.ex
+++ b/lib/credo/execution/task/convert_cli_options_to_config.ex
@@ -36,6 +36,10 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
       set_command_and_path(exec, options, @default_command_name, options.path)
     end
   end
+  defp determine_command(exec, %Options{command: "explain", args: args} = options) do
+    potential_path = List.first(args)
+    set_command_and_path(exec, options, "explain", potential_path)
+  end
   defp determine_command(exec, _options), do: exec
 
   defp set_command_and_path(exec, options, command, path) do


lib/credo/execution/task/convert_cli_options_to_config.ex
```diff
@@ -36,6 +36,10 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
       set_command_and_path(exec, options, @default_command_name, options.path)
     end
   end
+  defp determine_command(exec, %Options{command: "explain", args: args} = options) do
+    potential_path = List.first(args)
+    set_command_and_path(exec, options, "explain", potential_path)
+  end
   defp determine_command(exec, _options), do: exec
 
   defp set_command_and_path(exec, options, command, path) do
```
ivanovaleksey: By the way, can we use nested update here?
Something like:
```
exec
|> put_in([Access.key!(:cli_options), Access.key!(:command)], command)
|> put_in([Access.key!(:cli_options), Access.key!(:path)], path)

# or
exec = put_in(exec.cli_options.command, command)
put_in(exec.cli_options.path, path)
```

lib/credo/execution/task/convert_cli_options_to_config.ex
```diff
@@ -36,6 +36,10 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
       set_command_and_path(exec, options, @default_command_name, options.path)
     end
   end
+  defp determine_command(exec, %Options{command: "explain", args: args} = options) do
+    potential_path = List.first(args)
+    set_command_and_path(exec, options, "explain", potential_path)
```
ivanovaleksey: We don't need to set command here actually just path.
Should I use something like?
```
put_in(exec.cli_options.path, potential_path)
```

philipdexter:
Hey there,

Cool that you're starting with Elixir.

About your PR, maybe I'm missing something but it doesn't totally solve the problem.
Given `a.ex` as
```elixir
defmodule A do
  @moduledoc false
  def a do
    1+2
    1 + 1
    2 + 2
  end
end
```

```bash
$ mix credo a.ex
There are spaces around operators most of the time, but not here.
      a.ex:4:6 #(A.a)
```

And then

```bash
$ mix credo explain a.ex:4:6
Checking 1 source file ...
```
[not the expected output]

However if we bring in the other change from my initial attempt at the solution then the 'correct' output is shown (same as if I did `mix credo a.ex:4:6`).

Can you confirm that I tried this correctly?

rrrene:
@ivanovaleksey @philipdexter  Thx for working on this. Unfortunately, I could not merge your solution. I don't want to change `convert_cli_options_to_config.ex` to make this work.

Let me explain: For the future, we want to give as little "special treatment" to individual commands as possible. If we want to introduce a plugin system after Credo v1.0, we will have to allow plugin authors to create their own commands and for this, we have to figure out what information is needed for commands to "stand on their own".

Thx again for exploring this. I can't put as much time into Credo as I want to at the moment and I sincerely appreciate the community jumping in :+1: :1st_place_medal:

rrrene:
@ivanovaleksey @philipdexter  P.S. You can have a look at f07f6a1 to see what I tried to explain above.

ivanovaleksey:
@rrrene yes, I had the same thought about yet another special preparations are not good at all.
Yesterday I couldn't just provide such as good solution as your one.
But now I think it was not so difficult to do.

Issue #329, opened 2017-02-25
Credo fails on :unquote


walkr:
I'm working on a Phoenix app, which has as dependencies other custom apps.
It looks like every time I have this pattern (generating functions) in my libraries, credo fails to run:

Complaint is on line 36.

```elixir
  # Example

  for level <- [:debug, :info, :warn, :error] do
    def unquote(level)(message, details \\ []) do  #line 36
      report(unquote(level), message, details)
    end
  end
```

### Environment

* Credo version (`mix credo -v`): 0.6.1
* Erlang/Elixir version (`elixir -v`): 1.4.1
* Operating system: MacOS

### What were you trying to do?

Run credo normally

### Expected outcome

It should not fail

### Actual outcome

```shell
Error while running Elixir.Credo.Check.Readability.Specs on apps/reporter/lib/reporter.ex
** (EXIT from #PID<0.70.0>) an exception was raised:
    ** (Protocol.UndefinedError) protocol String.Chars not implemented for {:unquote, [line: 36], [{:level, [line: 36], nil}]}
        (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
        (elixir) lib/string/chars.ex:17: String.Chars.to_string/1
        lib/credo/source_file.ex:54: Credo.SourceFile.column/3
        lib/credo/check.ex:149: Credo.Check.add_custom_column/5
        lib/credo/check.ex:133: Credo.Check.format_issue/5
        lib/credo/check/readability/specs.ex:49: Credo.Check.Readability.Specs.traverse/4
        (elixir) lib/macro.ex:276: Macro.do_traverse/4
        (elixir) lib/enum.ex:1325: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
```

Any thoughts?

Thank you!

walkr:
My apologies. I was running credo 0.6.0 in my main app.
Issue does not exist in 0.6.1

Issue #311, opened 2017-01-25
Elixir.Credo.Check.Readability.Specs crash on unquote of def name


KronicDeth:
### Environment

* Credo version (`mix credo -v`): `0.6.0`
* Erlang/Elixir version (`elixir -v`):  `19.2` / `1.4.0`
* Operating system: OSX

### What were you trying to do?

`mix credo --strict` on https://github.com/C-S-D/calcinator/tree/fe1459c836be649d5e27816d5957315dcd1a5683 after updating `credo` to `0.6.0`

#### Git Patch
```
diff --git a/mix.exs b/mix.exs
index 8a4e168..f40c42d 100644
--- a/mix.exs
+++ b/mix.exs
@@ -43,7 +43,7 @@ defmodule Calcinator.Mixfile do
       # JSONAPI document coding
       {:alembic, ">= 3.1.1 and < 3.2.0"},
       # Static Analysis
-      {:credo, "0.5.3", only: [:test]},
+      {:credo, "0.6.0", only: [:test]},
       # Type checking
       {:dialyze, "~> 0.2.1", only: :test},
       {:ex_doc, "~> 0.14.0", only: [:dev, :test]},
diff --git a/mix.lock b/mix.lock
index c9b25a8..6f523c4 100644
--- a/mix.lock
+++ b/mix.lock
@@ -1,6 +1,6 @@
 %{"alembic": {:hex, :alembic, "3.1.1", "5d06c655b238369c8e0ee9206fdde1bdb8a45dbc5129b9091a0fe0883403c0c8", [:mix], [{:ecto, "~> 2.0", [hex: :ecto, optional: false]}, {:poison, "~> 1.5 or ~> 2.0", [hex: :poison, optional: false]}]},
-  "bunt": {:hex, :bunt, "0.1.6", "5d95a6882f73f3b9969fdfd1953798046664e6f77ec4e486e6fafc7caad97c6f", [:mix], []},
-  "credo": {:hex, :credo, "0.5.3", "0c405b36e7651245a8ed63c09e2d52c2e2b89b6d02b1570c4d611e0fcbecf4a2", [:mix], [{:bunt, "~> 0.1.6", [hex: :bunt, optional: false]}]},
+  "bunt": {:hex, :bunt, "0.2.0", "951c6e801e8b1d2cbe58ebbd3e616a869061ddadcc4863d0a2182541acae9a38", [:mix], []},
[...]

Issue #275, opened 2016-12-13
Bug: Credo crashes checking an unquote(something)


ConnorRigby:
### Environment

* Credo version (`mix credo -v`): 0.5.2

* Erlang/Elixir version (`elixir -v`): 1.3.3
* Operating system: Linux

### What were you trying to do?
This module is unable to be linted for some reason
``` elixir
defmodule Farmbot.StateTracker do
  @moduledoc """
    Common functionality for modules that need to track
    simple states that can be easily represented as a struct with key value
    pairs.
  """
  @callback load(any) :: {:ok, map}
  defmacro __using__(name: name, model: model) do
    quote do
      alias Farmbot.ConfigStorage, as: FBConfigStorage

      defmodule State, do: defstruct unquote(model)

      defp get_config(:all) do
        GenServer.call(FBConfigStorage, {:get, unquote(name), :all})
      end

      defp get_config(key) do
        GenServer.call(FBConfigStorage, {:get, unquote(name), key})
      end

      defp put_config(key, value) do
        GenServer.cast(FBConfigStorage, {:put, unquote(name), {key, value}})
      end

      @doc """
        Starts a #{unquote(name)} state tracker.
      """
      def start_link(), do: start_link([])
      def start_link(args),
        do: GenServer.start_link(unquote(name), args, name: unquote(name))

      def init(args) do
        n = Module.split(unquote(name)) |> List.last
        Logger.debug ">> is starting #{n}."
        case load(args) do
          {:ok, %State{} = state} ->
            {:ok, broadcast(state)}
          {:error, reason} ->
            Logger.error ">> encountered an error starting #{n}" <>
              "#{inspect reason}"
            Farmbot.factory_reset
          err ->
            Logger.error ">> encountered an unknown error in #{n}"
            <> ": #{inspect err}"
            Farmbot.factory_reset
        end
      end

      # this should be overriden.
      def load(_args), do: {:ok, %unquote(name).State{}}

      defp dispatch(reply, %unquote(name).State{} = state) do # THIS IS THE LINE IT ERRORS ON
        broadcast(state)
        {:reply, reply, state}
      end

      # If something bad happens in this module it's usually non recoverable.
      defp dispatch(_, {:error, reason}), do: dispatch({:error, reason})
      defp dispatch(%unquote(name).State{} = state) do
        broadcast(state)
        {:noreply, state}
      end

      defp dispatch({:error, reason}) do
        Logger.error ">> encountered a fatal error in #{unquote(name)}."
[...]

PR #425, opened 2017-08-14
Guard against unquoting attributes on `ModuleAttributeNames`.

lucasmazza:
Similar scenario as #329, #311, #275 and others. The fix is a carbon copy of how cb0622af407b9bf04fc2d5214020e935d111aca7 handles these cases.

### DIFF
--- a/lib/credo/check/readability/module_attribute_names.ex
+++ b/lib/credo/check/readability/module_attribute_names.ex
@@ -38,11 +38,12 @@ defmodule Credo.Check.Readability.ModuleAttributeNames do
     {ast, issues}
   end
 
-  defp issue_for_name(issue_meta, name, meta) do
+  defp issue_for_name(issue_meta, name, meta) when is_binary(name) or is_atom(name) do
     unless name |> to_string |> Name.snake_case? do
       issue_for(issue_meta, meta[:line], "@#{name}")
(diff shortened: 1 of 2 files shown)


lib/credo/check/readability/module_attribute_names.ex
```diff
@@ -38,12 +38,14 @@ defmodule Credo.Check.Readability.ModuleAttributeNames do
...
   end
 
-  defp issue_for_name(issue_meta, name, meta) do
+  defp issue_for_name(issue_meta, name, meta) when is_binary(name) or is_atom(name) do
     unless name |> to_string |> Name.snake_case? do
       issue_for(issue_meta, meta[:line], "@#{name}")
     end
   end
 
+  defp issue_for_name(_, _, _), do: nil
```
ivanovaleksey: I am new to Elixir and haven't know coding conventions yet.
Is it common to have same method's clauses line by line (without empty line), especially when one of the clauses is a oneliner?

lucasmazza: it usually is, but my brain in always on Ruby mode. I'll fix it, thanks.

rrrene:
Thx @lucasmazza! :+1:

Issue #432, opened 2017-09-04
Incorrect underlining in `list` mode


nekath:
### Precheck

* I didn't find similar issue

### Environment

* Credo version (`mix credo -v`): master branch
* Erlang/Elixir version (`elixir -v`): Erlang/OTP 20 / Elixir 1.5.1
* Operating system: OSX

### What were you trying to do?
I've used `mix credo list --strict` to show all my issues. But the indentation of underlining is incorrect.

I've also check a default option, and everything looks fine - using `mix credo suggest test.ex --strict --verbose`

### Expected outcome
![correct_underlining](https://user-images.githubusercontent.com/1890734/30035317-2b09039a-91a9-11e7-9522-2bef4f01f355.png)

### Actual outcome
![incorrect_underlining](https://user-images.githubusercontent.com/1890734/30097398-bfb16246-92dd-11e7-9cc8-69d27e4cfbef.png)

NobbZ:
Sorry, but except a slight difference in the colorschema used, I don't see a difference, can you elaborate?

nekath:
@NobbZ I've updated second screenshot - by mistake I've uploaded the same image twice.

rrrene:
@nekath Ouh, this is a bug. Thx for reporting! :+1:

On a related note: Are you actually using `list` mode? I was thinking about retiring it for Credo 1.0, because it seems like something I added in Credo's early days to have more than one command ^_^

nekath:
@rrrene I've been using `credo list` for 2 reasons:
- I was not aware of --verbose flag - now I know that I could use `mix credo --verbose`
- It's easier (to me) to fix issues file after file.

Imho it would be good to preserve grouping issues by file name/path. It could be done using some custom flag like `--file-grouping`.

PR #433, opened 2017-09-04
Fix underlining in `list` mode

nekath:
Fixes https://github.com/rrrene/credo/issues/432

I couldn't find any example of UI test. @rrrene Could you give me a tip how to handle test case for such scenario.
My idea is to add a helper function in test_helper module - to print message as string and match it against expectation.


rrrene:
@nekath Thx! :+1:

PR #437, opened 2017-09-08
Prevent exception thrown when run against empty umbrella project

mcelaney:
# Defect (Erlang 20, Elixir 1.5.1)
- create an empty umbrella project
- install credo
- run credo
- exception is thrown because there are no source files.

# Changes
- added test to demonstrate issue
- Added pattern match for empty source_files list to `Credo.CLI.Output.Summary.scope_count/1`


rrrene:
@mcelaney Thx! :+1:

Issue #389, opened 2017-05-31
Line length calculation is incorrect on lines containing strings


lee-dohm:
### Precheck

* For proposing a new feature, be sure to name your issue "Proposal: ..."
* For bugs, please do a quick search and make sure the bug has not yet been reported here

### Environment

* Credo version (`mix credo -v`): 0.8.0
* Erlang/Elixir version (`elixir -v`): Erlang/OTC 19 [erts-8.3] Elixir 1.4.4
* Operating system: Mac OS X 10.12.5

### What were you trying to do?

On code:

```elixir
defmodule CredoSampleModule do
  use ExUnit.Case
  def some_fun do
    assert "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" <> "1" == "2"
  end
end
```

Run `mix credo test.ex`

### Expected outcome

Line is too long (max is 80, was 113).

### Actual outcome

line is too long (max is 80, was 97).

Also, if I raise the max line length to 100, I won't get a warning at all.

NobbZ:
Skimmed over it, and it seems related to this bit of code:

https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/max_line_length.ex#L40-L45

```
    source =
      if ignore_strings do
        Credo.Code.Strings.replace_with_spaces(source, "")
      else
        source
      end
```

I do not have an Elixir capable environment at hands right now. But I think it should be sufficient to change the `Credo.Code.Strings.replace_with_spaces(source, "")` to `Credo.Code.Strings.replace_with_spaces(source)`

rrrene:
Yeah, this should be `Credo.Code.Strings.replace_with_spaces(source, ".")`. Good catch!

@lee-dohm If you want to contribute this to Credo, feel welcome to submit a PR. :+1:

NobbZ:
`Credo.Code.Strings.replace_with_spaces/2` has already a default of `" "` for the second argument, so just omitting it should totally be sufficient.

rrrene:
@NobbZ on second thought, this was probably intentional to "cut" strings from line length.

Maybe we should rethink this solution.

NobbZ:
Hmmm... I think I see what you are about.

We need to recognize constructs as in the example as too long while still not reporting a line that has a single very long string in it, because one might probably not be able to alter a long literal string easily into multiple lines...

rrrene:
[...]

PR #439, opened 2017-09-08
Work in progress: Track string-only lines when handling line length

philipdexter:
This is a work in progress pull request. I am showing one possible solution
to the problem in #389. The solution and its tests still need cleaning.

I did not make the changes to Credo.Code.Strings because
its `replace_with_spaces` is used in other checks, ones
which are sensitive to this update.

Thanks!


rrrene:
Closing this since #389 is solved. Thx for putting this together as a guideline! :+1:

Issue #429, opened 2017-08-21
Can't include custom checks when Credo is installed globally


wende:
### Environment

* Credo version (`mix credo -v`): 0.8.6
* Erlang/Elixir version (`elixir -v`): 1.5.1
* Operating system: MacOs Sierra

Hello there!

I'm trying to run credo after installing it globally. I've got my `.credo.exs` file that describes which checks to use, one of which is our custom made check. The problem is, we really need credo globally to work with our 3rd party tools, but whenever it is installed globally and ran inside the project it can't find the custom rule module.

<img width="480" alt="screen shot 2017-08-21 at 15 39 48" src="https://user-images.githubusercontent.com/4634386/29524171-006cd698-8687-11e7-9f8b-936b9db21681.png">

Is there a way to run Credo globally AND have our custom checks working?

rrrene:
@wende Ouh... This sounds like a bug. Thx for reporting!

We will have to investigate this further! :+1:

rrrene:
Thx for reporting this! :+1: It should be be fixed in the latest version of Credo.

If it is not, please feel free to re-open this issue!

PR #444, opened 2017-09-10
The arity of call() should be 2. Otherwise, task.ex's run/3 function …

nealfax:
…will never actually invoke this function, instead invoking the default call/2 defined in the task macro. This will result in the 'requires' section of .credo.exs to be ignored, causing custom checks specified there to never get loaded.


nealfax:
You might also want to consider having the default call() function raise an error, since I can't imagine there'd be a check where you **wouldn't** want to override it?

nealfax:
Refs: #429

PR #451, opened 2017-09-17
Add prefer unquoted atoms check

### LINKED ISSUE #447: Proposal: Favor unquoted atoms when quotes are unnecessary
People on my team deal with a lot of JSON and end up writing the Elixir equivalent of JSON in the language. Because of that, they often make the mistake of leaving a colon after a quoted string, which is an atom in Elixir.

They have this:

``` elixir
%{"x": 1}
```

Instead of:

``` elixir
%{"x" => 1}
```

If credo complained about quotes being used for an atom when not necessary, it would be helpful for them and me to find these types of issues.

blatyo:
Closes #447


rrrene:
@blatyo Thx! :+1:

PR #453, opened 2017-09-30
Fix loading of custom checks

sbrink:
This fixes the loading of the RequireRequires task.

Without the 2nd param, the task is not executed and loading of custom checks fails.


rrrene:
@sbrink Thx! :+1:

P.S. long time no see! :wink:

PR #458, opened 2017-10-06
Remove duplication from PipeChainStart moduledoc

mustardnoise:


rrrene:
Good catch! Thx! :+1:

PR #467, opened 2017-10-22
Skip pipe start check on erlang functions

### LINKED ISSUE #431: Proposal: Ignore refactoring opportunities in pipe chain when using erlang functions
### Environment

* Credo version (`mix credo -v`): 0.8.6
* Erlang/Elixir version (`elixir -v`): Elixir 1.5.1 and Erlang/OTP 20
* Operating system: macOS 10.12

:crypto.hash(:md5, "test") |> Base.encode16(case: :lower)

### Expected outcome

Expected credo to ignore piping for erlang functions.

### Actual outcome

Refactoring opportunities: _Pipe chain should start with a raw value._

In my opinion it doesn't make sense to report a refactoring opportunity in a pipe chain with erlang modules as they are not created with piping in mind.

Best wishes,
Patrick

jdollar:
Attempted to add a config option to allow credo to think erlang functions are valid for beginning a pipe stanza as per issue #431. Currently the only way I am seeing to determine for sure that the module is from erlang  is to call some erlang module specific functions and handle the try rescue response accordingly. If it doesn't throw a exception then we can pretty safely assume it is a erlang module. Otherwise it is something else.

### DIFF
--- a/lib/credo/check/refactor/pipe_chain_start.ex
+++ b/lib/credo/check/refactor/pipe_chain_start.ex
@@ -21,9 +21,10 @@ defmodule Credo.Check.Refactor.PipeChainStart do
 
   @explanation [
     check: @moduledoc,
-    excluded_functions: "All functions listed will be ignored."
+    excluded_functions: "All functions listed will be ignored.",
+    exclude_erlang_funcs: "Specifies that we want to treat erlang functions as raw values"
   ]
-  @default_params [excluded_functions: []]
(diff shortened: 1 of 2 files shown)


rrrene:
The PipeChainStart check is a controversial one and I am no longer sure if we should keep it.

I am closing this due to its age and incompatibility with `master`. If you still think this is a good idea, please feel free to re-open! :+1:

PR #474, opened 2017-12-01
adding a pattern to match @specs containing `with` clauses

zachdaniel:
Noticed that these kinds of specs (ones containing a `with` clause) weren't counting as a spec for the function.


rrrene:
LGTM. Thanks! :+1:

PR #479, opened 2017-12-11
Add :allow_shorter_alias param to AliasUsage

weisslj:
This adds an option allowing to use a shorter alias than possible, as this sometimes improves readability.

The AliasUsage check requires to alias used modules so that dependencies are easier to spot [[1]], for example:

```elixir
defmodule Test do
  alias MyApp.External.MyProto.Client

  def something do
    Client.search(...)
  end
end
```

But sometimes it is more readable to call "MyProto.Client.search" instead of "Client.search". The dependency can still be seen at the top of the file (only slightly less specific):

```elixir
defmodule Test do
  alias MyApp.External.MyProto

  def something do
    MyProto.Client.search(...)
  end
end
```

[1]: https://github.com/rrrene/elixir-style-guide#alias-modules

### DIFF
--- a/lib/credo/check/design/alias_usage.ex
+++ b/lib/credo/check/design/alias_usage.ex
@@ -34,6 +34,7 @@ defmodule Credo.Check.Design.AliasUsage do
     params: [
       if_nested_deeper_than: "Only raise an issue if a module is nested deeper than this.",
       if_called_more_often_than: "Only raise an issue if a module is called more often than this.",
+      allow_shorter_alias: "Allow using shorter aliases than possible for stylistic reasons.",
     ]
   ]
   @default_params [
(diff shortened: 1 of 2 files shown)


weisslj:
Please suggest a better name if "allow_shorter_alias" does not sound right.

rrrene:
Thanks for your PR and the work you put into it.

I am hesistant to merging this. This use-case is what the `if_nested_deeper_than:` is usually used for. Adding another option with an undescriptive name for a more narrow use-case does not seem beneficial to me at this point.

But I'm not strictly against this either. Let's see if we can find more pro/con-arguments.

rrrene:
Hi, thx again for this. Unfortunately, I am closing it due to "it being obsolete" IMHO. 

PR #480, opened 2017-12-15
Fix Markdown link to bunt in the README

bijanbwb:
I think the Markdown syntax was backwards for this link, which was pointing to the following (404) location:
https://github.com/rrrene/credo/blob/master/bunt

The updated syntax in this commit should fix the link so it points to the correct location:
https://github.com/rrrene/bunt

/cc @rrrene


rrrene:
Ouch. Thanks! :+1:

PR #497, opened 2018-01-24
Fix warning on unused @lint

halfdan:
We're using warnings as errors and this warning keeps popping up on Elixir 1.6. This PR is an attempt to suppress the warning.

```
==> credo
Compiling 178 files (.ex)
warning: module attribute @lint was set but never used
  lib/credo/check/find_lint_attributes.ex:23
```


rrrene:
:+1: Thx for this PR! 

Could you add a comment saying "Hack to avoid compiler warnings"?

PR #501, opened 2018-01-30
Command-line switch to load a custom configuration file

grossvogel:
This functionality would be useful for a [web service](https://github.com/revelrylabs/lintron) I use that lints GitHub PRs. It [uses credo](https://github.com/revelrylabs/lintron/blob/master/app/models/linters/credo.rb) with the `--read-from-stdin` switch, which works great. Since credo is invoked in the context of the web service, though, rather than the project whose PR is being linted, it can't use that project's credo config without a path override like this.

I think this option is definitely going to be helpful to anybody using stdin or passing a [list of files](https://github.com/rrrene/credo/pull/334#issuecomment-288332243).

Please let me know if a feature like this would be compatible with credo's goals and if so, what else I can do to help push this work forward.


rrrene:
I think this is a good idea and a valid scenario. :+1:

By default Credo loads the config in the current directory and then walks up to `/` to find all relevant configs, which it then merges together.

- Should we just load the given config file and ignore other config files above it?
- Should we just load the given config file and merge it with what would have our merged config given the current behaviour?
- Should we load the given config and traverse the filesystem from there to merge configs along the way (basically acting as if we had executed Credo in that directory)?

Like I said, I think this can be a great addition to Credo, we just have to get the details right first 😄

grossvogel:
Thanks for looking into this idea, @rrrene!

I know you have many other perspectives to think about, but here's a bit more context from mine:

In our use case, credo will be invoked in a completely separate context from the 'home' project of the code being linted. In that case, the ideal would be the first option, namely that the specified config is the _only_ one applied on top of credo's base defaults.

If additional configs were merged via a directory walk, those settings would be unknown to (and outside of the control of) the code author, possibly resulting in unintended lints.

techgaun:
This is a useful feature. I am voting for the first option as that seems like the most intuitive understanding of what the intent would be of this particular feature.

bryanjos:
@rrrene anything we can do to help get this PR into something that can be merged?

lib/credo/cli/options.ex
```diff
@@ -11,6 +11,7 @@ defmodule Credo.CLI.Options do
     all: :boolean,
     checks: :string,
     config_name: :string,
+    config_file_path: :string,
```
rrrene: Let's name this `config_file`, since I think

```
mix credo --config-file other-credo.exs
```

looks great! :+1:

lib/credo/config_builder.ex
```diff
@@ -2,23 +2,33 @@ defmodule Credo.ConfigBuilder do
...
     exec
     |> add_config_file_to_exec(config_file)
     |> add_strict_to_exec(config_file, options)
     |> add_switches_to_exec(options.switches)
   end
 
+  defp get_config_file(%Options{} = options) do
+    config_name = options.switches[:config_name]
+    config_file_path = options.switches[:config_file_path]
+    dir = options.path
```
rrrene: Please avoid using single pipes.

lib/credo/config_builder.ex
```diff
@@ -2,23 +2,33 @@ defmodule Credo.ConfigBuilder do
...
+      |> Filename.remove_line_no_and_column()
+
+    cond do
+      is_binary(config_file_path) ->
```
rrrene: Dumb question: Why isn't this an `if/else` block?

lib/credo/config_file.ex
```diff
@@ -30,6 +30,25 @@ defmodule Credo.ConfigFile do
...
+  Returns Execution struct representing a consolidated Execution for
+  the provided config_file_path merged into the default configuration.
+
+  - `config_file_path`: full path to the custom configuration file
+  - `config_name`: name of the configuration to load
+  - `safe`: if +true+, the config files are loaded using static analysis rather
+            than `Code.eval_string/1`
+  """
+  def read_from_file_path(dir, config_file_path, config_name \\ nil, safe \\ false) do
+    [config_file_path]
```
rrrene: Please avoid using single pipes.

grossvogel:
Thanks, @rrene! I hope to make time to address your comments this week

grossvogel:
@rrrene I made the code style fixes you requested, and added a test for `ConfigFile.read_from_file_path`. Let me know what else i can do to get this ready to merge!

rrrene:
@grossvogel This is looking very good! :+1:

I will merge it as soon as I find the time. Thanks!

PR #534, opened 2018-05-01
Shorter standalone installation instructions

Tuxified:
Elixir 1.5 (or 1.6?) introduces an option to install an archive from hex directly. IMO it makes it easier for folks who just want to give Credo a spin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,13 @@ Do note with the passed option as filename is a stub that is just used to prefix
 If you do not want or are not allowed to include Credo in the current project you can also install it as an archive. For this, you also need to install [bunt](https://github.com/rrrene/bunt):
 
 ```bash
+mix archive.install hex bunt
+mix archive.install hex credo
+```
+
+**Important**: if you're using an older version of Elixir, you'll have to resort to building from source:
+
+```bash
 git clone git@github.com:rrrene/bunt.git
 cd bunt
 mix archive.build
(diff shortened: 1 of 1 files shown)


NobbZ:
According to a recent discussion on the elixir core mailing list, archives should have no dependencies to avoid clashing version requirements between local (project) and global archive dependencies.

Therefore I'm not sure if we should advertise an archive installation of credo at all.

NobbZ:
Sorry, my fault, it was not on the Mailinglist, it was the actual issue tracker.

Full discussion at https://github.com/elixir-lang/elixir/issues/7577

Relevant Comment: https://github.com/elixir-lang/elixir/issues/7577#issuecomment-383290541 and https://github.com/elixir-lang/elixir/issues/7577#issuecomment-383300978

rrrene:
> Credo's only dependency is bunt (which hasn't changed in 3 years).

Credo also depends on `poison`, which is why we should probably remove the standalone instruction altogether.

Tuxified:
> Credo also depends on poison, which is why we should probably remove the standalone instruction altogether.
Makes sense, I'll change my PR to remove those instructions then. Follow up question: is it desirable to make it possible to install/run credo standalone (for ex. via hex install as my initial suggestion)?

NobbZ:
To be honest, I prefer beeing explicit in the `mix.exs`, "this project uses credo".

I've never seen much value in the standalone install, as in my eyes it complicates things when dealing with environments that involve more than a single developer and maybe CI.

And even though many of my projects (private and work related, elixir or not) are single person, I try to treat them as if there were other developers contributing. I try to keep the implicit dependencies at a minimum.

Every single dependency I do have explicitely in my mix file clearly signals that it is needed and in case of dialyzer or credo also communicates that we want to have them run.

Often I do alias the test-task accordingly to also run credo. And also I try to remember adding a short note in the README ;)

So I'm obviously on the contra side for providing an archive at all. But perhaps some from the PRO side can tell us about their use cases?

rrrene:
Hi, thx again for this. Unfortunately, I am closing it due to its age. 

PR #538, opened 2018-05-07
Add string options for the `--min-priority` switch

### LINKED ISSUE #532: Proposal - Filter out checks to run by priority too
### Environment

* Credo version (`mix credo -v`): 0.9.1
* Erlang/Elixir version (`elixir -v`): OTP20 / 1.6.4
* Operating system: Ubuntu

I was trying to run checks based on priorities only. 

### Expected outcome

Be able to filter out the checks to run based on the priorities

### Actual outcome

Can not filter out the checks based on the priorities

### Use Case

While the `--only warning,refactor` does work to filter out the checks based on the type, sometimes it is convenient to be able to run only the checks that have priority higher than certain value such as >= `high`. For example, I want to fail on `PipeChainStart` but not on other readability issues in my CI environment but I still want to gradually fix other pipelines so I can not really disable other checks under readability category.

Something like below:

`mix credo --priority high` -> can run `high` or `higher` checks

Alternatively, we could do:
`mix credo --priority high,higher`

I could not find any relevant docs but I am sorry if I am overlooking certain things

dotboris:
Fixes #532

This does 2 things:

1.  It documents the `--min-priority` switch
1.  It adds 5 new possible values to the `--min-priority` (`high`, `medium`, `normal`, `low`, `lower`)

There's one caveat with this PR. It piggies back on the unknown flags handling feature of credo. If the `--min-priority` switch is given a string it can't parse (not one of the values above nor an integer) it'll report that the `--min-priority` switch is unknown. 

This is not great for users but it's consistent with the current behavior of credo. I didn't want to improve the error reporting for non-parsed flags in this PR since it seemed complicated and beyond the scope of the original issue. At least it's not worse than it was before.

Error output from master:

```sh
$ mix credo --min-priority what
Unknown switch: --min-priority
```

Error output from this PR:

```sh
$ mix credo --min-priority what
Unknown switch: --min-priority
```

### DIFF
--- a/lib/credo/cli/command/list/list_output.ex
+++ b/lib/credo/cli/command/list/list_output.ex
@@ -29,6 +29,7 @@ defmodule Credo.CLI.Command.List.ListOutput do
     List options:
       -a, --all             Show all issues
       -A, --all-priorities  Show all issues including low priority ones
+          --min-priority    Minimum priority to show issues (higher,high,normal,low,ignore or number)
       -c, --checks          Only include checks that match the given strings
       -C, --config-name     Use the given config instead of "default"
       -i, --ignore-checks   Ignore checks that match the given strings
--- a/lib/credo/cli/command/list/list_output.ex
+++ b/lib/credo/cli/command/list/list_output.ex
@@ -29,7 +29,7 @@ defmodule Credo.CLI.Command.List.ListOutput do
     List options:
       -a, --all             Show all issues
       -A, --all-priorities  Show all issues including low priority ones
-          --min-priority    Minimum priority to show issues (higher,high,normal,low,ignore or number)
+          --min-priority    Minimum priority to show issues (high,medium,normal,low,lower or number)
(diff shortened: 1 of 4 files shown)


dotboris:
I've noticed something interesting about the name of priorities while working on this. There seems to be two source.

When creating a check the priorities are (https://github.com/rrrene/credo/blob/master/lib/credo/check.ex#L6-L12)

- `higher`
- `high`
- `normal`
- `low`
- `ignore`

When displayed back to the user the priorities are (https://github.com/rrrene/credo/blob/master/lib/credo/cli/output.ex#L83-L92)

- `high`
- `medium`
- `normal`
- `low`
- `lower`

I have matched the names that are displayed to the user and their respective values. This should be the least confusing things for users.

rrrene:
Great work! :+1:

And I agree with your second point: We should unify all priority aliases in a central place.

PR #547, opened 2018-06-03
Add new AliasOrder check

remiprev:
There are already [tests](https://github.com/rrrene/credo/blob/master/test/credo/check/readability/alias_order_test.exs) for this check, waiting to be implemented. They now all pass.

@rrrene This is my first time implementing a check as complex as this one, so I’m looking for feedback if I should have done things differently.

Thank you! 😊

### DIFF
--- a/lib/credo/check/readability/alias_order.ex
+++ b/lib/credo/check/readability/alias_order.ex
@@ -1,0 +1,152 @@
+defmodule Credo.Check.Readability.AliasOrder do
+  @moduledoc """
+  Alphabetically ordered lists are more easily scannable by the read.
+
+      # preferred
+
+      alias Module1
+      alias Module2
+      alias Module3
+
+      # NOT preferred
+
+      alias Module1
+      alias Module3
+      alias Module2
+
+  Alias should be alphabetically ordered among their group:
+
+      # preferred
+
+      alias Module3
+      alias Module4
+
+      alias Module1
+      alias Module2
+
+      # NOT preferred
+
+      alias Module3
+      alias Module4
+
+      alias Module2
+      alias Module1
+
+  Like all `Readability` issues, this one is not a technical concern.
+  But you can improve the odds of others reading and liking your code by making
+  it easier to follow.
+  """
+
+  @explanation [check: @moduledoc]
+
+  alias Credo.Code
+  alias Credo.Code.Name
+
+  use Credo.Check, base_priority: :high
+
+  @doc false
+  def run(source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:defmodule, _, _} = ast, issues, issue_meta) do
+    new_issues =
+      ast
+      |> extract_alias_groups()
+      |> Enum.reduce([], &traverse_groups(&1, &2, issue_meta))
+
+    {ast, issues ++ new_issues}
+  end
+
+  defp traverse(ast, issues, _), do: {ast, issues}
+
+  defp traverse_groups(group, acc, issue_meta) do
+    group
+    |> Credo.Backports.Enum.chunk_every(2, 1)
+    |> Enum.reduce_while(nil, &process_group/2)
+    |> case do
+      nil ->
+        acc
+
+      line ->
+        acc ++ [issue_for(issue_meta, line)]
+    end
+  end
+
+  defp process_group([{_, first}, {line_no, second}], _) when first > second do
+    line = [
+      line_no: line_no,
+      trigger: Name.full(second -- first),
+      module: second
+    ]
+
+    {:halt, line}
+  end
+
+  defp process_group(_, _), do: {:cont, nil}
+
+  defp extract_alias_groups({:defmodule, _, _} = ast) do
+    ast
+    |> Code.postwalk(&find_alias_groups/2)
+    |> Enum.reverse()
+    |> Enum.reduce([[]], fn definition, acc ->
+      case definition do
+        nil ->
+          [[]] ++ acc
+
+        definition ->
+          [group | groups] = acc
+          [group ++ [definition]] ++ groups
+      end
+    end)
+    |> Enum.reverse()
+  end
+
+  defp find_alias_groups(
+         {:alias, _, [{:__aliases__, meta, mod_list}]} = ast,
+         aliases
+       ) do
+    modules =
+      mod_list
+      |> (&[{meta[:line], &1}]).()
+
+    accumulate_alias_into_group(ast, modules, meta[:line], aliases)
+  end
+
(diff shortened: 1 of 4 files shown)


rrrene:
@remiprev :+1: that looks pretty good already!

One thing I overlooked in the tests is something like

```elixir
      alias Credo.CLI.{Sorter,Filename,Options}
```

Could you add those to the tests and ensure that Multi-Alias is ordered as well?

remiprev:
What would be the behavior in that case? It should raise an error saying that "Sorter" is not ordered correctly? So the following line should not raise errors, is that correct?

``` 
alias Credo.CLI.{Filename,Options,Sorter}
```

remiprev:
@rrrene Done! 🎉 I added two new tests and made them pass.

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+  defp traverse_groups(group, acc, issue_meta) do
+    chunk_method =
+      if function_exported?(Enum, :chunk_every, 2) do
+        :chunk_every
+      else
+        :chunk
+      end
+
+    group
+    |> (&apply(Enum, chunk_method, [&1, 2, 1])).()
```
rrrene: There is a module called `Credo.Backports.Enum` where you can add a `chunk_every` function and delegate it based on the user's Elixir version (hint: `chunk_every` was introduced in Elixir 1.5).

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+      |> (&[{line, &1}]).()
+
+    acculumate_alias_into_group(ast, modules, line, aliases)
+  end
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue(
+      issue_meta,
+      message: "The alias `#{trigger}` is not alphabetically ordered among its group.",
+      trigger: trigger,
```
rrrene: The trigger has to be the substring which actually causes the issue in the given line.

Third party tools, like editor integrations, use this to highlight the portion of the source code that is causing the issue.

In your implementation, you expand all the multi-alias members to check their order, which is fine. But you can not use the expanded version for the `trigger` field.

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+       when line_no != 0 and line_no != line - 1 do
+    {ast, modules ++ [nil] ++ aliases}
+  end
+
+  defp acculumate_alias_into_group(ast, modules, _, aliases) do
+    {ast, modules ++ aliases}
+  end
+
+  defp find_alias_groups(
+         {:alias, _, [{:__aliases__, [line: line, column: _], mod_list}]} = ast,
```
rrrene: You can not match on `[line: line, column: _]` like this, because it is a keyword list and it will only match if both and only both values are present.

If, at some point in the future, we get another field in the metadata of the AST, our code would be incompatible between Elixir versions (this actually happened in the past when `column:` was introduced).

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+         aliases
+       ),
+       do: process_single_alias(ast, line, mod_list, aliases)
+
+  defp find_alias_groups({:alias, _, [{:__aliases__, [line: line], mod_list}]} = ast, aliases),
```
rrrene: See above re: metadata match.

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+    do: process_single_alias(ast, line, mod_list, aliases)
+
+  defp find_alias_groups(
+         {:alias, _,
+          [
+            {{:., _, [{:__aliases__, [line: line, column: _], mod_list}, :{}]}, _, multi_mod_list}
```
rrrene: See above re: metadata match.

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+          ]} = ast,
+         aliases
+       ),
+       do: process_multi_alias(ast, line, mod_list, multi_mod_list, aliases)
+
+  defp find_alias_groups(
+         {:alias, _,
+          [
+            {{:., _, [{:__aliases__, [line: line], mod_list}, :{}]}, _, multi_mod_list}
```
rrrene: See above re: metadata match.

lib/credo/check/readability/alias_order.ex
```diff
@@ -0,0 +1,174 @@
...
+        definition ->
+          [group | groups] = acc
+          [group ++ [definition]] ++ groups
+      end
+    end)
+    |> Enum.reverse()
+  end
+
+  defp find_alias_groups(
+         {:alias, _, [{:__aliases__, [line: line, column: _], mod_list}]} = ast,
```
rrrene: You can not match on `[line: line, column: _]` like this, because it is a keyword list and it will only match if both and only both values are present.

If, at some point in the future, we get another field in the metadata of the AST, our code would be incompatible between Elixir versions (this actually happened in the past when `column:` was introduced).

That's why we have do this by hand: [Example from the AliasUsage Check](https://github.com/rrrene/credo/blob/82466268e845b7f40764593ca4420bfd28f5422f/lib/credo/check/design/alias_usage.ex#L154)

lib/credo/check/readability/alias_order.ex
remiprev: Thank you for your feedback, the code is now much cleaner! There’s only this comment left and I want to make sure I get the specs right before I start implementing it 😄 

***

```elixir
defmodule Text do
  alias Module2.Module20
  alias Module1.Module10
end
```

The trigger should be `"Module1.Module10"`.

***

```elixir
defmodule Text do
  alias Module1.Module10
  alias Module1.Module09.Module009
end
```

The trigger should be `"Module09.Module009"`.

***

```elixir
defmodule Text do
  alias Module1.Module10
  alias Module2.{Module20, Module19}
end
```

The trigger should be `"Module19"`.

***

```elixir
defmodule Text do
  alias Module1.Module10.Module100
  alias Module2.Module20.{Module202, Module201, Module203}
  alias Module3.Module30.Module300
end
```

The trigger should be `"Module201"`.

***

Does that make sense?

remiprev: @rrrene I modified my implementation to fix `trigger`. We now refrain from converting AST modules into strings except when we really need them.

I also fixed a bug where `line_no` was not computed properly when multi-alias expansion was used on multiple lines.

Let me know if you’d like me to fix something else!

rrrene:
Thanks @remiprev :+1:

remiprev:
My pleasure! Do you need me to create another pull request to add a line in `CHANGELOG.md`?

rrrene:
@remiprev No need, I'm already on it ;-)

PR #550, opened 2018-06-08
Add Specs to readability checks

craiglyons:
## Problem
Running `mix credo gen.config` results in an incomplete example config, as it's missing the Spec readability check.

## Solution
Add it to `.credo.exs`, which is used in generation.


rrrene:
@craiglyons Thx for your PR. Unfortunately, I can not merge it as the check is disabled on purpose. 

It is part of the example config, at the bottom, in a "deprecated" section: https://github.com/rrrene/credo/pull/550/files#diff-78b22540b368040b6266f21dc737a6baR153

craiglyons:
Ah indeed, missed that.

PR #551, opened 2018-06-09
Add warning for Enum.map/2 |> Enum.join/2 to prefer Enum.map_join/3. 

### LINKED ISSUE #549: Proposal: Warn on Enum.map/2 |> Enum.join/2 to use Enum.map_join/3 instead
One thing that comes up in PR frequently is calling out uses of Enum.map/2 |> Enum.into/2 to use Enum.map_join/3 instead. I'd like to see this added as a compiler warning.There's other things in this "class" of warning, like chained uses of Enum.map/2 or Enum.map/2 |> Enum.into/2 ==> Enum.into/3, so this could open the floodgates to a lot of other cool warnings if we want, as well as leading to more efficient code all around without having to consciously think about it as a developer.

asummers:
Because `Enum.map/2 |> Enum.join/2` requires two iterations, `Enum.map_join/3` is preferred for performance reasons.

Unsure if this should just live in `Refactor` or get its own home. 

Resolves #549 

Point of note, I noticed a failing test on master unrelated to my changes.


isaacsanders:
@asummers Can you update this?

isaacsanders:
@rrrene Do you have time to look at this? It seems pretty straightforward, and we could use it on my team.

lib/credo/check/refactor/map_join.ex
```diff
@@ -0,0 +1,100 @@
...
+  `Enum.map/2` and `Enum.join/2` require two iterations whereas
+  `Enum.map_join/3` only requires one.
+  """
+
+  @explanation [check: @moduledoc]
+
+  use Credo.Check, base_priority: :high
+
+  @doc false
+  def run(source_file, params \\ []) do
```
isaacsanders, line 25:
```suggestion
  @impl Credo.Check
  def run(source_file, params \\ []) do
```

lib/credo/check/refactor/map_join.ex
```diff
@@ -0,0 +1,100 @@
+defmodule Credo.Check.Refactor.MapJoin do
+  @moduledoc """
```
isaacsanders, line 2: It looks like there is an error/warning when this is used rather than the explanation value below.

It also seems that you need to pass explanation into the `use` invocation

test/credo/check/refactor/map_join_test.exs
```diff
@@ -0,0 +1,107 @@
+defmodule Credo.Check.Refactor.MapJoinTest do
+  use Credo.TestHelper
```
isaacsanders, line 2:
```suggestion
  use Credo.Test.Case
```

rrrene:
@isaacsanders Do you know when `Enum.map_join/3` was introduced? We'll have to make sure that this does not cause false negatives on Elixir versions not having that function.

asummers:
It's been there since [at least 1.0](https://github.com/elixir-lang/elixir/blob/v1.0.0/lib/elixir/lib/enum.ex#L1004)

isaacsanders:
@rrrene @asummers I think the functions that were added after 1.0.0 or so might have the `@since` tag, if this function doesn't, we are probably good to go relying on it.

PR #552, opened 2018-06-09
Add warning for Enum.map/2 |> Enum.into/2 to prefer Enum.into/3.

asummers:
Enum.into/3 only requires one iteration, not two.


rrrene:
@asummers Thx Andrew! :+1:

PR #553, opened 2018-06-10
Added missing explanations for some parameters

bmbferreira:
Some parameters explanations were missing from the @explanation definition.


rrrene:
@bmbferreira Thx Bruno! :+1:

PR #554, opened 2018-06-10
Switch to Jason for --format json

doughsay:
Poison prints deprecation warnings during compilation on Elixir 1.6, and it seems like the general community is largely switching to `jason` over `poison`.  I thought I'd give the conversion a shot in credo, since it's a dependency of many projects and is probably annoying a lot of people with the deprecation warnings.

Things of note:
* `jason` doesn't do pretty printing (yet: https://github.com/michalmuskala/jason/pull/19) so the behavior of `mix credo --format json` changes to printing compact json instead of formatted json. Don't know if this is a concern or not...
* I ran `mix deps.unlock --unused` and it cleaned up the mix.lock file a bunch, since many of those dependencies are not listed anywhere. But it seems like according to the repo badges they're supposed to be there somewhere?  Confused...

Anyway, let me know if you'd like anything changed, or if you're just not interested in switching at this time.

### DIFF
--- a/lib/credo/cli/output/formatter/json.ex
+++ b/lib/credo/cli/output/formatter/json.ex
@@ -10,7 +10,7 @@ defmodule Credo.CLI.Output.Formatter.JSON do
   end
 
   def print_map(map) do
-    UI.puts(Poison.encode!(map, pretty: true))
+    UI.puts(Jason.encode!(map, pretty: true))
   end
 
   def to_json(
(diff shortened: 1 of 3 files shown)


rrrene:
I see what you mean :+1: Let's change the requirement to 

```
{:jason, "~> 1.0"}
```

and merge it. 👏

doughsay:
Updated.  Thanks!

whatyouhide:
👏

rrrene:
@doughsay 👏 Thank you!

PR #555, opened 2018-06-12
Add warning for Enum.map/2 |> Enum.map/2 to prefer one Enum.map/2, as well as all permutations of filter/reject for same concept.

asummers:
#551 and #552 can use this new helper, too, as they have the same shape.


lib/credo/check/refactor/filter_filter.ex
```diff
@@ -0,0 +1,38 @@
...
+  This should be refactored:
+
+      ["a", "b", "c"]
+      |> Enum.filter(&String.contains?(&1, "x"))
+      |> Enum.filter(&String.contains?(&1, "a"))
+
+  to look like this:
+
+      Enum.filter(["a", "b", "c"], fn letter ->
+        String.contains?(letter, "x") && String.contains?(letter, "a")
```
whatyouhide: Use `and` for boolean-only expressions, not `&&` :). Same for the other cases.

asummers:
@whatyouhide I was teetering on breaking them out into their own checks or combining them. For first cut, I broke them all out, so that they could be disabled individually, but I can see that argument. There's definitely more repetition than I'd like. 

I'm not sure if there exists a primitive in Credo to thread opts to it (could be, but I'm also pretty unfamiliar!), but we could probably use a `__using__` macro to cut down on a bunch of the repetition. That should be a trivial refactor. Then it'd be something like 

```elixir
defmodule Credor.Warning.Refactor.MapMap do
  use EnumFunctionOrderWarning,
    first_function: :map,
    second_function: :map
end
```

My hesitation there being that nothing else seems to do it like that, in Credo, but I have no a priori reason to dislike it.

whatyouhide:
That is a way to go. Ideally I would prefer no repetition in the config either, with something like:

```elixir
{Credo.Warning.Refactor.EnumChainedFunctions, enabled: [:map_map, :filter_filter, ...]}
```

asummers:
There's probably going to be about 10-12 or them in all (I'm missing a few like map/reduce --> map_reduce, as well as the Stream analogs). Would probably want to blacklist those instead maybe?

rrrene:
@asummers It *cough* only tooks 2.5 years, but this will be part of Credo 1.6 :+1: Thx!

PR #558, opened 2018-06-15
Report violation for not/2 as well in Refactor.NegatedConditionWithElse

whatyouhide:
`!` should be used only when the condition is not strictly boolean. In my experience, most of the time `not/2` is the better choice so I defaulted to `not/2` in the examples (but mentioned `!` as well) and fixed the check so that it checks for `not/2` as well.


rrrene:
@whatyouhide Thx! :+1:

PR #562, opened 2018-06-27
Fix AliasOrder to make it work with “as” option

remi:
Aliases using the `as` option were previously dismissed when extracting alias groups. They are not anymore.


remiprev:
It’s just an UTF-8 parsing error. I’ll send a pull request to remove the faulty `“` `”` characters.

PR #563, opened 2018-06-27
Remove UTF-8 quotation marks from AliasOrder test name

remiprev:
This should fix the OTP 18 and 19 compilation errors I introduced with #562 😬


rrrene:
I wonder if this would classify as a bug in ExUnit ...

Thx! :+1:

PR #566, opened 2018-07-20
Dialyzer does not ignore functions without typespecs

Stratus3D:
Correct the documentation on the value of function typespecs. 
Dialyzer analyzes the types used in all expression whether or 
not the related functions have typespecs. Typespecs only 
assist in the analysis.

http://erlang.org/doc/reference_manual/typespec.html
http://erlang.org/doc/man/dialyzer.html


rrrene:
@Stratus3D Thx!

PR #570, opened 2018-07-30
Add check for unsafe conversions to atoms

seancribbs:
Allowing input from unknown sources to be converted into atoms, if exploited, may cause the virtual machine to halt when the atom table overflows. Except in rare circumstances, one should avoid creating atoms at runtime for this reason, instead relying on conversion of input to _existing_ atoms.


lib/credo/check/warning/unsafe_to_atom.ex
```diff
@@ -0,0 +1,86 @@
+defmodule Credo.Check.Warning.UnsafeToAtom do
+  @moduledoc """
+  Creating atoms from unknown or external sources dynamically is a potentially
+  unsafe operation because atoms are not garbage-collected by the runtime.
+
+  Creating an atom from a string or charlist should be done by using
+
+    String.to_existing_atom(string)
```
rrrene: Since there is literally nothing else to criticise about this PR:

Can you indent the examples in `@moduledoc` with 4 spaces to be consistent with the other checks?

seancribbs: 😆 Of course!

seancribbs: Addressed in a022d92.

rrrene:
> how to avoid warning on values that will be computed at compile-time

Puh, this is a tough one. We could certainly ignore values computed outside of regular functions (i.e. module attribute values, quote blocks, macros, etc.), but that would not fix all instances.

seancribbs:
> We could certainly ignore values computed outside of regular functions

Yeah, after I typed the question I realized that in many cases, excluding macros wouldn't even be sufficient because it's totally fine (and expected) for them to define functions that themselves could create atoms at runtime. We should probably leave it up to the coder to determine whether to silence the warning in places that it is safe (compile-time). Given that, should I apply the magic comment to those safe places inside credo?

seancribbs:
@rrrene Would you review again?

seancribbs:
❤️ 💙 💚

PR #571, opened 2018-08-01
The Spec check shouldn't require functions that implement behaviours to also have a spec

doomspork:
The current `Credo.Check.Readability.Specs` check requires that functions which implement a behaviour (`@impl true`) also include a `@spec`, this PR amends the check to exclude those functions from the final report.

Since the existing check doesn't verify the `@spec` is valid I didn't think it necessary for it to verify that `@impl true` is correct (that you included the behaviour and the function is a callback).

@rrrene what are your thoughts?

### DIFF
--- a/lib/credo/check/readability/specs.ex
+++ b/lib/credo/check/readability/specs.ex
@@ -47,6 +47,14 @@ defmodule Credo.Check.Readability.Specs do
     {ast, [:impl | specs]}
   end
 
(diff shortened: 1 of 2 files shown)


rrrene:
@doomspork Thx for this PR. I'll merge it, as there seems to be no downside to merging.

BUT: Please be aware that the Specs check is deprecated and will be removed in a future version of Credo.

doomspork:
Thanks @rrrene!  Out of curiosity, why do you intend to remove the check?  It's been rather useful for us.

PR #574, opened 2018-08-10
Clarify what should be used in place of a two condition `cond`.

Stratus3D:
An `if` statement should be recommended when a cond with less than two conditions besides `true` is found.


rrrene:
@Stratus3D This is a great clarification! 

Please use `mix format` to format the source code :+1:

Stratus3D:
Fixed!

rrrene:
@Stratus3D Thx!

PR #576, opened 2018-08-26
handle a timeout when reading files

jeremyowensboggs:
Credo in our CI environment has been timing out in load_and_validate_source_files.ex. Here is what is displayed:
```
** (exit) exited in: Task.await(%Task{owner: #PID<0.74.0>, pid: #PID<0.406.0>, ref: #Reference<0.2449986766.523763713.102619>}, 5000)
    ** (EXIT) time out
    (elixir) lib/task.ex:501: Task.await/2
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    lib/credo/cli/task/load_and_validate_source_files.ex:11: anonymous fn/1 in Credo.CLI.Task.LoadAndValidateSourceFiles.call/2
    lib/credo/cli/task/load_and_validate_source_files.ex:9: Credo.CLI.Task.LoadAndValidateSourceFiles.call/2
    lib/credo/cli/command/suggest/suggest_command.ex:15: Credo.CLI.Command.Suggest.SuggestCommand.call/2
    lib/credo/execution/task.ex:45: Credo.Execution.Task.do_run/3
```
I determined that we had one fairly large exs file that we use to mock an external service was causing the issue.  This PR will list any files that timeout, in a manner similar to files that will not parse.

### DIFF
--- a/lib/credo/cli/output.ex
+++ b/lib/credo/cli/output.ex
@@ -131,9 +131,9 @@ defmodule Credo.CLI.Output do
     print_numbered_list(invalid_source_filenames)
   end
 
-  def complain_about_large_source_files([]), do: nil
+  def complain_about_timed_out_source_files([]), do: nil
 
-  def complain_about_large_source_files(large_source_files) do
+  def complain_about_timed_out_source_files(large_source_files) do
     large_source_filenames = Enum.map(large_source_files, & &1.filename)
 
     output = [
--- a/lib/credo/cli/output.ex
+++ b/lib/credo/cli/output.ex
@@ -131,6 +131,25 @@ defmodule Credo.CLI.Output do
     print_numbered_list(invalid_source_filenames)
   end
 
+  def complain_about_large_source_files([]), do: nil
+
+  def complain_about_large_source_files(large_source_files) do
+    large_source_filenames = Enum.map(large_source_files, & &1.filename)
+
+    output = [
+      :reset,
+      :bright,
+      :orange,
+      "info: ",
+      :red,
+      "Some source files were not parsed in the time allotted:\n"
+    ]
+
+    UI.puts(output)
+
+    print_numbered_list(large_source_filenames)
+  end
+
   def print_skipped_checks(%Execution{skipped_checks: []}), do: nil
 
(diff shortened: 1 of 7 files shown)


lib/credo/cli/output.ex
```diff
@@ -131,6 +131,25 @@ defmodule Credo.CLI.Output do
     print_numbered_list(invalid_source_filenames)
   end
 
+  def complain_about_large_source_files([]), do: nil
```
rrrene: Let's call this 

> complain_about_timed_out_source_files

for consistency's sake :+1:

lib/credo/sources.ex
```diff
@@ -113,6 +112,31 @@ defmodule Credo.Sources do
...
 
+  defp read_files(filenames) do
+    tasks = Enum.map(filenames, &Task.async(fn -> to_source_file(&1) end))
+
+    task_dictionary =
+      tasks
+      |> Enum.zip(filenames)
+      |> Enum.into(%{})
+
+    tasks_with_results = Task.yield_many(tasks, 5000)
```
rrrene: Why are you explicitly setting the timeout to its default value?

lib/credo/sources.ex
```diff
@@ -113,6 +112,31 @@ defmodule Credo.Sources do
...
+
+    results =
+      Enum.map(tasks_with_results, fn {task, res} ->
+        # Shutdown the tasks that did not reply nor exit
+        {task, res || Task.shutdown(task, :brutal_kill)}
+      end)
+
+    completed =
```
rrrene: I think we can lose this assignment, since it's already the last statement in the function.

lib/credo/cli/output.ex
jeremyowensboggs: sounds good - code has been updated.
BTW, the files size is 1,777,731 - it contains cached results from an external service that we use for testing.

rrrene:
Thx @jeremyowensboggs :+1:

PR #584, opened 2018-09-13
Enable 'safe' mode for parsing untrusted .credo.exs

grossvogel:
Connects to #584 

- Add a cli option `--safe-mode` to forbid execution of non-credo code while linting
  - Parse config files via reflection instead of `Code. eval_string()`
  - Throw a config error if there are any `requires` in the config
  - Do not accept any checks outside of `Credo.Check` (Undefined checks starting with `Credo.Check` will be ignored with a message, via [validate_config](https://github.com/rrrene/credo/blob/master/lib/credo/execution/task/validate_config.ex#L58)
  - Informative error messages when something can't be parsed because of safe mode
- Update README to mention both `--safe-mode` and `--config-file`

I know there are likely issues I missed around advanced config or use cases I'm not familiar with, so any input is welcome.


lib/credo/exs_loader.ex
```diff
@@ -51,18 +51,24 @@ defmodule Credo.ExsLoader do
...
-  defp process_exs({:__aliases__, _meta, name_list}) do
-    Module.safe_concat(name_list)
+  defp process_exs({:__aliases__, _meta, name_list = [:Credo, :Check | _]}) do
+    Module.concat(name_list)
   end
 
-  defp process_exs({{:__aliases__, _meta, name_list}, options}) do
-    {Module.safe_concat(name_list), process_exs(options)}
+  defp process_exs({{:__aliases__, _meta, name_list = [:Credo, :Check | _]}, options}) do
+    {Module.concat(name_list), process_exs(options)}
```
grossvogel: I'm not sure this is the best way to handle this, but here's my thinking...

The error that comes out of `safe_concat` for an undefined check is rather unfriendly:
```
  :erlang.binary_to_existing_atom("Elixir.Credo.Check.Warning.NameRedeclarationByFn", :utf8)
```
But there is a check later in the `validate_configs` task to warn about and skip undefined checks:
```
  :erlang.binary_to_existing_atom("Elixir.Credo.Check.Warning.NameRedeclarationByFn", :utf8)
```

Also, maybe only things we know are checks should be run inside of Credo in safe mode... not any arbitrary code that happens to be defined with the correct signature.

grossvogel: Thank for the comment, @rrrene, that's the just perspective I was missing.

I guess to me a `safe` mode isn't very safe if you can still require and execute arbitrary code, so I think I'd like to update the PR to revert the default behavior, reverse and rename the CLI option (`--safe-mode` or `--no-outside-code` or ???), and ignore `requires`.

Would that be acceptable?

rrrene: `--safe-mode` sounds good :+1:

lib/credo/exs_loader.ex
```diff
@@ -51,18 +51,32 @@ defmodule Credo.ExsLoader do
     process_tuple(body, {})
   end
 
+  defp process_exs({:__aliases__, _meta, name_list = [:Credo, :Check | _]}) do
+    Module.concat(name_list)
```
grossvogel: Is it OK to use `concat` instead of `safe_concat` here? It seems appropriate to let unknown checks be handled by the nice logic in `validate_configs`, I just don't want to overlook a vulnerability of some kind...

grossvogel:
@rrrene Thanks for your help so far! I think this is ready for another look

grossvogel:
@rrrene I just updated this branch to the latest master. Let me know what else I can do to move it forward, if you're still open to the idea of a `--safe-mode` option. Thanks!

rrrene:
Hi, sorry, but I'm closing this since Credo has changed a lot since this PR was opened.

If you are still interested in making this work, you can probably write a plugin to avoid executing code in configs and requires. :+1:

PR #587, opened 2018-10-01
Check that the filename and module are consistent

### LINKED ISSUE #568: Proposal: check for modulename vs filename consistency
It would be awesome to have a tool that ensures that module names and filenames follow a certain convention, e.g. `ModuleX.ModuleY.ModuleZ` sits in `module_x/module_y/module_z.ex`, test modules in `/test/` have `Test` suffix and files end in `_test.exs` and so forth.

I am aware that having such a tool that satisfies everyone might not be easy, but maybe some degree of checking of respecting these rules could be devised? Or a full blown alternative to allow to define the conventions that these names should respect (e.g. every `.` in module name is a `/` in path, or different)?

In case you think such feature is not an option or unrealistic, can you think how to practically ensure that some convention is respected throughout the project? I find it quite useful to easily figure out what file a particular module lives in.

doomspork:
This closes #568.  @rrrene / @axelson I'd love to get your feedback on this.  I tried to keep this as simple as possible.


doomspork:
@rrrene / @axelson the more I think about this the more I think perhaps it should live under `readability` instead of `consistency`, what do you think?

test/credo/check/consistency/module_filenames_test.exs
```diff
@@ -0,0 +1,24 @@
...
+  use Credo.TestHelper
+
+  @described_check Credo.Check.Consistency.ModuleFilenames
+
+  test "it should NOT report modules with matching filenames" do
+    """
+    defmodule CredoFilenameTest do
+    end
+    """
+    |> to_source_file("credo_filename_test.ex")
```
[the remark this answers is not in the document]
doomspork: This file does not exist, this is a test helper: https://github.com/rrrene/credo/blob/14df806ad444e9c44adf125c2fb2faef33576aaf/test/test_helper.exs#L39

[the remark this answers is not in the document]
axelson: Ah, that makes sense, especially given the content you're piping into it 😅

lib/credo/check/consistency/module_filenames.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Credo.Check.Consistency.ModuleFilenames do
+  @moduledoc """
+  Filenames should reflect the name of the included module.
```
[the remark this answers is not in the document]
doomspork: I've added two examples from Credo @axelson, what do you think?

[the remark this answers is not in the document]
axelson: Much better, although I would think that "matches" would be a better word than "contains". Because to me contains would mean that a filename of `lib/credo/collector.ex` would work with `Credo.CollectorThingy` since `Credo.CollectorThingy` contains `Credo.Collector`.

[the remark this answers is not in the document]
doomspork: Thank you @axelson!  I have updated it with your feedback.

test/credo/check/consistency/module_filenames_test.exs
rrrene: How about Phoenix contexts, where I have `MyApp.Tasks` and `MyApp.Tasks.Checklist` and the directory structure looks like this:

```
lib/
  my_app/
    tasks/
      checklist.ex
      tasks.ex
```

We should probably not trigger on that. WDYT?

doomspork: I'll add a test for this case and see what we get 😁

doomspork: Actually @rrrene wouldn't the correct structure look like this:

```elixir
lib/
  my_app/
    tasks/
      checklist.ex
    tasks.ex
```

This is what would be expected in most languages.  I don't use Phoenix so I'm not familiar with their practices.

doomspork: @rrrene so we want to allow a Module to sit within the first level folder, right?

`/lib/credo/credo.ex` -> `defmodule Credo`

doomspork: @rrrene I have updated the check and tests to include support for this scenario 👍

rrrene:
Thx for putting this together! :+1:

Unfortunately, I get an error when I use the check on Credo itself:

```
Generated credo app
Checking 295 source files (this might take a while) ...
Error while running Elixir.Credo.Check.Consistency.ModuleFilenames on test/fixtures/file_exclusion/foo.ex
Error while running Elixir.Credo.Check.Consistency.ModuleFilenames on test/fixtures/custom-config.exs
Error while running Elixir.Credo.Check.Consistency.ModuleFilenames on test/fixtures/file_exclusion/bar.ex
Error while running Elixir.Credo.Check.Consistency.ModuleFilenames on test/fixtures/options/foo.ex
Error while running Elixir.Credo.Check.Consistency.ModuleFilenames on lib/credo/execution/process_definition.ex
** (EXIT from #PID<0.73.0>) an exception was raised:
    ** (FunctionClauseError) no function clause matching in Path.join/1
        (elixir) lib/path.ex:481: Path.join([])
        lib/credo/check/consistency/module_filenames.ex:46: Credo.Check.Consistency.ModuleFilenames.expected_path/1
        lib/credo/check/consistency/module_filenames.ex:23: Credo.Check.Consistency.ModuleFilenames.run/2
        lib/credo/check/runner.ex:164: Credo.Check.Runner.do_run_check_on_single_source_file/3
        lib/credo/check/runner.ex:157: Credo.Check.Runner.run_check_on_single_source_file/3
        (elixir) lib/enum.ex:1314: Enum."-map/2-lists^map/1-0-"/2
        lib/credo/check/runner.ex:35: Credo.Check.Runner.run_checks_and_append_issues/3
        (elixir) lib/task/supervised.ex:89: Task.Supervised.do_apply/2


21:09:56.564 [error] Task #PID<0.1910.0> started from #PID<0.73.0> terminating
** (FunctionClauseError) no function clause matching in Path.join/1
    (elixir) lib/path.ex:481: Path.join([])
    lib/credo/check/consistency/module_filenames.ex:46: Credo.Check.Consistency.ModuleFilenames.expected_path/1
    lib/credo/check/consistency/module_filenames.ex:23: Credo.Check.Consistency.ModuleFilenames.run/2
    lib/credo/check/runner.ex:164: Credo.Check.Runner.do_run_check_on_single_source_file/3
    lib/credo/check/runner.ex:157: Credo.Check.Runner.run_check_on_single_source_file/3
    (elixir) lib/enum.ex:1314: Enum."-map/2-lists^map/1-0-"/2
    lib/credo/check/runner.ex:35: Credo.Check.Runner.run_checks_and_append_issues/3
    (elixir) lib/task/supervised.ex:89: Task.Supervised.do_apply/2
Function: #Function<9.116522749/0 in Credo.Check.Runner.run/2>
    Args: []
```

Could you take another look?

doomspork:
@rrrene what did you run exactly?  What version of Elixir are you on?

### $ mix compile
```shell
credo git:(feature/module-name-check) mix compile
Compiling 108 files (.ex)
Generated credo app
```

### $ mix credo
```shell
(venv) ➜  credo git:(feature/module-name-check) mix credo
Checking 295 source files (this might take a while) ...

  Software Design
┃
┃ [D] ↑ Duplicate code found in
┃       lib/credo/check/warning/unused_keyword_operation.ex:19,
┃       lib/credo/check/warning/unused_list_operation.ex:19 (mass: 61).
┃       lib/credo/check/warning/unused_tuple_operation.ex:19 #(Credo.Check.Warning.UnusedTupleOperation.run)
┃ [D] ↑ Duplicate code found in
┃       lib/credo/check/warning/unused_file_operation.ex:20,
┃       lib/credo/check/warning/unused_path_operation.ex:20 (mass: 63).
┃       lib/credo/check/warning/unused_regex_operation.ex:20 #(Credo.Check.Warning.UnusedRegexOperation.run)
┃ [D] ↑ Duplicate code found in
┃       lib/credo/check/warning/unused_file_operation.ex:20,
┃       lib/credo/check/warning/unused_regex_operation.ex:20 (mass: 63).
┃       lib/credo/check/warning/unused_path_operation.ex:20 #(Credo.Check.Warning.UnusedPathOperation.run)
┃ [D] ↑ Duplicate code found in
┃       lib/credo/check/warning/unused_keyword_operation.ex:19,
┃       lib/credo/check/warning/unused_tuple_operation.ex:19 (mass: 61).
┃       lib/credo/check/warning/unused_list_operation.ex:19 #(Credo.Check.Warning.UnusedListOperation.run)
┃ [D] ↑ Duplicate code found in
┃       lib/credo/check/warning/unused_list_operation.ex:19,
┃       lib/credo/check/warning/unused_tuple_operation.ex:19 (mass: 61).
┃       lib/credo/check/warning/unused_keyword_operation.ex:19 #(Credo.Check.Warning.UnusedKeywordOperation.run)
┃  ...  (23 more, use `-a` to show them)

  Code Readability
┃
┃ [R] ↗ Module names should be written in PascalCase.
┃       lib/credo/code/token.ex:231:13 #(Credo.Code.Token.ElixirPre1_6_0)
┃ [R] ↗ Large numbers should be written with underscores: 50_000
┃       test/credo/check/readability/large_numbers_test.exs:136:58 #(Credo.Check.Readability.LargeNumbersTest)
┃ [R] → Modules should have a @moduledoc tag.
┃       lib/credo/sources.ex:1:11 #(Credo.Sources)
┃ [R] → Modules should have a @moduledoc tag.
┃       lib/credo/service/ets_table_helper.ex:1:11 #(Credo.Service.ETSTableHelper)
┃ [R] → Modules should have a @moduledoc tag.
┃       lib/credo/service/commands.ex:1:11 #(Credo.Service.Commands)
┃  ...  (67 more, use `-a` to show them)

  Refactoring opportunities
┃
┃ [F] → Function body is nested too deep (max depth is 2, was 3).
┃       lib/credo/cli/command/list/output/default.ex:147:11 #(Credo.CLI.Command.List.Output.Default.print_issue)
┃ [F] → Function body is nested too deep (max depth is 2, was 3).
┃       lib/credo/cli/command/explain/output/default.ex:229:11 #(Credo.CLI.Command.Explain.Output.Default.print_issue)
┃ [F] → Function body is nested too deep (max depth is 2, was 4).
┃       lib/credo/check/design/duplicated_code.ex:86:11 #(Credo.Check.Design.DuplicatedCode.append_issues_via_issue_service)
┃ [F] → Pipe chain should start with a raw value.
┃       lib/credo/check/readability/prefer_unquoted_atoms.ex:83 #(Credo.Check.Readability.PreferUnquotedAtoms.safe_atom_name)

Analysis took 6.8 seconds (0.1s to load, 6.7s running checks)
1983 mods/funs, found 4 refactoring opportunities, 72 code readability issues, 28 software design suggestions.

Showing priority issues: ↑ ↗ →  (use `--strict` to show all issues, `--help` for options).
```

rrrene:
@doomspork You have to add your check to the config (`.credo.exs`):

```elixir
        {Credo.Check.Consistency.ModuleFilenames, []},
```

doomspork:
> @doomspork You have to add your check to the config (.credo.exs):

🤦‍♂️  — sorry @rrrene.  I see the issue, it's files that lack `defmodule`.  That's something I hadn't considered.

rrrene:
No problem. Be sure to run `mix credo --strict` :+1:

doomspork:
@rrrene sorry for delay in wrapping up this PRs, work and the holidays got in the way.  I've updated this PR and ran this new rule against the Credo code base.  

I had to fixed a couple of issues that were reported, let me know your thoughts 👍

lib/credo/check/consistency/module_filenames.ex
```diff
@@ -0,0 +1,106 @@
...
+  defp module_filepath(module_name_parts) do
+    module_name_parts
+    |> Enum.map(&module_name_part/1)
+    |> expected_path()
+  end
+
+  defp module_name_part(module_name) do
+    module_name
+    |> to_string()
+    |> String.replace("IEx", "iex")
```
doomspork: This is a special case that isn't easily matched by any other regex.

test/support/test_application.exs
```diff
@@ -1,3 +1,4 @@
+# credo:disable-for-this-file Credo.Check.Consistency.ModuleFilenames
```
doomspork: I'm open to removing this exception and renaming either the file or module.

test/test_helper.exs
```diff
@@ -1,3 +1,4 @@
+# credo:disable-for-this-file Credo.Check.Consistency.ModuleFilenames
```
doomspork: I can remove this exception but I'll need to rename `Credo.TestHelper` and all instances of it to `TestHelper`.

lib/credo/check/consistency/module_filenames.ex
rrrene: Could you elaborate? 

Isn't there an implied rule here? Like, if there are multiple uppercase letters, you can put an underscore, but don't have to (if you do keep acronyms all uppercase, you will run into this frequently, e.g. what's the "correct" filename for `MyApp.JSONHelper`)?

doomspork: @rrrene that is handled by the RegEx below (`~r/(?<=[A-Z])([A-Z][a-z])/`):

```elixir
iex> Example.module_name_part(MyApp.JSONHelper)
"elixir.my_app.json_helper"
```

The reason for this exemption is `IEx` also matches the above RegEx so without we get:

```elixir
iex> Example.module_name_part(MyApp.IExHelper)
"elixir.my_app.i_ex_helper"
```

doomspork: @rrrene I just noticed that someone opened #623 which attempts to implement this same check. You'll notice they do not have this exemption and as a result this change is required:

image

remiprev: @doomspork I am that someone 😄 I just fixed it by adding an `acronyms` option to the check.

https://github.com/rrrene/credo/pull/623/files#diff-012c34f07d670617f8244a0c6d294964R105

yordis:
Will this flag be enabled by default or being under strict mode?

Coming from Rails, where you have to match your classes to file structures, makes it harder to restructure the directory structure and I don't believe that this is a good design to have by default or even in strict mode.

Elixir has the power of structuring your directories and file the way you want it without having to worry about this.

Lately has been using DDD paradigms in my OTPS applications. Domains drive the folder structures and file names, but my module names do not necessarily follow any folder structure or file name.

The directory and files names are structured in a way that it is easier to understand the domain of the application and straightforward to fuzzy find for people that heavily use Vim; simple to navigate.

But the module names are structured in a way that it is friendly to type without having such long names because of the directory structure.

Also, this allows me to move files around and remove the entropy of the file structures without having to refactor my module names because of Credo.

I would like you to be considerate having to generate a Credo config to enable this.

People like me that uses Credo default without any customization would be affected by this, and we will be the one that generates these files instead which I don't this that is fair if that happens.

Embrace what is right about Elixir language by default, I don't have to follow conversations that make the source code harder to maintain in the long term and constraints application designs.

MainShayne233:
Is someone currently working on getting this passing? I'd happily pick it up if not.

doomspork:
@MainShayne233 it works on TravisCI, it's @rrrene's private CI that is failing and I don't have access to it to see what the failures are.

Waiting on feedback from @rrrene as to what is needed next here.

MainShayne233:
@doomspork awesome! happy to jump in if i can help :)

remiprev:
In my opinion, two things from [my implementation]https://github.com/rrrene/credo/pull/623 could be integrated into this one 😄

1. A more user-friendly message when the filename is not consistent:

    Currently it’s:
    ```
    Filepath does not match module namespace
    ```

    While mine is:
    ```
      The module defined in `lib/foo/baz.ex` is not named consistently with the filename. The file should be named either:
      ["lib/foo_bar.ex", "lib/foo/bar.ex"]
    ```
2. Support for acronyms (eg. not having to hardcode `IEx` → `iex` in the check):

    ```elixir
    {Credo.Check.Consistency.ModuleFilenames, [acronyms: ["IEx", "JSONHelper"]]}
    ```

The code for these is already implemented elsewhere so I’m guessing it wouldn’t be too hard to implement it here as well.

@doomspork @rrrene What do you think?

doomspork:
> 2. Support for acronyms (eg. not having to hardcode `IEx` → `iex` in the check):
    ```elixir
    {Credo.Check.Consistency.ModuleFilenames, [acronyms: ["IEx", "JSONHelper"]]}
    ```

I thought about this and I'm not sure about it. Why would you use this instead of excluding the file from this rule instead?  IEx is so prevalent in Elixir that it's a special case. Other filenames should just be exempted like is done with other rules.

doomspork:
@remiprev upon second look, I think I already accounted for situations like `JSONHelper` and they do not require the same exception as `IEx`. See my previous comment here: https://github.com/rrrene/credo/pull/587#discussion_r247593806

rrrene:
@remiprev @doomspork Credo should always try to "work" without configuration using sensible defaults.

Regarding the "acronym problem", we should simply be liberal in what we consider valid (we can always make checks more strict based on user feedback).

Here are there example I would like to see work

```
module name -> possible filename bases

IexHelper -> iex_helper

IEXHelper -> iex_helper | ie_x_helper | i_ex_helper | i_e_x_helper

IEnterMod -> i_enter_mod
```

This is what I meant with "if there are multiple uppercase letters, you can put an underscore, but don't have to" above 👍

rrrene:
> Will this flag be enabled by default or being under strict mode?

How this check integrates into Credo's standard config will be decided based on the final implementation of the check.

doomspork:
@rrrene thanks for the clarification @rrrene! I'll work on integration these changes.  Any thoughts on updating the error message?

yordis:
> How this check integrates into Credo's standard config will be decided based on the final implementation of the check.

We are maintaining a fairly amount of Elixir packages (especially on private) and we are relying on default values of Credo since we do not want to introduce any config based on personal or organizational opinion.

This will introduce the needs of having a config file per project if you add it to the default and/or strict rules.

There are some caveats in this rule, and since most of our projects tend to find the balance between module name and file structure.

The module names do not follow the file structure.

Either you have a massive folder with all the files since the names are not scoped or your module names are super long and tedious to type.

That is pretty much why we wouldn't follow this rule. We have a more DDD folder structure but with a more flat naming for the module names.

Anyway, I hope we could have a conversation before you take that decision.

yordis:
How will the files under the `test` folder be handled?

```
test
  support
    my_module.ex
```

I am not sure what would be the rule for naming the module. I had been doing `PACKAGE_NAME.TestSupport.MODULE_NAME` but it would be `PACKAGE_NAME.Test.MODULE_NAME`.

For the most part, I put my fixtures and everything under `support` folder instead of having that many folders in that level.

```
test
  package_name_here
    // all tests here
  support
     // all support modules here
  test_helper.exs
```

So I am curious about the testing folder and the naming convention.

cc: @doomspork

doomspork:
Howdy @rrrene, thought I'd follow-up and see if there's anything we can do to get this moving again.

rrrene:
> Howdy @rrrene, thought I'd follow-up and see if there's anything we can do to get this moving again.

Hi, I'm not sure if I'm missing something, but it seems that the edge cases we talked about (Phoenix-style context filenames and acrynom based filenames) are not covered by the test suite yet. Could we add those?

doomspork:
Howdy @rrrene I am going to close this PR for now. This was part of a push to get my employer to do more Elixir open source but this has taken so long, they're not onboard with more time being spent on this.

It looks like #623 was closed and released as a standalone package.

PR #588, opened 2018-10-02
Enforce module attribute ordering

### LINKED ISSUE #540: Proposal: check use, import, alias, require having same order across files
Hi there! Thanks for package!
I didn't find rule to check order for directives: use, import, alias, require 
Wish it would be, to make code more consistent and prevent human mistakes
In future we can add option to set particular order, like `{Credo.Check.Readability.ModuleDirectivesOrder, order: [:use, :import, :require, :alias]}`

So, if you don't mind I'll try to implement this and do pull request

Here are examples:
- Valid
`use Ecto.Schema`
`import Ecto.Query`
`alias Ecto.Paging`
`require Logger`

- Also valid 
`import Ecto.Query`
`require Logger`

- Invalid
`require Logger`
`import Ecto.Query`
`alias Ecto.Paging`

What do you think?)

doomspork:
This attempts to close #540.  @rrrene I'd love to get your thoughts.  I'll work to refactor this to be a bit prettier.

### DIFF
--- a/lib/credo/check/consistency/collector.ex
+++ b/lib/credo/check/consistency/collector.ex
@@ -15,7 +15,7 @@ defmodule Credo.Check.Consistency.Collector do
         def run(source_files, exec, params) when is_list(source_files) do
           issue_formatter = &issues_for/3
 
-          @collector.find_and_append_issues(source_files, exec, params, issue_formatter)
+          @collector.find_and_append_issues(source_files, exec, params, &issues_for/3)
         end
 
         defp issues_for(expected, source_file, params) do
--- a/lib/credo/check/consistency/module_attribute_order.ex
+++ b/lib/credo/check/consistency/module_attribute_order.ex
@@ -1,0 +1,102 @@
+defmodule Credo.Check.Consistency.ModuleAttributeOrder do
+  @moduledoc """
+  List module attributes and directives in the following order:
+
+  1.  @moduledoc
+  2.  @behaviour
+  3.  use
+  4.  import
+  5.  alias
+  6.  require
+  7.  defstruct
+  8.  @type
+  9.  @module_attribute
+  10. @callback
+  11. @macrocallback
+  12. @optional_callbacks
+
+  Like all `Readability` issues, this one is not a technical concern.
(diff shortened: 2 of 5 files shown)


doomspork:
With @michaelstalker's help we've updated the rule to not only enforce the correct ordering of attribute types but that within each type, attributes are sorted alphabetically.

doomspork:
I am _not_ applying this to the Credo codebase because there are a number of situations where Credo violates the community styleguide on this and I don't think we want that amount of code churn.

rrrene:
@doomspork I just realized that I did not comment on this yet.

IMHO this should be a consistent check that points out inconsistencies in your ordering, i.e. raise an issue if you are following the community styleguide (or any other styleguide) and then you do not adhere to that order in a module.

WDYT?

doomspork:
Good morning @rrrene!  I hope things are well with you and yours.

I think `consistency` is a good home for this.  Do you have any other thoughts?

doomspork:
Updated 😁

rrrene:
Hi, sorry for not replying earlier.

> If that's the case, we'll need to track every module's attributes order to assert this correct? 

Yep, that's what I wanted to say.

> Won't that be a bit cumbersome given the number of attributes and modules in larger projects?

Yep, that's why I want Credo to do it instead of keeping track myself 😄

doomspork:
Wonderful @rrrene! I will be returning from holiday in a few days and intend to wrap these PRs up 👍

Zeeker:
Hey there, what's the status on this? Can I be of any help? 🙂

doomspork:
Howdy @Zeeker! Please do. I no longer work at RentPath and I'm not sure if they intend to continue work on this either. cc. @michaelstalker

chevinbrown:
@rrrene what's the possibility of this being added? I'd be stoked to use credo instead of a style guide for module attributes. 😄

sascha-wolf:
I'm not René but I plan on starting to work on this soon.

Kevin Brown <notifications@github.com> schrieb am Mo., 5. Aug. 2019, 22:38:

doomspork:
Closing this as #747 did the job :+1:

Issue #591, opened 2018-10-05
Proposal: CodeClimate Checker changes


fazibear:
I'm working on the rewrite of codeclime engine: https://github.com/fazibear/codeclimate-credo

Right now it's quite complicated. It's a separate application that uses `credo` as the dependency. The problem is that `credo` internals are changing. Updating to the new version is not so easy.

Most important thing is `--include` and `--exclude` switches in the command line. 
I need to control these from the command line. Don't want to mess with config file.

Another thing is the output formatter. We have 2 options there.
1) Rewrite it to work with `json` as input, and leave it in engine repo.
2) Push it into `credo` repository. 
What solution would be better?

fazibear:
Anyway I've made two pull requests to push things forward:

1) Codeclimate formatter: https://github.com/rrrene/credo/pull/592
2) Include/Exclude switches https://github.com/rrrene/credo/pull/595

fazibear:
One more things to support codeclimate checker. Not all warnings was redirected to stderr. This PR https://github.com/rrrene/credo/pull/597 fixes that.

fazibear:
Any thoughts on that?

Kitton:
@fazibear @rrrene is it possible to move on a solution for codeclimate somehow?

rrrene:
@Kitton @fazibear I'm unsure where we stand on this. Is this still a problem?

Kitton:
I believe it's blocking from using the last version of credo with codeclimate, is it correct @fazibear ?

PR #592, opened 2018-10-08
Add codeclimate output format

fazibear:
Part of work related to https://github.com/rrrene/credo/issues/591

Add output format for codeclimate checker.


larkinscott:
@rrrene Sorry to bother you again and thanks for your reply on: https://github.com/rrrene/credo/pull/595#issuecomment-573428600

I think we'd still want to get this formatter merged in to get the `codeclimate-credo` plugin working without needing to maintain a fork of `credo` itself. I see some conflicts but am happy to get them cleaned up or open a new PR if that helps. Thanks again!

rrrene:
@larkinscott I really appreciate that there is a CodeClimate integration for Credo as much as, say, SourceLevel's (formerly Plataformatec's Ebert). That being said, merging a single provider's format into Credo's own source code seems odd. For example, having the mapping of specific Credo checks to CodeClimate categories hard-wired inside Credo would tightly couple two otherwise unrelated software projects.

`---format json` should expose all the necessary information to write an external "adapter" (that maps Credo's JSON to whatever the integration needs). 

Maybe you can elaborate why you have to maintain a fork of Credo for your integration with CodeClimate?

NobbZ:
Going from `--format json` and reformatting it into whatever an adapter needs seems to be a nice thing, though it requires to do the work in another OS-process.

Have you considered `--formatter Formatter` where `Formatter` is a module that implements a behaviour?

`--format json` would still be available for those that prefer to deal with JSON in language X, but also through the formatter option those that want to use Elixir or Erlang get real advantage as they can get all the tools they like and you could even provide library functions later on that help dealing with formatting related stuff.

rrrene:
> Have you considered `--formatter Formatter` where `Formatter` is a module that implements a behaviour?

Yes and no. 😄 We'll have to discuss how plugins, introduced in `v1.1` and enhanced in the upcoming `v1.2` can provide both additional as well as different formatting to existing checks (e.g. both new formats like YAML as well as modify the existing output for built-in commands).

But that does not concern the issue discussed here, which is to merge CodeClimate specific code into Credo. :+1:

larkinscott:
@rrrene Good point. In fact, most of our other plugins work this way. I'm not super familiar with the credo codebase so I was just sort of following the pattern this PR set, but I can tweak our plugin to work with credo's JSON output instead. Thanks again!

rrrene:
@larkinscott To expand on @NobbZ's point: You can also run Credo "natively" from your Elixir project `codeclimate-credo` by adding `credo` as a  dependency and invoking [`Credo.CLI.run(argv)`](https://github.com/rrrene/credo/blob/d1e3a56f07f96203d86d6416cefbccbd8684ea13/lib/credo/cli.ex#L24) with `argv` being the command line arguments you would provide. The function returns the found `%Credo.Issue{}` structs. :+1:

PR #593, opened 2018-10-09
Adds missing descriptions for some parameters

bmbferreira:
The `allow_empty_enums` already had a description but it was not inside a `params` array as it seems to be for all the other patterns


rrrene:
Ha! Thx! :+1: 🎉

PR #594, opened 2018-10-09
Continued Credo dogfooding

doomspork:
Howdy @rrrene!  As you can probably tell by now, @michaelstalker and I are using Credo quite a bit at work as of late and have decided it would be good to contribute back!

In addition to the new rules we're proposing to help enforce the [community styleguide](https://github.com/christopheradams/elixir_style_guide), we'd like to help improve Credo however you think most useful.  This PR continues our work on satisfying all of the complaints raised by  invoking `mix credo` on master.

If there's any changes in here you're not keen on I can drop those commits 😁

### DIFF
--- a/lib/credo/check/readability/prefer_unquoted_atoms.ex
+++ b/lib/credo/check/readability/prefer_unquoted_atoms.ex
@@ -79,7 +79,8 @@ defmodule Credo.Check.Readability.PreferUnquotedAtoms do
   end
 
   defp safe_atom_name(token) when is_binary(token) do
-    :elixir_tokenizer.tokenize(':#{token}', 1, [])
+    ':#{token}'
+    |> :elixir_tokenizer.tokenize(1, [])
     |> safe_atom_name(token)
   end
 
--- a/lib/credo/check/warning/unused_enum_operation.ex
+++ b/lib/credo/check/warning/unused_enum_operation.ex
@@ -29,52 +29,22 @@ defmodule Credo.Check.Warning.UnusedEnumOperation do
   @explanation [check: @moduledoc]
   @checked_module :Enum
   @funs_with_return_value ~w(
-    all? any? at chunk chunk chunk_by concat concat count count dedup
-    dedup_by drop drop_while empty? fetch fetch! filter filter_map
-    find find_index find_value flat_map flat_map_reduce group_by
-    intersperse into into join map map_join map_reduce max max_by
-    member min min_by min_max min_max_by partition random reduce
-    reduce reduce_while reject reverse reverse reverse_slice scan
-    scan shuffle slice slice sort sort sort_by split split_while
-    sum take take_every take_random take_while to_list uniq uniq_by
-    unzip with_index zip
-  )a
-
-  alias Credo.Check.Warning.UnusedFunctionReturnHelper
+      all? any? at chunk chunk chunk_by concat concat count count dedup
+      dedup_by drop drop_while empty? fetch fetch! filter filter_map
+      find find_index find_value flat_map flat_map_reduce group_by
+      intersperse into into join map map_join map_reduce max max_by
+      member min min_by min_max min_max_by partition random reduce
+      reduce reduce_while reject reverse reverse reverse_slice scan
+      scan shuffle slice slice sort sort sort_by split split_while
+      sum take take_every take_random take_while to_list uniq uniq_by
+      unzip with_index zip
+    )a
 
   use Credo.Check, base_priority: :high
 
-  @doc false
+  alias Credo.Check.Warning.UnusedOperation
+
   def run(source_file, params \\ []) do
-    issue_meta = IssueMeta.for(source_file, params)
-
-    all_unused_calls =
-      UnusedFunctionReturnHelper.find_unused_calls(
-        source_file,
-        params,
-        [@checked_module],
-        @funs_with_return_value
-      )
-
-    Enum.reduce(all_unused_calls, [], fn invalid_call, issues ->
-      {_, meta, _} = invalid_call
-
-      trigger =
-        invalid_call
-        |> Macro.to_string()
-        |> String.split("(")
-        |> List.first()
-
-      issues ++ [issue_for(issue_meta, meta[:line], trigger)]
-    end)
-  end
-
-  defp issue_for(issue_meta, line_no, trigger) do
-    format_issue(
-      issue_meta,
-      message: "There should be no unused return values for #{trigger}().",
-      trigger: trigger,
-      line_no: line_no
-    )
+    UnusedOperation.run(source_file, params, @checked_module, @funs_with_return_value, &format_issue/2)
   end
 end
--- a/lib/credo/check/design/duplicated_code.ex
+++ b/lib/credo/check/design/duplicated_code.ex
@@ -46,49 +46,30 @@ defmodule Credo.Check.Design.DuplicatedCode do
 
     source_files
     |> duplicate_nodes(mass_threshold)
-    |> append_issues_via_issue_service(
-      source_files,
-      nodes_threshold,
-      params,
-      exec
-    )
+    |> append_issues_via_issue_service(source_files, nodes_threshold, params, exec)
 
     :ok
   end
 
-  defp append_issues_via_issue_service(
-         found_hashes,
-         source_files,
-         nodes_threshold,
-         params,
-         exec
-       )
-       when is_map(found_hashes) do
+  defp append_issues_via_issue_service(found_hashes, source_files, nodes_threshold, params, exec)
+    when is_map(found_hashes) do
     Enum.each(found_hashes, fn {_hash, nodes} ->
       filenames = Enum.map(nodes, & &1.filename)
-
-      Enum.each(source_files, fn source_file ->
-        if Enum.member?(filenames, source_file.filename) do
-          this_node = Enum.find(nodes, &(&1.filename == source_file.filename))
-          other_nodes = List.delete(nodes, this_node)
-
-          issue_meta = IssueMeta.for(source_file, params)
-
-          issue =
-            issue_for(
-              issue_meta,
-              this_node,
-              other_nodes,
-              nodes_threshold,
-              params
-            )
-
-          if issue do
-            Credo.Execution.Issues.append(exec, source_file, issue)
-          end
-        end
-      end)
+      Enum.each(source_files, &new_issue_for_members(filenames, &1, nodes_threshold, nodes, params, exec))
     end)
+  end
+
+  defp new_issue_for_members(filenames, source_file, nodes_threshold, nodes, params, exec) do
+    if Enum.member?(filenames, source_file.filename) do
+      this_node = Enum.find(nodes, &(&1.filename == source_file.filename))
+      other_nodes = List.delete(nodes, this_node)
+      issue_meta = IssueMeta.for(source_file, params)
+      issue = issue_for(issue_meta, this_node, other_nodes, nodes_threshold, params)
+
+      if issue do
+        Credo.Execution.Issues.append(exec, source_file, issue)
+      end
+    end
   end
 
   defp duplicate_nodes(source_files, mass_threshold) do
--- a/lib/credo/check/warning/unused_file_operation.ex
+++ b/lib/credo/check/warning/unused_file_operation.ex
@@ -12,41 +12,11 @@ defmodule Credo.Check.Warning.UnusedFileOperation do
   @checked_module :File
   @funs_with_return_value ~w(cwd cwd! dir? exists? read read! regular? stat stat!)a
 
-  alias Credo.Check.Warning.UnusedFunctionReturnHelper
-
   use Credo.Check, base_priority: :high
 
-  @doc false
+  alias Credo.Check.Warning.UnusedOperation
+
   def run(source_file, params \\ []) do
-    issue_meta = IssueMeta.for(source_file, params)
-
-    all_unused_calls =
-      UnusedFunctionReturnHelper.find_unused_calls(
-        source_file,
-        params,
-        [@checked_module],
-        @funs_with_return_value
-      )
-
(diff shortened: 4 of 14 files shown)


doomspork:
@rrrene would you be open to a PR that removes 1.4 from the Travis build matrix.  It looks like both `ex_doc` and the format check are failing the build under 1.4.

lib/credo/check/readability/prefer_unquoted_atoms.ex
```diff
@@ -79,7 +79,10 @@ defmodule Credo.Check.Readability.PreferUnquotedAtoms do
   end
 
   defp safe_atom_name(token) when is_binary(token) do
-    :elixir_tokenizer.tokenize(':#{token}', 1, [])
```
rrrene: Why is this a problem? Is this one of those cases where Credo is "in the wrong"?

lib/credo/check/warning/unused_enum_operation.ex
```diff
@@ -26,55 +26,18 @@ defmodule Credo.Check.Warning.UnusedEnumOperation do
...
-
-  defp issue_for(issue_meta, line_no, trigger) do
-    format_issue(
-      issue_meta,
-      message: "There should be no unused return values for #{trigger}().",
-      trigger: trigger,
-      line_no: line_no
-    )
-  end
+  use Credo.Check.Warning.UnusedOperation,
```
rrrene: If really like the idea of the `Credo.Check.Warning.UnusedOperation` module.

But I think this should not be using a `use`, but simply call a function in the normal `run/2`:

```elixir
  UnusedOperation.run(source_file, params, @checked_module, @funs_with_return_value)
end
```

Rule of thumb: If you are using `use`, but are not registering a compile hook, take a second look if a function call will suffice :+1:

doomspork: I will make this change, thanks for the pointer @rrrene I like it 😁

lib/credo/check/readability/prefer_unquoted_atoms.ex
doomspork: The original complaint is that the pipeline does not start with a bare value but changing to this also resulted in a different error, using the capture did the trick:

```elixir
':#{token}'
|> :elixir_tokenizer.tokenize(1, [])
|> safe_atom_name(token)
```

michaelstalker: > Is this one of those cases where Credo is "in the wrong"?

I think it depends on whether you count creating a charlist with an interpolated string a bare value. If we'd consider that a bare value, then it may be fruitful to tweak Credo to allow that operation. If you interpolation and charlist creation count as two separate operations, and `':#{token}'` isn't a bare value, then @doomspork's approach seems good.

An alternative is:
```ex
token
|> String.to_atom()
|> :elixir_tokenizer.tokenize(1, [])
|> safe_atom_name(token)
```
I think that should work, since the guard on line 81 ensures that `token` is a binary value.

doomspork: Sweet!  Thanks @rrrene.  I'll rebase and get your other feedback incorporated ASAP.

lib/credo/check/warning/unused_enum_operation.ex
doomspork: @rrrene unless I'm missing something I'm not sure this approach will work 🤔 

`use Credo.Check` in `UnusedOperation` looks for the `@explanation` module attribute.  I looked at implementing `explanation_for` but that didn't do the trick either.  Thoughts?

rrrene: @doomspork Oh, maybe I did not phrase this right. The `use Credo.Check` directive should be inside the check itself. The `UnusedOperation` module should just provide a "helper" for the common parts of all the `Unused<X>Operation` checks.

Here is a more elaborated sketch of what these checks could look like:

```elixir
defmodule Credo.Check.Warning.UnusedEnumOperation do
  @moduledoc """
  ...
  """

  @explanation [check: @moduledoc]
  @checked_module :Enum
  @funs_with_return_value ~w(
    ...
  )a

  use Credo.Check, base_priority: :high

  alias Credo.Check.Warning.UnusedOperation

    UnusedOperation.run(source_file, params, @checked_module, @funs_with_return_value)
  end
end
```

doomspork: @rrrene I swear I've used Elixir before but I'm not seeing the light here 😆

If I take that approach I end up with other errors because `use Credo.Check`  provides access to `format_issue/2` which `UnusedOperation` needs:

```shell
== Compilation error in file lib/credo/check/warning/unused_operation.ex ==
** (CompileError) lib/credo/check/warning/unused_operation.ex:35: undefined function format_issue/2
    (stdlib) lists.erl:1338: :lists.foreach/2
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
```

With some tweaks I can get it compile but the result is more warnings about module attributes:

```elixir
warning: variable "explanation" is unused
  lib/credo/check/warning/unused_operation.ex:10

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_file_operation.ex:11: Credo.Check.Warning.UnusedFileOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_file_operation.ex:11: Credo.Check.Warning.UnusedFileOperation.explanation_for_params/0

warning: this clause cannot match because a previous clause at line 11 always matches
  lib/credo/check/warning/unused_file_operation.ex:19

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_file_operation.ex:15

warning: function Credo.Check.Warning.UnusedOperation.format_issue/2 is undefined or private
  lib/credo/check/warning/unused_operation.ex:35
```

Unfortunately it seems the way `use Credo.Check` is setup won't allow me to take this approach without a `use` macro to handle the generated code.

It may not be worth spending much more time on this PR if we want to avoid `use UnusedOperation` and not go crazy refactoring `Credo.Check`.

rrrene: @doomspork You are right. We have to hand down the `format_issue/2` function for this to work:

```elixir
defmodule Credo.Check.Warning.UnusedEnumOperation do
  @moduledoc """
  ...
  """

  @explanation [check: @moduledoc]
  @checked_module :Enum
  @funs_with_return_value ~w(...)a

  use Credo.Check, base_priority: :high

  alias Credo.Check.Warning.UnusedOperation

    UnusedOperation.run(
      source_file,
      params,
      @checked_module,
      @funs_with_return_value,
      &format_issue/2
    )
  end
end


defmodule Credo.Check.Warning.UnusedOperation do
  @moduledoc """
  The result of a call to the provided module's functions has to be used.
  """

  alias Credo.Check.Warning.UnusedFunctionReturnHelper
  alias Credo.IssueMeta

  @doc false
  def run(source_file, params, checked_module, funs_with_return_value, format_issue_fun) do

    all_unused_calls =
      UnusedFunctionReturnHelper.find_unused_calls(
        source_file,
        params,
        [checked_module],
        funs_with_return_value
      )

    Enum.reduce(all_unused_calls, [], fn invalid_call, issues ->
      {_, meta, _} = invalid_call

      trigger =
        invalid_call
        |> Macro.to_string()
        |> String.split("(")
        |> List.first()

      issues ++ [issue_for(format_issue_fun, issue_meta, meta[:line], trigger)]
    end)
  end

  defp issue_for(format_issue_fun, issue_meta, line_no, trigger) do
    format_issue_fun.(
      issue_meta,
      message: "There should be no unused return values for `#{trigger}`.",
      trigger: trigger,
      line_no: line_no
    )
  end
end
```

doomspork: If you're okay with that I'm happy to make the change.  I was trying to refrain from taking too many liberties with how these checks are structure, don't want to indirectly contribute to future debt 😁

doomspork: Unfortunately the suggested change does not work either.  I also tried passing in `&issue_for/3` but it still complains.  I'm guessing however it's calling `@explanation` is somehow resolving to being the `UnusedOperation` module each time.

I'm afraid these changes may not be worth the headache:

```shell
warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_enum_operation.ex:29: Credo.Check.Warning.UnusedEnumOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_enum_operation.ex:29: Credo.Check.Warning.UnusedEnumOperation.explanation_for_params/0

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_enum_operation.ex:33

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_file_operation.ex:11: Credo.Check.Warning.UnusedFileOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_file_operation.ex:11: Credo.Check.Warning.UnusedFileOperation.explanation_for_params/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_keyword_operation.ex:11: Credo.Check.Warning.UnusedKeywordOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_keyword_operation.ex:11: Credo.Check.Warning.UnusedKeywordOperation.explanation_for_params/0

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_file_operation.ex:15

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_list_operation.ex:11: Credo.Check.Warning.UnusedListOperation.explanation/0

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_keyword_operation.ex:15

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_list_operation.ex:11: Credo.Check.Warning.UnusedListOperation.explanation_for_params/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_path_operation.ex:11: Credo.Check.Warning.UnusedPathOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_path_operation.ex:11: Credo.Check.Warning.UnusedPathOperation.explanation_for_params/0

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_list_operation.ex:15

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_path_operation.ex:15

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_regex_operation.ex:11: Credo.Check.Warning.UnusedRegexOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_tuple_operation.ex:11: Credo.Check.Warning.UnusedTupleOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_string_operation.ex:27: Credo.Check.Warning.UnusedStringOperation.explanation/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_regex_operation.ex:11: Credo.Check.Warning.UnusedRegexOperation.explanation_for_params/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_tuple_operation.ex:11: Credo.Check.Warning.UnusedTupleOperation.explanation_for_params/0

warning: undefined module attribute @explanation, please remove access to @explanation or explicitly set it before access
  lib/credo/check/warning/unused_string_operation.ex:27: Credo.Check.Warning.UnusedStringOperation.explanation_for_params/0

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_string_operation.ex:31

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_tuple_operation.ex:15

warning: module attribute @explanation was set but never used
  lib/credo/check/warning/unused_regex_operation.ex:15
```

doomspork: @rrrene to make this work I had to move all of the module attributes above `use Credo.Check` which, while maybe not critical, violates the community styleguide check here: https://github.com/rrrene/credo/pull/588

Should we consider passing `explanation` into `use Credo.Check` to avoid requiring `@explanation` happen beforehand?  That certainly tripped me up.  I hadn't encountered an order issue like this before.

doomspork:
@rrrene I incorporated your feedback and now the tests are all passing.  I left a comment but it's now been collapsed so I'll copy it here to make sure you see it:

> @rrrene to make this work I had to move all of the module attributes above `use Credo.Check` which, while maybe not critical, violates the community styleguide check here: #588

> Should we consider passing explanation into use Credo.Check to avoid requiring `@explanation` happen beforehand? That certainly tripped me up. I hadn't encountered an order issue like this before.

doomspork:
Looks like you pushed these commits to master which explains the conflicts.  Looks like you got my `fixup!` commit though not a big deal.

Closing the PR since the change has merged.

rrrene:
@doomspork Yeah, I was happy with the way it was and merged it on the command line.

Thanks for your work on this :+1:

doomspork:
@rrrene you're welcome.  Any thoughts on `@explanation`?  This is the first time I've encountered an issue with the order of my module attributes.  Was there a reason to rely on the module attribute versus passing this into the `use`?

```elixir
use Credo.Check, 
  base_priority: :high, 
  explanation: [check: @moduledoc]
```

PR #595, opened 2018-10-10
Add include/exclude switches

fazibear:
Part of work related to https://github.com/rrrene/credo/issues/591

This PR adds `--include` and `--exclude` switches. 

This is useful when you want to control what needs to be check from the outside.


larkinscott:
:wave: @rrrene 

Scott here from Code Climate. Looking into getting our Credo plugin updated to the most recent stable version. I noticed @fazibear did some work to make Credo compatible with Code Climate. Happy to get this up-to-date with master or prepare a fresh version of the PR, but wasn't sure if there was any other discussion about these changes that I might not be aware of. Thanks!

rrrene:
Hi. I seem to remember a discussion about this, but there is nothing in this PR. Maybe it was happening in an issue or somewhere on Twitter.

Anyhow, sorry for this taking this long. The parameters are now called `--files-included` and `--files-excluded` and they are working like you described in this PR. The implementation had to defer since Credo has "dynamic" switches for the `OptionParser` so that plugins can register their own plugin CLI switches.

This is now on `master`.

You can try this by setting the Credo dep to

```elixir
{:credo, github: "rrrene/credo"}
```

Thx again, especially to @fazibear for raising this issue :+1:

rrrene:
FYI: This is now also available via Hex in Credo `v.1.2.0-rc4`.

PR #597, opened 2018-10-11
Write all warnings to stderr

fazibear:
Because codeclimate engine read data from stdin, all warnings should go to stderr.

This PR fixes this problem.


rrrene:
This also makes a lot of sense apart from CodeClimate! Thanks! :+1:

Issue #561, opened 2018-06-26
ABC size with Ecto query


Apelsinka223:
### Environment

* Credo version (`mix credo -v`): 0.9.0-rc8
* Erlang/Elixir version (`elixir -v`): 1.6.1
* Operating system: macOS

Got a warning on this function
```
def fun() do
  Favorite
  |> where(user_id: ^user.id)
  |> join(:left, [f], t in Template, f.entity_id == t.id and f.entity_type == "template")
  |> join(:left, [f, t], d in Document, f.entity_id == d.id and f.entity_type == "document")
  |> join(:left, [f, t, d], dt in Template, dt.id == d.template_id)
  |> join(:left, [f, t, d, dt], c in Category, c.id == t.category_id or c.id == dt.category_id)
  |> select([f, t, d, dt, c], c)
  |> distinct(true)
  |> Repo.all()
end
```

`Function is too complex (ABC size is 34, max is 30).`

Not sure if it is correct behavior, but I didn't expect that Ecto query will be taken into ABC size

rrrene:
@Apelsinka223 You are right! :+1:

roehst:
@rrrene want me to take a look at how this could work smoother?

rrrene:
@roehst sure. The challenge will be how we can differentiate between cases where the check should raise an issue and where it doesn't. What are your thoughts on a solution here? White-listing some funs? Looking forward to hearing your thoughts!

roehst:
I will first look at how ABC works and what in a ecto SQL query command value makes it go up in each of the A, B and C counts. Then I'll come up with ideas. working on it.

roehst:
Can you assign me to it?

roehst:
Well one thing for certain is that here where are analyzing SQL and not Elixir because that's wthat the query boils down to. It is just running through the ABC counter because it is expressed as a DSL.

White-listing the Ecto DSL could be reasonable - it is not Elixir code - but then it brings the question, is the SQL query itself too complex? Would that be a good feature for Credo, I mean, split ABC into ABC for elixir code in general and other one for the Ecto or DSLs in general? I don't like adding support for DSLs in general as that would widen the scope of the tool beyond what it is now, and checking Ecto DSL would be like adding support not only for Elixir but for SQL as well, I will first try and check what in the example is making ABC too high and then how to white list specific DSLs.

roehst:
[...]

PR #599, opened 2018-10-15
Add ecto-specific whitelisting to ABC size check

roehst:
- The Ecto DSL inflates the ABC metric as in #561.
- This patch adds minimal white-listing capabilities as a quick fix.


rrrene:
@roehst Thanks for putting this together. :+1:

Unfortunately, I cannot merge this, since your patch would affect any `join`, `where` or `select` functions in the user's codebase. But the idea of listing exclusions is the right one for a quick fix.

I recently put together c43e5a4 as a proof-of-concept. Maybe we can build a more substantial solution off of it.

WDYT?

roehst:
I commented this on the commit you made. I am new to the workflows of opensource. Should I continue the discussion here or there @rrrene ?

PR #604, opened 2018-10-26
Fix a few more issues shown by mix credo

michaelstalker:
This PR doesn't fix more than a handful of issues, but here's what it does:
* Ignore a readability issue. We want a module name to have underscores when designating an Elixir version.
* Remove deep nesting.
* Allow `IO.inspect/1`, but in a way that `mix format` will not undo. `mix format` added a space before `credo:disable-for-previous-line`, so we needed a different way to ignore that `IO.inspect/1` call.

There are a few readability issues flagged by `mix credo` which involve missing `@moduledoc`s. Would you like me to work on documenting those modules, or would you prefer `@moduledoc false`? Additionally, `mix credo` is flagging a 18 TODOs and one FIXME. Would you like me to convert those to GitHub issues, and remove the TODO and FIXME comments from the codebase? I can do one or both of those things in separate PRs if you'd like me to.

### DIFF
--- a/lib/credo/code/token.ex
+++ b/lib/credo/code/token.ex
@@ -228,6 +228,7 @@ defmodule Credo.Code.Token do
     end
   end
 
+  # credo:disable-for-next-line Credo.Check.Readability.ModuleNames
   defmodule ElixirPre1_6_0 do
     @moduledoc false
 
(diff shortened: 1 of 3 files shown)


rrrene:
@michaelstalker Could you rebase against the upstream master?

michaelstalker:
@rrrene Done! 😄

rrrene:
@michaelstalker Thank you! :+1:

Issue #602, opened 2018-10-23
ABC Size incorrect with function calls


roehst:
### Environment

* Credo version (`mix credo -v`): 0.10.2
* Erlang/Elixir version (`elixir -v`): elixir 171 compiled with otp 19
* Operating system: Ubuntu

### What were you trying to do?

I was working on the #561 issue and while working on it i noticed that there were function calls that the ABC check was missing. Digging deeper I noticed that function calls with parameter lists go untouched.

The bug can be reproduced with this simple Gist: https://gist.github.com/roehst/24a676385415d8dbb72ff8ab2774cd22.

The following code has a function call:

    def foo(a, b) do
        other_fun a, b
    end

But its ABC size is measured as 0.0 while it should be 1.0.

In comparison, try the following code:

    def foo(a, b) do
        other_fun
    end

And its ABC size is measured correctly as 1.0.

To my best understanding of the problem, there isn't a clause that matches on {atom, _meta, list} to handle this case of a function with arguments. I have a fix available. I would just have to normalize it to match your code style. There are also a few other minor improvements on handling bound names as a stack that I could submit into a new PR. 

If you are interested in something a bit deeper I worked a bit on annotating the Elixir AST's meta field with syntatic information while debugging the ABC check, but it could be useful in other places as well.

rrrene:
> I have a fix available

Please submit a PR! I did not get to respond on my commit where you hinted at all this (sorry for that).

> I worked a bit on annotating the Elixir AST's meta field with syntatic information

I would be interested in this as well. :+1:

roehst:
@rrrene here is the branch I'm working on https://github.com/roehst/credo/tree/ast-annotations

It is in my fork, but the diff should be visible here: https://github.com/roehst/credo/pull/1

roehst:
@rrrene  The fix for this one is in #606

axelson:
Since #606 is merged, I believe that this can be closed.

rrrene:
@axelson Thx for pointing this out! :+1:

PR #605, opened 2018-10-29
Fix bugs on function calls in ABC check

roehst:
There was no clause for the AST pattern {name, _meta, args} with
name being an atom and args a list, which is what local function
calls look like.

I added two tests.

:if and :__aliases__ syntatically are function calls, so
I had to rely on a list of non-function calls to skip for now.

Refers to #602


roehst:
closing - i missed a conflict, will fix

PR #606, opened 2018-10-29
Fix bug on counting function calls in ABC

roehst:
The {atom, _, list} pattern matches function calls but
was not counting any points as branches in the B component
of the metric.

This commit fixes it.

Also, I had to add a whitelist (@non_calls) so that :__aliases__,
:__block__ and other constructs that syntactically are functions
calls are properly skipped.

I added two tests to check for the new cases.


roehst:
Refers to #602

PR #608, opened 2018-10-29
 Use call scope in ABC for excluded functions

roehst:
This commit refactors the ABCSize check.

The ABCSize has an option to ignore any scores when inside calls to
"excluded functions", so that, for example, the Ecto DSL will not
drive the metric upwards.

The previous code removes from the AST anything within a call to
a function marked as excluded, using the AST (under analysis) to
keep track of state in the traversal.

This version pushes the function name to a stack modeling the scope,
and checks the stack for any excluded function when scoring the code,
so the state of the traversal becomes separated from the AST itself.

It is not exactly a problem, but I would love feedback on the result.

### DIFF
--- a/lib/credo/check/refactor/abc_size.ex
+++ b/lib/credo/check/refactor/abc_size.ex
@@ -104,10 +104,15 @@ defmodule Credo.Check.Refactor.ABCSize do
   def abc_size_for(nil, _arguments, _excluded_functions), do: 0
 
   def abc_size_for(ast, arguments, excluded_functions) do
-    initial_acc = [a: 0, b: 0, c: 0, var_names: get_parameters(arguments)]
+    initial_acc = [a: 0, b: 0, c: 0, scope: [], depth: 0, var_names: get_parameters(arguments)]
 
-    [a: a, b: b, c: c, var_names: _] =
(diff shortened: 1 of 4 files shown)


rrrene:
@roehst Thanks for putting this together! The code looks solid :+1:

It might be the lack of coffee, but please re-iterate the basic premise of "why do we want this?"

I understand that we are dismissing everything in the traversal after finding an "excluded" function call right now and that this might be bad, but please explain why this is bad and how this PR improves upon the current solution from a user's perspective.

roehst:
@rrrene thanks for the feedback,

It improves upon the user's perspective only indirectly since it might make the code more maintainable and it might also make room for future improvements.

Also, I think the current code was actually skipping the 'where/join/select' call themselves, while they should count a point as calls (but I am not entirely sure because I can't look at the code right now).

I think if we can use the scoping technique successfully here, we can use it in other places, and learning how to work with scopes while analyzing Elixir code is an important capability in itself :) the scope can be used not only for nested function calls but for many other kinds of analysis.

I can take a look at other of the refactoring checks and see if they could benefit from keeping a scope of definitions.

rrrene:
Closing this for age and because the ABCSize check has been moved to the "controversial" section of Credo's default config.

PR #614, opened 2018-12-07
Add new command and checks.

TreyE:
This adds a new check:
Credo.Check.Warning.ExCoverallsUncovered.

If this check is configured to run, and you have also generated a
previous "cover/excoveralls.json" file, it will search for files missing
test coverage.

It also adds a new command:


TreyE:
Ugh, OFC I forgot to add the pretty formatting for the remediation time in the html report.

PR #615, opened 2018-12-07
Add new command and checks.

TreyE:
This adds a new check:
Credo.Check.Warning.ExCoverallsUncovered.

If this check is configured to run, and you have also generated a
previous "cover/excoveralls.json" file, it will search for files missing
test coverage.

It also adds a new command:

### DIFF
--- a/lib/credo/application.ex
+++ b/lib/credo/application.ex
@@ -12,7 +12,8 @@ defmodule Credo.Application do
       worker(Credo.Service.SourceFileScopes, []),
       worker(Credo.Service.SourceFileAST, []),
       worker(Credo.Service.SourceFileLines, []),
-      worker(Credo.Service.SourceFileSource, [])
+      worker(Credo.Service.SourceFileSource, []),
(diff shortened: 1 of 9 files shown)


rrrene:
@TreyE Wow, that's an impressive PR. Thanks for all the work you put into this! 

Unfortunately, I cannot merge this PR at this time, because it is outside the scope of what Credo is supposed to do. Being a static analysis tool, Credo should just "look" at code and not run it nor be dependent on the code being compiled and run by another party.

BUT: We will start working on a plugin system for Credo in 2019 for exactly this kind of commands and checks, to keep the "core" of Credo lean and focused, but also allow users to integrate Dialyzer, Coveralls, CodeClimate, Inch and many other tools into their Credo workflow.

I think the experience you proposed in this PR will be a great showcase for this! Thank you for putting this together! :+1:

TreyE:
Ok, then I'll just submit the report card command separately from the Code Coverage check.

Issue #601, opened 2018-10-23
Proposal: Variable Rebinding is erroneous without a sigil


isaac-rstor:
A friend who uses erlang complained that elixir allows variable rebinding but thought it would be accepable if variable rebinding was allowed when the variable name contains a sigil...  As it turns out, elixir does allow you to name variables with an exclamation point, would this be a useful thing to offer, even if it's not included in credo by default?  I would be happy to contribute a change as I have started doing this in my code.

PR #618, opened 2018-12-30
Bang sigil option

isaac-rstor:
modifies the 'no variable rebinding' to allow for a bang sigil, if the ":allow_bang" option is set.


isaac-rstor:
refers to:

https://github.com/rrrene/credo/issues/601

rrrene:
@isaac-rstor Thx for this PR.

On a personal note: Please be aware that the `VariableRebinding` check is a controversial one that is no longer enabled by default.

isaac-rstor:
you're welcome!  I am aware that it's controversial!  But I think if this becomes an "accepted convention" it makes it less controversial, and it makes it "easier to enable" in my codebase.

PR #619, opened 2018-12-31
Module Report Card

TreyE:
Add a new command.

### DIFF
--- a/lib/credo/cli/command/report_card/output/console.ex
+++ b/lib/credo/cli/command/report_card/output/console.ex
@@ -1,0 +1,81 @@
+defmodule Credo.CLI.Command.ReportCard.Output.Console do
+  @moduledoc false
+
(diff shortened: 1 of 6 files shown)


rrrene:
Hi @TreyE,

thx for putting this together. 

Unfortunately, I can not merge this, as you misread my [last response](
https://github.com/rrrene/credo/pull/615#issuecomment-450379489). As I said there

> We will start working on a plugin system for Credo in 2019 [...] to keep the "core" of Credo lean and focused, but also allow users to integrate [...] other tools into their Credo workflow.

The same sentiment applies to this PR: We are trying to keep Credo lean and focused and adding a new command to add a new layer of analysis results (grades) is not working towards that goal.

That said, I appreciate the passion and effort that went into this, I really do. :+1:

PR #623, opened 2019-01-14
Add Consistency.Filenames check

remiprev:
This is my take on #568. I started working on it before I noticed @doomspork  had already started working on it, sorry 😕 

* We look for files that only have a single top-level module defined in it — files with more than one top-level modules are ignored.

* We convert the top-level module name to a list of valid filenames. After reviewing a few popular libraries, we concluded that for a `Foo.Bar` module, its valid filenames would be `lib/foo/bar.ex`, `lib/foo/bar/bar.ex` or `lib/foo.bar.ex`.

* If the module filename is not in the list, we return an issue with a message containing a list of valid names the developer should be using for the file.

* We have an `excluded_paths` option, to exclude specific paths. For example, we noticed that we usually want to exclude `test/support` in our Elixir applications because they mostly contain helper modules.

* We have an `acronyms` option to replace acronyms such as `IEx` before splitting a module name to parts.

Thank you for your feedback! ✌️


remiprev:
I’m going to close this since there has been zero activity.

I just released the `credo_filename_consistency` if people want to start/keep using the check: https://github.com/mirego/credo_filename_consistency

Cheers!

Issue #621, opened 2019-01-04
Proposal: check number of module dependencies (warn about tightly coupled modules)


mkorszun:
### Environment

* Credo version (`mix credo -v`): `1.0.0`
* Erlang/Elixir version (`elixir -v`): `1.7.3`
* Operating system: `OSX 10.13.4`

### What were you trying to do?

Users are warned whenever their modules are:

* doing too much / many different things
* tightly coupled with too many other modules

by measuring number of module dependencies (cross module function calls).

### Expected outcome

By adding check:
~~~elixir
{Credo.Check.Refactor.ModuleDependencies, max_deps: 10, dependency_namespaces: ["MyApp"], excluded_namespaces: ["MyApp.Application", "MyAppWeb"]},
~~~

And running:
~~~
mix credo --only moduledependencies
~~~

Users should be warned whenever their modules have more dependencies than configured:
~~~bash
mix credo --only moduledependencies
Checking 189 source files (this might take a while) ...

  Refactoring opportunities
┃
┃ [F] → Module ModuleA has too many dependencies: 10
┃       lib/my_app/module_a.ex:1 #(MyApp.ModuleA)
┃ [F] → Module ModuleB has too many dependencies: 12
┃       lib/my_app/module_b.ex:1 #(MyApp.ModuleB)
┃ [F] → Module ModuleC has too many dependencies: 14
┃       lib/my_app/module_c.ex:1 #(MyApp.ModuleC)
~~~

And check warning details:
~~~bash
mix credo explain lib/my_app/module_a.ex:1

Checking 1 source file ...

  MyApp.ModuleA
┃
┃   [F] Category: refactor
┃    →  Priority: normal
┃
┃       Module ModuleA has too many dependencies: 10
┃       lib/my_app/module_a.ex:1 #(MyApp.ModuleA)
┃
┃    __ CODE IN QUESTION
┃
┃     1 defmodule  MyApp.ModuleA do
┃     2   @moduledoc false
┃     3
┃
┃    __ WHY IT MATTERS
┃
┃       Module is most probably doing too much. Consider limiting the number of
┃       module dependencies.
┃
┃       As always: This is just a suggestion. Check the configuration options for
┃       tweaking or disabling this check.
┃
┃    __ CONFIGURATION OPTIONS
┃
┃       To configure this check, use this tuple
┃
┃         {Credo.Check.Refactor.ModuleDependencies, <params>}
┃
┃       with <params> being false or any combination of these keywords:
┃
[...]

PR #628, opened 2019-01-25
Introduce max module dependency check

mkorszun:
This introduces a feature described in https://github.com/rrrene/credo/issues/621.
It is a check to warn whenever modules have more dependencies than configured (defaults to `10`).

Configuration options:

* `max_deps` - Maximum number of module dependencies

* `dependency_namespaces` - Specify which namespaces should be included while counting module dependencies - by default it will count all dependencies (project modules, built-in modules, external dependencies). If you only care about project modules, specify `[MyAppNamespace]`

* `excluded_namespaces` - Specify which namespaces within a project to exclude from this check

* `excluded_paths` - Specify paths within a project to exclude from this check - defaults to `["test"]`

### DIFF
--- a/lib/credo/check/refactor/module_dependencies.ex
+++ b/lib/credo/check/refactor/module_dependencies.ex
@@ -1,0 +1,143 @@
+defmodule Credo.Check.Refactor.ModuleDependencies do
+  @moduledoc false
+
+  @checkdoc """
+  Module is most probably doing too much. Consider limiting the number of
+  module dependencies.
+
+  As always: This is just a suggestion. Check the configuration options for
+  tweaking or disabling this check.
+  """
+  @explanation [
+    check: @checkdoc,
+    params: [
+      max_deps: "Maximum number of module dependencies.",
+      dependency_namespaces: "List of dependency namespaces to include in this check",
+      excluded_namespaces: "List of namespaces to exclude from this check",
+      excluded_paths: "List of paths to exclude from this check"
+    ]
+  ]
+
+  @default_params [
+    max_deps: 10,
+    dependency_namespaces: [],
+    excluded_namespaces: [],
+    excluded_paths: ["test"]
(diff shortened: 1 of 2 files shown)


lib/credo/check/refactor/module_dependencies.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Credo.Check.Refactor.ModuleDependencies do
+  @moduledoc false
+
+  @checkdoc """
+  Module is most probably doing too much. Consider limiting the number of
```
rrrene: Let's rephrase this to be less confident. Maybe:

> This module might be doing too much.

lib/credo/check/refactor/module_dependencies.ex
```diff
@@ -0,0 +1,143 @@
...
+  defp issues_for_module(_, _, _, _, _), do: []
+
+  defp with_fullnames(dependencies, aliases) do
+    dependencies
+    |> Enum.map(&full_name(&1, aliases))
+    |> Enum.uniq()
+  end
+
+  defp filter_namespaces(dependencies, namespaces) do
+    dependencies
```
rrrene: Please remove the pipe on this function call.

lib/credo/check/refactor/module_dependencies.ex
```diff
@@ -0,0 +1,143 @@
...
+    |> Enum.filter(&keep?(&1, namespaces))
+  end
+
+  defp keep?(_module_name, []), do: true
+
+  defp keep?(module_name, namespaces), do: has_namespace?(module_name, namespaces)
+
+  defp has_namespace?(module_name, namespaces) do
+    namespaces
```
rrrene: Please remove the pipe on this function call.

mkorszun:
@rrrene applied all review fixes

PR #638, opened 2019-02-18
Disable logger check on elixir >= 1.7

timbuchwaldt:
As logger is lazy per default on 1.7+ this check is not needed any longer.


rrrene:
@timbuchwaldt Thx Tim :+1:

PR #640, opened 2019-03-07
Add UnnecessaryAliasExpansion check

remiprev:
I noticed we don’t have a check for this kind of code : 

```elixir
alias Foo.{Bar}
```

Alias expansion with braces is great but I think it affects the code readability when it’s used with a list containing a single item; it should be written : 

```elixir
alias Foo.Bar
```

This new `UnnecessaryAliasExpansion` check makes sure this rule is followed 😄


rrrene:
@remiprev Nice work :+1: Thx!

Issue #641, opened 2019-03-08
Proposal: Experimental "Avoid Multi Alias" Rule


asummers:
### What were you trying to do?
One thing that I find when entering a new large code base for the first time is that having multi-aliases, e.g.

```elixir
alias Foo.{Bar, Baz}
```

makes the code significantly harder to grep for things. While I recognize not everyone shares my opinion here (same as with the `@spec` rule, for example), I would like to propose an off-by-default rule to disallow multi aliases. In the example above, it would encourage the use of:

```elixir
alias Foo.Bar
alias Foo.Baz
```

rrrene:
I would merge a PR for the proposed check. :+1:

PR #644, opened 2019-03-22
Add experimental multi-alias expansion warning.

asummers:
Resolves https://github.com/rrrene/credo/issues/641


asummers:
Perfectly willing to entertain alternate name suggestions here too! This was my stab at it, but naming rules is tough.

rrrene:
@asummers Thx! :+1:

Issue #648, opened 2019-04-04
Proposal: Add anonymous functions to CyclomaticComplexity check


sabudaye:
### Environment

* Credo version (`mix credo -v`): 1.0.4
* Erlang/Elixir version (`elixir -v`): Elixir 1.7.4 (compiled with Erlang/OTP 21) [erts-10.2.1]
* Operating system: fedora 29

### What were you trying to do?
I found that CyclomaticComplexity check doesn't count anonymous functions (`fn`, `&`), but I think it is probably better to count them, because they add complexety for sure, as well as `def` or `defp`
1 for `fn`
0.5 for & (because usually it used like this `&(do_something(&1))`)

### Expected outcome
```
    def first_fun(map) do
      map
      |> Enum.reduce(fn(item) -> Enum.map(item, &(&1)) end)
    end
```
has CC 3, 1 for `def`,  1 for `fn`, 1 for `&` pair

### Actual outcome
The code above has CC 1

NobbZ:
As far as I understand [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity), creating anonymous functions do **not** add to CC, as they do not create a branch in the functions call graph.

They have a CC on their own though.

sabudaye:
Agree, anonymous functions do not add branch in call graph. But in this example:
```
    [1, "bin", :atom] |> Enum.map(
    fn
      num when is_integer(num) -> num + 1
      bin when is_binary(bin) -> bin <> "ary"
      atom when is_atom(atom) -> Atom.to_string(atom)
    end) |> Enum.chunk_by(&(&1 == "atom"))
```
there is few brunches, however according to current credo CC of this code - 0
According to my proposal is 2, which is probably wrong from CC definition point of view, but I think it is better than zero.

NobbZ:
Yeah, I think I understand what you are talking about and it makes kind of sense.

But single clause anonymous functions shouldn't give any penalties to the CC.

sabudaye:
Okay, then I'll close my "suggestion" PR. How it should work then? Which function should be the root of CC counter, in that cases:
1. One anonymous function with one branch inside the normal function:
```elixir
    def first_fun(some_map) do
      Enum.map(some_map,
      fn
        num when is_integer(num) -> num + 1
        other -> other
      end)
    end
```
2. Two anonymous function with one branch inside the normal function:
```elixir
def first_fun(some_map) do
  Enum.map(some_map, 
  fn
[...]

PR #649, opened 2019-04-04
Add & and fn to cyclomatic complexety check

sabudaye:
This is my suggestion for #648


sabudaye:
Closed because of https://github.com/rrrene/credo/issues/648#issuecomment-480304377

Issue #410, opened 2017-07-13
Proposal: add an "excluded" option to disable check


fertapric:
### Environment

* Credo version (`mix credo -v`): 0.8.3
* Erlang/Elixir version (`elixir -v`): 1.4.4
* Operating system: Darwin 16.6.0 x86_64

### Proposal

Exclude a specific set of files for a specific check. Here is an example:

```elixir
%{
      # ...other configuration
      checks: [
        {Credo.Check.Consistency.ExceptionNames. exclude: ["lib/foo/bar.ex:13"]},
        {Credo.Check.Consistency.LineEndings, excluded: ["lib/foo/bar.ex", ~r"/test/"]},
```

rrrene:
I can't find the issue where this was proposed the first time, but I must say: I still don't like this idea ...

We should not have too many places where code can be excluded from analysis and your example amplifies my concerns:

* As someone new to the project and probably Elixir, I am sitting in my editor and wonder why the clear violation of `ExceptionNames` is not reported in the instance I am staring at (`lib/foo/bar.ex:13`).
* Why would you want to exclude a single check from *all* tests, but not your production code?

As I said, I really don't think this is a good addition. You can exclude specific lines or single files for specific checks using the config comment syntax and exclude complete folders from the analysis as a whole (e.g. for third party code you don't control).

Do you have other examples/scenarios where you need this? I am willing to change my mind on this, but not because of "another linter provides this".

fertapric:
> As someone new to the project and probably Elixir, I am sitting in my editor and wonder why the clear violation of `ExceptionNames` is not reported in the instance I am staring at (`lib/foo/bar.ex:13`).

I see it the other way around :) 

Everything related to Credo is confined to the file `.credo.exs`: checks, color, requires, paths/files...and **exceptions**. A new user would only need to check that file to know what's going on.

> As I said, I really don't think this is a good addition. You can exclude specific lines or single files for specific checks using the config comment syntax and exclude complete folders from the analysis as a whole (e.g. for third party code you don't control).

I think the inline comments can produce some noise when reading the code. 

As example:

```elixir
defp my_fun() do
  # credo:disable-for-next-line Credo.Check.Warning.IoInspect
  IO.inspect {:we_want_this_inspect_in_production!}
end
```

The inline comment holds the first line of the method, which is the first thing I read when trying to understand the business logic behind the function `my_func`. However, that line is not really that important :)

> Why would you want to exclude a single check from all tests, but not your production code?

[...]

PR #651, opened 2019-04-26
Exclude config option for IO.inspect check.

Ivor:
Add configuration option to exclude filenames matching the`:excluded` patterns for io_inspect check.

For example, to skip the check for `.exs` files: 

  `{Credo.Check.Warning.IoInspect, [excluded: [~r/.exs$/]]},`

### DIFF
--- a/lib/credo/check/warning/io_inspect.ex
+++ b/lib/credo/check/warning/io_inspect.ex
@@ -7,6 +7,14 @@ defmodule Credo.Check.Warning.IoInspect do
 
   This check warns about those calls, because they might have been committed
   in error.
+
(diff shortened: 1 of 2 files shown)


rrrene:
Hi there, thanks for putting this together. 

This goes in the same direction as the discussion in #410 which is why we should unfortunately not merge this functionality for a single check. Maybe you could join the discussion there and provide your perspective? :+1:

Ivor:
Great, I am happy that the discussion is happening. It's much better to do on a credo level. I just saw this would be nice and rolled out the PR. I'm keen to see where this goes. 

Thanks!

PR #652, opened 2019-04-29
Fix DuplicatedCode not matching blocks with external calls

liskin:
DuplicatedCode (and other checks which use Credo.Code.remove_metadata)
would fail to detect duplicities whenever there are any external
function calls (aliases in AST) and the duplicities don't happen to be
at exactly the same line numbers. So credo would detect if two modules
are exactly equal, but adding a "# creeedo:disabled-four-that-foul"
comment on top is enough to disable the check, which doesn't seem very
intentional. :-)

This is because Credo.Code.update_metadata doesn't recurse into

    {{:., …, …}, …, …}

nodes. Fix this by dropping the hand-crafted implementation and use
Macro.prewalk and Macro.update_meta instead.

### DIFF
--- a/lib/credo/code.ex
+++ b/lib/credo/code.ex
@@ -196,50 +196,7 @@ defmodule Credo.Code do
   @doc """
   Returns an AST without its metadata.
   """
-  def remove_metadata(ast) when is_tuple(ast) do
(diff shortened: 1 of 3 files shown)


rrrene:
@liskin Thx! :+1:

PR #661, opened 2019-06-06
Add `SinglePipe` to list of opt-in checks

nathanl:
After finding https://github.com/rrrene/credo/issues/221, I expected this to be listed as available


rrrene:
@nathanl :+1: Good catch!

PR #662, opened 2019-06-11
Add Credo.Check.Consistency.UnusedVariableNames

remiprev:
We want to enforce a single way of naming unused variables, either name them all `_` or give them all meaningful names (eg. `_user`, `_params`, etc.).

What do you think? Do you think this belongs in Credo?

Thank you!


whatyouhide:
I think I'm on board with this one. I'm not a fan of `_` (since it has no context) but it's not something I point out a lot in PRs since it seems a bit too nitpicky, but maybe if the tool does it it's better.

rrrene:
@remiprev Thx! :+1:

PR #666, opened 2019-06-27
add `nil_else` check for else blocks containing only `nil`

zachdaniel:
Adds a check that will fail on an `else` block containing only `nil`.

### DIFF
--- a/lib/credo/check/readability/nil_else.ex
+++ b/lib/credo/check/readability/nil_else.ex
@@ -1,0 +1,52 @@
+defmodule Credo.Check.Readability.NilElse do
+  @moduledoc false
+
+  @checkdoc """
+  Do not use `nil` as a return value of `else` in `if` or `unless`.
+  The function returns `nil` by default, so adding an else block
+  just adds unnecessary code.
+
+  The code in this example ...
+      if x > 5 do
+        x * 2
+      else
+        nil
+      end
+  ... should be refactored to look like this:
+      if x > 5 do
+        x * 2
+      end
+  """
+  @explanation [check: @checkdoc]
+
+  use Credo.Check, base_priority: :high
+
+  @doc false
+  def run(source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse(
+         {conditional, meta, [_, [{:do, _} | [{:else, nil} | _]]]} = ast,
+         issues,
+         issue_meta
+       )
+       when conditional in [:if, :unless] do
+    {ast, [issue_for(issue_meta, meta[:line]) | issues]}
+  end
+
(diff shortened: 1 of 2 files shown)


OvermindDL1:
Ehh, not a fan of that.  If anything there should be a  check on `if`'s containing no `else` block.  Shouldn't rely on a missing `else` being `nil` as it makes it very easy to skip over visually.

zachdaniel:
@OvermindDL1 I agree :) I wouldn't personally enable this check. However, it appears on https://github.com/lexmag/elixir-style-guide this style guide, and I'm working w/ a client that follows it. I figured we could just have it disabled by default.

OvermindDL1:
Heh, as long as it's not default enabled or so, including 'hidden' pathways like that I find to be very bad for maintainability...  ^.^;

zachdaniel:
Yeah, I see where you are coming from. I think that style guide is fairly popular though, so I don't think it would be the kind of thing that no one uses for two years and then someone tries it and it doesn't work 🤣

OvermindDL1:
\<bikeshedding>

> _I think that style guide is fairly popular though_

_Well Elixir also encourages using spaces for indentation and alignment instead of [tabs for indentation and spaces for alignment](http://lea.verou.me/2012/01/why-tabs-are-clearly-superior/), so not all popular things are good.  ^.^_

\</bikshedding>

But I think that style guide basically died when the official formatter came out.  It does still specify some things that the formatter doesn't since the formatter won't change the AST, but I haven't seen anyone link to it in well over a year now.  ^.^;

zachdaniel:
I think credo has a config file for exactly this reason though, to allow toggling these on and off. I could see a case for having a separate repo that just contains checks that exist to conform to the style guide, that way general users only get the "curated" checks that the maintainers think are worthwhile. But since its so easy to leave them off and just let people use them as the please it doesn't seem harmful really.

OvermindDL1:
I personally don't see any problem including them in credo, disabled by default when conflicting like this one.  Credo could have a special section for them in the config documented as for the purposes of the style guide at that link perhaps?

zachdaniel:
Yeah, I was thinking it could be nice to have some kind of switch/tag that says "enable all the lexmag style guide checks" or something like that.

whatyouhide:
I very strongly disagree with this check as I find that when `if` is used without an `else` to leverage the implicit `nil` the code becomes much harder to read. However, maybe the compromise here is to have an option that decides what style you prefer? Maybe it can default to forcing an `else` if the return value of `if` is used, but at least having an option everyone's happy :(

zachdaniel:
I think that what you’re suggesting now should be a separate check, and you can enable this, that, or neither.  I do agree though, I’d rather see an else w/ a nil honestly.``

zachdaniel:
But this implementation is to support a specific style guide, so even though I’m not a fan, I’d still like to see it as an option.

rrrene:

As for this specific check: I don't think giving people a tool to force writing less explicit code is a good idea.

PR #668, opened 2019-06-27
add check to use `true` for always matching cond clause

zachdaniel:
There would technically be a false positive case for this, which is if the always matching cond was a tuple, like `{true, true}` but I don't see why anyone would do that 🤣

### DIFF
--- a/lib/credo/check/readability/always_matching_cond_clause.ex
+++ b/lib/credo/check/readability/always_matching_cond_clause.ex
@@ -1,0 +1,77 @@
+defmodule Credo.Check.Readability.AlwaysMatchingCondClause do
+  @moduledoc false
+
+  @checkdoc """
+  If you provide an "always true" clause in `cond`, it should be the literal value `true`.
+
+  Correct:
+      cond do
+        x > 5 -> 10
+        x < 5 -> 0
+        true -> 5
+      end
+  Incorrect
+      cond do
+        x > 5 -> 10
+        x < 5 -> 0
+        :other -> 5
+      end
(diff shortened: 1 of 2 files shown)


zachdaniel:
I'd prefer this to be an opt-in (disabled by default) check. The rule appears in this style guide: https://github.com/lexmag/elixir-style-guide

OvermindDL1:
I'd allow both `true` and `:else` as both forms see routine usage for the catch-all case in `cond`'s from what I've seen.

zachdaniel:
Perhaps I'll have it take a list of allowed atoms, and have that list default to `[nil, :else]`? That way some can be more specific.

OvermindDL1:
Ah I have seen `nil` used on occasion too, so perhaps `[nil, true, :else]` as the default?

zachdaniel:
Yeah, that would be fine with me :)

NobbZ:
How could `nil` be a valid catch-all?

```
iex(1)> cond do
...(1)>   nil -> :bar
...(1)> end
** (CondClauseError) no cond clause evaluated to a truthy value
```

OvermindDL1:
Oh right, I've seen `nil` to essentially comment out an entire clause...

So just `[true, :else]` then?

zachdaniel:
Made the recommended change :)

rrrene:

As for this specific check: I really like it, and it would be a great consistency check that determines your preferred clause `true`/`:else`   and enforces that. The current implementation let's me mix and match `true` and `:else` throughout a codebase, which is IMHO not preferable.

PR #669, opened 2019-06-27
check to detect piping into anonymous functions

zachdaniel:
Adding this per the style guide here: https://github.com/lexmag/elixir-style-guide

### DIFF
--- a/lib/credo/check/readability/pipes_to_anonymous_functions.ex
+++ b/lib/credo/check/readability/pipes_to_anonymous_functions.ex
@@ -19,7 +19,7 @@ defmodule Credo.Check.Readability.PipesToAnonymousFunctions do
       |> times_2()
       |> times_2()
   Piping into anonymous functions is cluttered and difficult to read.
-  Instead, define a private function and use that instead.
+  Instead, define a private function and use that.
   """
   @explanation [check: @checkdoc]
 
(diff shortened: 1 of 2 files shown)


OvermindDL1:
I personally prefer piping into `case`'s.  ^.^;

```elixir
do_something
|> case do
  {:ok, value} -> blah(value)
  {:error, _reason} = err -> []
|> case do lst? -> more_things(List.wrap(lst)) end
```

Which I'm not sure is better than anon functions, but eh, less parenthesis...  >.>

zachdaniel:
I've never really decided how I feel about piping into cases 🤷‍♂ . But this shouldn't stop that from happening anyway :)

OvermindDL1:
I'm just thinking it might be good to have both in the same check... or perhaps different checks might be better...

zachdaniel:
I think they are different enough to be separate checks. Case statements are a lot more common in pipelines as well, for good reason.

whatyouhide:
Agreed, they feel like separate checks.

lib/credo/check/readability/pipes_to_anonymous_functions.ex
```diff
@@ -0,0 +1,55 @@
...
+      |> (fn i -> i * 2 end).()
+      |> (fn i -> i * 2 end).()
+  ... should be refactored to look like this:
+      defp timex_2(i), do: i * 2
+
+      foo
+      |> times_2()
+      |> times_2()
+  Piping into anonymous functions is cluttered and difficult to read.
+  Instead, define a private function and use that instead.
```
whatyouhide: Repetition of "instead" here. :)

```suggestion
  Instead, define a private function and use that.
```

rrrene:
Hi, I merged this as a new default check for Credo, but somehow this PR does not reflect that (says something about "unknown repository"). Thx for your work on it!

PR #670, opened 2019-06-27
add check to prefer pipes for nested function calls

zachdaniel:
This one was difficult. It might warrant moving some code out to helpers, and might be finicky. I took the code out of the check that pipes must start with raw values, and use that for this. It additionally ignores calls that have `do` blocks, because `def` is a function for example, and so triggers this issue. Basically if you call a single function, and the first argument to that function isn't something that could *start* a pipe, then the function is a candidate for this issue.

### DIFF
--- a/lib/credo/check/readability/prefer_pipes.ex
+++ b/lib/credo/check/readability/prefer_pipes.ex
@@ -1,0 +1,186 @@
+defmodule Credo.Check.Readability.PreferPipes do
+  @moduledoc false
+
+  @checkdoc """
+  Pipes (`|>`) should be preferred when performing nested function calls.
+
+  So while this is fine:
+      foo(bar)
+  The code in this example ...
+      foo(bar(baz))
+  ... should be refactored to look like this:
+      foo
+      |> bar()
+      |> baz()
+  Nesting function calls makes it harder to change the inner values, and harder to read.
+  Instead, move the nested calls to successive pipes.
+  """
+  @explanation [check: @checkdoc]
+
+  @special_cases [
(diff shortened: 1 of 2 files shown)


rrrene:
@zachdaniel Thx for putting these together.

I am wondering if you might want to create a plugin that bundles these checks, so that it is easier for people people to require them "in one go".

We should add a "Featured Plugins" section in the README anyway and these checks would make a great "lexmag Styleguide" plugin for people who have to adhere to that styleguide (and it could be our first "featured plugin").

WDYT?

zachdaniel:
I think that would be a nice addition to credo, for sure. My PRs definitely don't cover all of the lexmag style guide, but I could pretty easily list what is/isn't covered yet, and let the community add more if they'd like. My only concern is that, by having the checks live in that separate plugin, we are adding lots of overhead for someone that just likes a single check, for instance. Not even sure if it sounds like a good idea yet, but what about having the checks live in credo proper, but deactivated by default, and having the plugin just set the checks up/configure them? That way, the actual logic/checks can be leveraged by other plugins/anyone.

Alternatively, I could contribute any substantial code back to credo core as helpers/utilities, and then make the checks use those. That ensures that we don't see a bunch of duplication of code, and other problems that come with spreading the check logic out.

Not sure where I fall on this one. Thoughts?

OvermindDL1:
I personally would love a pre-built set of config files that could event by copied into the current project by a mix credo command.  `lexmag` would of course be one variant but I can think of another as well.  ^.^

rrrene:

PR #673, opened 2019-06-30
Add an option to enforce parens or not in zero-arity defs

whatyouhide:
@rrrene I'm not sure this aligns with the direction you see for Credo. In the Elixir formatter, we went with a clear decision: the formatter is as opinionated as possible, and configuration is the enemy. However Credo seems to be used as a linter more geared towards consistency: for example, look at all the consistency checks which don't enforce a specific style bur rather just a consistent one. For this reason, I think that it might be nice to be able to configure some checks like the one in this PR. I didn't go with any other checks as I want to test the waters first, and get the discussion started on this. Let me know your thoughts! :)

### DIFF
--- a/lib/credo/check/readability/parentheses_on_zero_arity_defs.ex
+++ b/lib/credo/check/readability/parentheses_on_zero_arity_defs.ex
@@ -2,17 +2,19 @@ defmodule Credo.Check.Readability.ParenthesesOnZeroArityDefs do
   @moduledoc false
 
   @checkdoc """
-  Do not use parentheses when defining a function which has no arguments.
+  Either use parentheses or not when defininig a function with no arguments.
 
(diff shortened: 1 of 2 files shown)


whatyouhide:
@rrrene ping! 🚂

rrrene:
Hi, sorry, I was still in vacation-mode.

This is great. Thank you! :+1:

In an ideal world, this would also be a consistency check that automatically catches your most used pattern. But consistency checks are slow (at least for now) and therefore I think this is a great addition! :tada:

rrrene:
@whatyouhide FYI: This was released in the latest version of Credo (`v1.1.1`).

whatyouhide:
Sweet, thank you @rrrene! 💟

PR #676, opened 2019-07-12
Improve Readability.FunctionNames

sasa1977:
Adds some fixes and improvements to `Readability.FunctionNames`:

- functions defined with parentheses are now verified (previously this check was only verifying a function if it accepts zero args and is defined without parens, i.e. `def some_fun do end`)
- functions with guards (`when` clause) are properly handled
- verifies private macros and custom guards
- generates single report for multiclause functions


rrrene:
Thx @sasa1977 :+1:

PR #677, opened 2019-07-17
Fix outdented heredoc warnings

melpon:
`mix credo test.ex` produces outdented heredoc warnings.

test.ex:

```
defmodule Test do
  # ああああ

  @doc """
  a

  b
  """
end
```

result:

```
warning: outdented heredoc line. The contents inside the heredoc should be indented at the same level as the closing """. The following is forbidden:

    def text do
      """
    contents
      """
    end

Instead make sure the contents are indented as much as the heredoc closing:

    def text do
      """
      contents
      """
    end

The current heredoc line is indented too little
  test.ex:6
```

This PR will fix above problem. Other problems may also be fixed.


melpon:
It fails other tests. I changed to WIP.

melpon:
`mix test` is passed at local.

rrrene:
@melpon Thx for this! :+1:

PR #686, opened 2019-08-08
Add check for alias as

### LINKED ISSUE #684: Proposal: report `alias Foo, as: Bar`
I'd like to introduce a new check which reports issues for every usage of `alias SomeMod, as: SomethingElse`. I propose that the check initially starts with the low base priority.

## Background

Recently I've entered a larger codebase where the usage of `alias ..., as:` is frequent and widespread. I personally find such renaming of modules highly confusing. Looking back, I can't recall a single time where I wanted to use the `:as` option.

Unless I'm missing something, credo doesn't have a check for this, and I definitely want to automatically ensure that the `:as` option is not used in our projects. I've already written the basic version, and now I'd like to see if you'd be open to include this check in Credo?

sasa1977:
resolves #684

The new check is set to low priority and it's included in the list of experimental/controversial checks, which means that by default it will not run, and the user must opt in.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -142,6 +142,7 @@
         {Credo.Check.Consistency.MultiAliasImportRequireUse, false},
         {Credo.Check.Consistency.UnusedVariableNames, false},
         {Credo.Check.Design.DuplicatedCode, false},
+        {Credo.Check.Readability.AliasAs, false},
         {Credo.Check.Readability.MultiAlias, false},
         {Credo.Check.Readability.Specs, false},
         {Credo.Check.Readability.SinglePipe, false},
--- a/lib/credo/check/readability/alias_as.ex
+++ b/lib/credo/check/readability/alias_as.ex
@@ -1,0 +1,50 @@
+defmodule Credo.Check.Readability.AliasAs do
+  @moduledoc false
+
+  @checkdoc """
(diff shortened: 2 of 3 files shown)


sasa1977:
Do you need me to do anything else here? I noticed some formatter failures on older versions. From what I can tell, this happens because formatting rules have been slightly changed on 1.9:

```elixir
Interactive Elixir (1.9.0)
iex(1)> Code.format_string!(":::")
[":\"::\""]

Interactive Elixir (1.8.2)
iex(1)> Code.format_string!(":::")
[":::"]
```

This could be fixed by checking for formatting only on the latest Elixir version, but I don't think that change belongs to this PR.

rrrene:
@sasa1977 Thx! I will remove the formatter check from CI for now :+1:

sasa1977:
> I will remove the formatter check from CI for now

I think it would be good to check for formatting, to keep the code consistent. However, since formatting rules may change from one version to another, I think this check should be done only on the official development Elixir version, which is defined in `.tool-versions`. I think that this could be done by adapting the `check_formatted.sh` script in the following way:

1. Read the development Elixir version from `.tool-versions`.
2. Invoke `elixir -e 'IO.puts(System.version())'` to get the current Elixir version.
3. Run the format check only if both versions are the same.

WDYT?

PR #693, opened 2019-08-20
Fix name parsing bug in Credo.Check.Consistency.UnusedVariableNames

remiprev:
There was a bug in my `unused_variable_name?/1` helper function. I fixed it 😄


rrrene:
@remiprev :+1:

Issue #696, opened 2019-08-31
snake case warning triggering on... more ununsual forms.


ityonemo:
### Environment

* Credo version (`mix credo -v`):
1.1.3
* Erlang/Elixir version (`elixir -v`):
1.9.1
* Operating system:
linux

### What were you trying to do?

```
  defmacro @{_, _, _} do:
     ...code...
```

see https://github.com/ityonemo/state_server/blob/master/lib/state_server/at_state_graph_macro.ex

### Expected outcome
no complaint

### Actual outcome
```
Module attribute names should be written in snake_case.
```

ityonemo:
will submit PR soon.

ityonemo:
https://github.com/rrrene/credo/pull/697

rrrene:
Thx you very much for this! 👏 

Your changes are released as part of `v1.1.4`!

PR #697, opened 2019-08-31
fix: fixes issue #696

ityonemo:
sometimes it's useful (but, of course, potentially dangerous) to redefine an "@" parameter inside of a using statement.  This form currently gets trapped by credo.  I attempted a strategy of trapping on the outer context being a *defmacro* statement, but this strategy wound up being simpler.  In general, the "@" AST will not be followed by a bracket except in this singular situation.


ityonemo:
for certain DSLs, not warning on operator defines is useful (e.g. linear algebra dsl which overrides + - *)

PR #706, opened 2019-10-15
Add regex support and umbrella test path regex to excluded_paths for ModuleDependencies.

### LINKED ISSUE #705: ModuleDependencies doesn't ignore umbrella test folders by default
### Environment

* Credo version (`mix credo -v`): `1.1.5`
* Erlang/Elixir version (`elixir -v`): Erlang/OTP 21 [erts-10.2.2] Elixir 1.9.0 (compiled with Erlang/OTP 20)
* Operating system: Mac OS 10.13.6

run `ModuleDependencies` check

### Expected outcome

Ignores umbrella test folders by default

### Actual outcome

warns on too many dependencies in umbrella tests.

TheFirstAvenger:
Adds umbrella app test directories to default ignore in ModuleDependencies check when in umbrella app.

Fixes #705.

### DIFF
--- a/lib/credo/check/refactor/module_dependencies.ex
+++ b/lib/credo/check/refactor/module_dependencies.ex
@@ -14,7 +14,7 @@ defmodule Credo.Check.Refactor.ModuleDependencies do
       max_deps: "Maximum number of module dependencies.",
       dependency_namespaces: "List of dependency namespaces to include in this check",
       excluded_namespaces: "List of namespaces to exclude from this check",
-      excluded_paths: "List of paths to exclude from this check"
+      excluded_paths: "List of paths or regex to exclude from this check"
     ]
   ]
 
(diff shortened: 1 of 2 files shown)


rrrene:
Thank you for putting this together. 👍

I am unfortunately not sure how this fits with Credo's existing code/behaviour: [The `test/` folder in the project's root is explicitly included](https://github.com/rrrene/credo/blob/fa5c374a0887674f1f219deddcdc3d24cabdd604/.credo.exs#L24), but your PR would exclude the umbrella app's `test/` folders?

Could you clarify what your intentions are?

rrrene:
Ah, you are right, I see. 👍

Wouldn't changing `@default_params[:excluded_paths]` to also accept regexes (and making the default value `~r"/test/"`) also accomplish that and make the logic more straightforward? We already accept string and regexes in other configs, so this seems a bit more consistent to me while still achieving your original goal. Thoughts?

TheFirstAvenger:
@rrrene Good idea, much simpler. I have updated the code to accept regexes in addition to the strings it currently accepts.

rrrene:
@TheFirstAvenger 👍 Thx!

PR #707, opened 2019-10-18
Add check to ensure aliases and requires are together in a file.

TheFirstAvenger:
Adds check to ensure aliases and requires are together in a file.

For example:
```
    defmodule Foo do
      require Logger
      alias Foo.Bar

      alias Foo.Baz
      require Integer
      ...
    end
```
should be changed to:
```
    defmodule Foo do
      require Integer
      require Logger

      alias Foo.Bar
      alias Foo.Baz
```
Will fail check if aliases appear on non-consecutive lines within a file, or if requires do likewise.
Ignores code in quotes to allow separate aliases/requires in things like `__using__`.
Accepts multiline multi-alias including surrounding bracket line and newline, and an alias-as that wraps including surrounding newlines.

### DIFF
--- a/test/credo/check/readability/separate_alias_import_require_use_test.exs
+++ b/test/credo/check/readability/separate_alias_import_require_use_test.exs
@@ -1,0 +1,160 @@
+defmodule Credo.Check.Readability.SeparateAliasImportRequireUseTest do
+  use Credo.TestHelper
+
+  @described_check Credo.Check.Readability.SeparateAliasRequire
+
+  test "it should NOT report violation on consecutive aliases" do
+    """
+    defmodule Test do
+      alias App.Module1
+      alias App.Module2
+    end
+    """
+    |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should report violation on separate aliases" do
+    """
+    defmodule Test do
+      alias App.Module1
+
+      alias App.Module2
+    end
+    """
+    |> to_source_file
+    |> assert_issue(@described_check)
+  end
+
+  test "it should NOT report violation on consecutive single-line multi-aliases" do
+    """
+    defmodule Test do
+      alias App.{Module1, Module2}
+      alias App.Module2
+    end
+    """
+    |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should report violation on separate single-line multi-aliases" do
+    """
+    defmodule Test do
+      alias App.{Module1, Module2}
+
+      alias App.Module2
+    end
+    """
+    |> to_source_file
+    |> assert_issue(@described_check)
+  end
+
+  test "it should NOT report violation on consecutive multi-line multi-aliases" do
+    """
+    defmodule Test do
+      alias App
+
+      alias App.{
+        Module1,
+        Module2
+      }
+
+      alias App.Module3
+    end
+    """
+    |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should report violation on separate multi-line multi-aliases" do
+    """
+    defmodule Test do
+      alias App
+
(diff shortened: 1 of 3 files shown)


test/credo/check/readability/separate_alias_import_require_use_test.exs
```diff
@@ -0,0 +1,160 @@
...
+    defmodule Test do
+      alias App.Module1
+      alias App.Module2
+    end
+    """
+    |> to_source_file
+    |> refute_issues(@described_check)
+  end
+
+  test "it should report violation on separate aliases" do
```
rrrene: This spec contradicts [`Credo.Check.Readability.AliasOrder`](https://github.com/rrrene/credo/blob/c34136ef32e2342dfb02e01dfa2da0cc073e7da0/lib/credo/check/readability/alias_order.ex) which states that you can have groups of aliases seperated by a blank line.

TheFirstAvenger: Yes, I see this check as a slightly more opinionated version of the AliasOrder check, but also one that compliments it in this way:

I have had numerous instances in code whereby the first dev adds these aliases:
```elixir
ModuleA
ModuleD
ModuleG
ModuleJ
```
Then later on, that dev or another one comes along and needs `ModuleE` and `ModuleH`, so adds it like this:
```elixir
ModuleA
ModuleD
ModuleG
ModuleJ

ModuleE
ModuleH
```

The next time they need `ModuleB` so it becomes:

```elixir
ModuleA
ModuleD
ModuleG
ModuleJ

ModuleE
ModuleH

ModuleB
```

and the last time they need `ModuleC` and `ModuleK` so it becomes

```elixir
ModuleA
ModuleD
ModuleG
ModuleJ

ModuleE
ModuleH

ModuleB

ModuleC
ModuleK
```

To me, this adding of single lines between sets of aliases in this way is often the quick and easy choice made by the dev rather than figuring out where it goes in the existing group, but it circumvents the intent of `AliasOrder` (which is to make aliases easier to find), so I wanted to add a check that would discourage this practice.

This also helps combat the practice of adding aliases later in the file, something that phoenix context generators do by default. If left unfixed, aliases can be buried hundreds of lines down in the code. This check helps avoid them.

As with all checks, it can be disabled if users want to group aliases, but I personally see the positives outweighing the negatives in most cases.

rrrene: I do not follow. How is this a "more opinionated version of the AliasOrder check, but also one that compliments it"?

As I mentioned, this contradicts the `AliasOrder` check in certain situations. As soon as you have both checks enabled, you cannot satisfy both at once. We could solve this by adding an opt-in (!) param `disallow_groups: true`.

Thoughts?

TheFirstAvenger: >I do not follow. How is this a "more opinionated version of the AliasOrder check, but also one that compliments it"?

The intent of AliasOrder is to make aliases more readable. Even with AliasOrder on, I find multiple alias groups as less readable than combining them all into one group, because you only have to scan one group alphabetically to find an item, whereas if you have 5 groups, you have to scan all 5 for the one you are looking for. You also may miss aliases defined 300 lines into the file because someone decided to add it down there instead of at the top.

When I say it "compliments it", I mean that AliasOrder alone is good, but AliasOrder with this check is even better. With both I have one place to look for aliases in my file, and they are all in order.

>As I mentioned, this contradicts the AliasOrder check in certain situations. As soon as you have both checks enabled, you cannot satisfy both at once

I may be missing something here. If both checks are enabled, you would need to have all your aliases in one group (enforced by this check), and in order within that group (enforced by the AliasOrder). You can still satisfy both at once.

PR #708, opened 2019-10-18
Fix TrailingWhiteSpace check on Windows

### LINKED ISSUE #698: Check "Trailing white-space at end of a line" with Windows
### Environment
* Credo version (`mix credo -v`): 1.1.4
* Erlang/Elixir version (`elixir -v`): 1.7.1
* Operating system: Windows 10

I am running 'mix credo --strict' on my elixir project. I am using Visual Studio Code which saves my files with line endings in Windows style.

### Expected outcome

The check "There should be no trailing white-space at the end of a line" should not tell me my files have trailing whitespaces as they do not have them.

### Actual outcome

The check "There should be no trailing white-space at the end of a line" tells me my files have trailing whitespaces.

Workaround: I can set in Visual Studio Code to save my files with line endings in Linux style. Got this advise from [Elixir Forum](https://elixirforum.com/t/please-help-me-using-credo/). Of course I would rather keep my files with line endings in Windows style as I am using Windows.

iagomoreira:
Fixes the TrailingWhiteSpace check on windows style line endings. 
The current regex `~r/\s+$/` was catching `\r` line endings when it shouldn't catch. 
That commit replace the `\s` with `\h` which is a horizontal whitespace and it's also compliant with PCRE  (Perl Compatible Regular Expressions) that is used by elixir and erlang. 

Fixes #698

### DIFF
--- a/lib/credo/check/readability/trailing_white_space.ex
+++ b/lib/credo/check/readability/trailing_white_space.ex
@@ -45,7 +45,7 @@ defmodule Credo.Check.Readability.TrailingWhiteSpace do
 
   defp traverse_line([{line_no, line} | tail], issues, issue_meta) do
     issues =
-      case Regex.run(~r/\s+$/, line, return: :index) do
+      case Regex.run(~r/\h+$/, line, return: :index) do
         [{column, line_length}] ->
           [issue_for(issue_meta, line_no, column + 1, line_length) | issues]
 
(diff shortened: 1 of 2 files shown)


OvermindDL1:
This actually seems a good check to have, that newlines are only `\n` (probably someone's IDE set up wrong but any half decent one should default to `\n`), but maybe as it's own check that is default enabled?

iagomoreira:
I agree that it should be the IDE default. But it’s not enforced by credo. And there’s actually another check that have the “\r\n” in mind.

rrrene:
@iagomoreira Sorry for the long wait. Much aprreciated! :+1:

PR #711, opened 2019-11-01
Fix crash in AliasAs check when __MODULE__ is aliased. Fixes #710.

### LINKED ISSUE #710: alias __MODULE__, as: Foo crashes AliasAs check
(PR with fix incoming)

Enable `AliasAs` check on a project that had `alias __MODULE__, as: Foo`

### Expected outcome

Successfully warns about use of alias as

### Actual outcome

Credo crashes without indication as to why

TheFirstAvenger:
Fix crash in AliasAs check when `__MODULE__` is aliased. Fixes #710.

### DIFF
--- a/lib/credo/check/readability/alias_as.ex
+++ b/lib/credo/check/readability/alias_as.ex
@@ -34,6 +34,9 @@ defmodule Credo.Check.Readability.AliasAs do
   defp add_issue(issues, nil), do: issues
   defp add_issue(issues, issue), do: [issue | issues]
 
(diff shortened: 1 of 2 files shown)


rrrene:
@TheFirstAvenger Can we put this behind a config param? Something like 

```elixir
{Credo.Check.Readability.AliasAs, [ignore: ["__MODULE__"]]}
```

This way, everyone could ignore the aliasing of the modules they want.

`AliasAs` is an opinionated check as it is, we should avoid adding more opinionated defaults to it 👍

rrrene:
@TheFirstAvenger I misread what this does and have minimized my earlier comment.

Great work :+1:

PR #714, opened 2019-11-07
Make Credo.Check.Readability.MultiAlias check work with submodule expansion

remi:
The `Credo.Check.Readability.MultiAlias` check was raising an error when trying to parse this code because of the `Bar.Baz` part.

```elixir
alias Foo.{Bar.Baz, Quux}
```

I fixed it and added a test.


rrrene:
@remi Thx! :+1:

PR #725, opened 2019-12-09
Remove old CodeHelper references which was removed in v1.0.0

tverlaan:
The alias and the references to this module no longer make sense since the module was removed in v1.0.0


rrrene:
@tverlaan Thx! :+1:

PR #730, opened 2019-12-20
Improve UnsafeToAtom to recognize Jason.decode using keys: :atoms

TheFirstAvenger:
## What

This PR improves the UnsafeToAtom check to recognize and warn about the use of `keys: :atoms` in `Jason.decode` and `Jason.decode!`.

## Why

Jason.decode and Jason.decode! take a `:keys` opt which allows you to choose what type of keys your maps are decoded with. Passing `keys: :atoms` will use the unsafe `String.to_atom` under the hood. Passing `keys: :atoms!` will use the safe `String.to_existing_atom`.

### DIFF
--- a/lib/credo/check/warning/unsafe_to_atom.ex
+++ b/lib/credo/check/warning/unsafe_to_atom.ex
@@ -19,6 +19,14 @@ defmodule Credo.Check.Warning.UnsafeToAtom do
 
       Module.safe_concat([prefix, infix, suffix])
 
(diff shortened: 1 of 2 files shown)


rrrene:
@TheFirstAvenger Nice! I am very much in favor of merging this, :+1:

Please update the tests to be "static" and match the rest of the codebase in that regard, i.e. write out the 12 tests you are generating right now so we have ~100 LoC which we can maintain in a straight forward way (instead of the current ~60 lines of meta-programming).

TheFirstAvenger:
@rrrene done!

PR #733, opened 2020-01-16
Fix issue with interpolation of strings vs charlists. Fixes #729.

### LINKED ISSUE #729: Unexpected parser error
### Environment

* Credo version (`mix credo -v`): `1.1.5`

```
Erlang/OTP 22 [erts-10.5.6] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.9.4 (compiled with Erlang/OTP 20)
```

* Operating system: macOS Catalina

For one codebase I've got a parser error, and reproduced the error in this repo: https://github.com/rodrigues/credo_interpolation_error

```elixir
defmodule CredoInterpolationError do
  @moduledoc ~S"""
  Example to demonstrate a problem with credo 1.1.5 and:
  - elixir 1.9.4
  - otp 22.1.8
  """

  def foo do
    env = Application.get_env(:credo_interpolation_error, :foo)

    case "#{env}" do
      "A" <> _ = env -> "🇿🇼 #{String.upcase(env)}"
      "B" <> _ = env -> "🇻🇺 #{String.upcase(env)}"
      env -> String.upcase(env)
    end
  end
end
```

### Expected outcome

No issues.

### Actual outcome

For some reason this seems to happen when there are flag emojis in the mix. Tried other emojis and couldn't reproduce the error, but it reproduced with other flags 🤔

There's also no actual extra whitespace, when I change to another emoji the code readability warning is also not there anymore.

```
➜ mc
Checking 1 source file ...

info: some checks were skipped because they're not compatible with
your version of Elixir (1.9.4).

You can deactivate these checks by adding this to the `checks` list in your config:

    {Credo.Check.Refactor.MapInto, false},
    {Credo.Check.Warning.LazyLogging, false},

warning: Unexpected error while parsing lib/credo_interpolation_error.ex: [%Credo.Issue{category: :error, check: Credo.Code.ParserError, column: nil, exit_status: 0, filename: "nofilename", line_no: 13, message: "unexpected token: ", meta: [], priority: 0, scope: nil, severity: nil, trigger: nil}]
  lib/credo/check/readability/string_sigils.ex:50: Credo.Check.Readability.StringSigils.run/2
  lib/credo/check/runner.ex:168: Credo.Check.Runner.do_run_check_on_single_source_file/3
  lib/credo/check/runner.ex:161: Credo.Check.Runner.run_check_on_single_source_file/3
  (elixir) lib/enum.ex:1336: Enum."-map/2-lists^map/1-0-"/2
  (elixir) lib/enum.ex:1336: Enum."-map/2-lists^map/1-0-"/2


  Code Readability
┃
┃ [R] ↘ There should be no trailing white-space at the end of a line.
┃       lib/credo_interpolation_error.ex:13:36 #(CredoInterpolationError.foo)

Analysis took 0.06 seconds (0.00s to load, 0.06s running 52 checks)
2 mods/funs, found 1 code readability issue.
```

TheFirstAvenger:
This PR fixes #729. This was a fun one to track down. When we pass code into `:elixir_tokenizer` to identify the token locations, we first convert the string to a charlist:
https://github.com/rrrene/credo/blob/d77eab696f7a7e485ba6ef3c1a7cc41ccd834fba/lib/credo/code.ex#L132-L134
This results in column numbers consistent with the charlist indexes of the characters. For the unicode character in question (🇿🇼), it translates to two values:
```elixir
iex(1)> String.to_charlist("🇿🇼")
[127487, 127484]
```
This means that when the `:elixir_tokenizer` identifies the `#` token in this string:

```elixir
    source = ~S"""
    "🇿🇼 #{String.upcase(env)}"
    """
```
it identifies it at the 5th column in the output, which is correct because it is the 5th item in the charlist:
```elixir
   {:bin_string, {1, 1, nil},
    [
      "🇿🇼 ",
      {{1, 5, nil}, {1, 25, nil},
```

The problem is that our code to split and replace that interpolation uses `String.slice`, which uses the `String` indexes, not `charlist` indexes, so when it grabs the 5th character, it is off by one and starts the replacement at the `{` instead of the `#`. The resulting code has an unexpected/unparsable `#`:
```elixir
     Assertion with == failed
     code:  assert expected == InterpolationHelper.replace_interpolations(source, " ")
     left:  "\"🇿🇼                      \"\n"
     right: "\"🇿🇼 #                     \n"
```

The fix was simply to convert the strings to charlists and use `Enum.slice` instead of `String.slice` to replace the values, as the numbers we are working with are the charlist indexes, not the string indexes.

### DIFF
--- a/lib/credo/code/interpolation_helper.ex
+++ b/lib/credo/code/interpolation_helper.ex
@@ -55,9 +55,11 @@ defmodule Credo.Code.InterpolationHelper do
 
   defp replace_line(line, col_start, col_end, char) do
     length = max(col_end - col_start, 0)
-
-    String.slice(line, 0, col_start - 1) <>
-      String.duplicate(char, length) <> String.slice(line, (col_end - 1)..-1)
+    line = String.to_charlist(line)
+    part1 = Enum.slice(line, 0, col_start - 1)
+    part2 = String.to_charlist(String.duplicate(char, length))
+    part3 = Enum.slice(line, (col_end - 1)..-1)
+    List.to_string(part1 ++ part2 ++ part3)
   end
 
   @doc false
--- a/test/credo/check/readability/string_sigils_test.exs
+++ b/test/credo/check/readability/string_sigils_test.exs
@@ -1,6 +1,8 @@
 defmodule Credo.Check.Readability.StringSigilsTest do
   use Credo.TestHelper
 
+  import ExUnit.CaptureIO
+
   @described_check Credo.Check.Readability.StringSigils
 
   def create_snippet(string_literal) do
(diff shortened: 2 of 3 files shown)


rrrene:
@TheFirstAvenger Nice detective work! :+1:

PR #734, opened 2020-01-16
Include line number in line ending check. Fixes #253.

### LINKED ISSUE #253: Windows line break check doesn't report line number
### Environment

* Credo version (`mix credo -v`): 0.5.3
* Erlang/Elixir version (`elixir -v`): Erlang/OTP 19, Elixir 1.3.4
* Operating system: OSX 10.10.5

### What where you trying to do?

I saved the file lib/credo/check/consistency/line_endings.ex with windows line breaks to see the outcome of mix credo. 

### Expected outcome

I expected the line and column of the first windows line break to be displayed in the report.
e.g. 
```
┃ [C] ↗ File is using unix line endings while most of the files use windows line
┃       endings.
┃       lib/credo/check/consistency/line_endings.ex:1:49 ()
```
### Actual outcome
```
┃ [C] ↗ File is using unix line endings while most of the files use windows line
┃       endings.
┃       lib/credo/check/consistency/line_endings.ex ()
```

TheFirstAvenger:
Include first offending line number in each file for the line ending check. Probably will always be line 1 but we should report it anyway for consistent reports. Fixes #253.

### DIFF
--- a/lib/credo/check/consistency/line_endings.ex
+++ b/lib/credo/check/consistency/line_endings.ex
@@ -18,9 +18,11 @@ defmodule Credo.Check.Consistency.LineEndings do
   end
 
   defp issues_for(expected, source_file, params) do
+    first_line_with_issue = @collector.first_line_with_issue(expected, source_file)
+
     source_file
     |> IssueMeta.for(params)
-    |> format_issue(message: message_for(expected))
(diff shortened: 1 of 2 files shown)


rrrene:
:+1:

PR #736, opened 2020-01-23
Fix typos in SourceFile

SherSpock:


rrrene:
@SherSpock 👍 Thanks! Good catch.

PR #738, opened 2020-01-24
Implement NegatedIsNil check.

pdgonzalez872:
@rrrene, like we discussed here: https://github.com/rrrene/credo-proposals/issues/49

Please let me know if this would add any value! If not, no worries!

### DIFF
--- a/lib/credo/check/refactor/negated_is_nil.ex
+++ b/lib/credo/check/refactor/negated_is_nil.ex
@@ -1,0 +1,54 @@
+defmodule Credo.Check.Refactor.NegatedIsNil do
+  @moduledoc false
+
+  @checkdoc """
+  We should avoid negating the `is_nil` predicate function.
+
+  Here are a couple of examples:
+
+  The code here ...
+
+      def fun(%{external_id: external_id, id: id}) when not is_nil(external_id) do
+         ...
+      end
+
+  ... can be refactored to look like this:
+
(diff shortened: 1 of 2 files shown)


lib/credo/check/refactor/negated_is_nil.ex
```diff
@@ -0,0 +1,71 @@
...
+      if not is_nil(arg) do
+        proceed_as_planned()
+      end
+
+  ... can be refactored to look like this:
+
+      if arg do
+        proceed_as_planned()
+      end
+
```
rrrene: I do not think we should include this in the check, since the refactored piece of code is both losing its explicitness and concerning a very different use-case than the one we discussed in  [your original proposal](https://github.com/rrrene/credo-proposals/issues/49).

There, we were talking about guards which can be refactored to be more explicit under certain conditions, making this a worthwhile opt-in check. 

NOTE: There are reasons why this 

```elixir
def work(arg) when not is_nil(arg), do: more_work(arg)
```

is perfectly reasonable code, thus making the check opt-in is the right way to go.

Please refactor the check to check for guards only, and not concern itself with `if`-statements. 👍

pdgonzalez872: Does it look better now?

rrrene:
@pdgonzalez872  👍 Thx for your contribution!

pdgonzalez872:
@rrrene awesome, thank you!

PR #743, opened 2020-02-04
Enable by default `UnnecessaryAliasExpansion`

linduxed:
The comment suggests that this should already be a default. Let's make it so.


linduxed:
The Travis pipeline seems to have timed out on fetching dependencies. I think it should go green if it is rerun.

rrrene:
Hi, thx for this. You are right 👍

PR #744, opened 2020-02-10
Add WithToken check

sasa1977:
This PR brings in a new readability check which detects `with` token placeholder pattern (not sure if this is the correct name):

```elixir
with {:resource, {:ok, resource}} <- {:resource, Resource.fetch(user)},
      {:authz, :ok} <- {:authz, Resource.authorize(resource, user)} do
  do_something_with(resource)
else
  {:resource, _} -> {:error, :not_found}
  {:authz, _} -> {:error, :unauthorized}
end
```

I've introduced this check a few months ago at my client's projects, because they have been overusing the pattern quite aggressively. The check is therefore tested in production, and it has been working well for us.

The main reason I didn't submit it earlier is b/c I was worried that it might be controversial. But after a few people asked me about it, I figured we could include the check as a low-priority controversial check, so it's turned off by default

I'm not really sure about the name (is there official name for this pattern?). Back in our projects we called this `WithPlaceholder`, and here I went for `WithToken`, but I'm pretty much ok with anything else, especially if there is some more established name.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -163,7 +163,7 @@
         {Credo.Check.Readability.MultiAlias, false},
         {Credo.Check.Readability.Specs, false},
         {Credo.Check.Readability.SinglePipe, false},
-        {Credo.Check.Readability.WithToken, false},
+        {Credo.Check.Readability.WithTaggedTuple, false},
         {Credo.Check.Refactor.ABCSize, false},
         {Credo.Check.Refactor.AppendSingleItem, false},
         {Credo.Check.Refactor.DoubleBooleanNegation, false},
--- a/.credo.exs
+++ b/.credo.exs
@@ -163,6 +163,7 @@
         {Credo.Check.Readability.MultiAlias, false},
         {Credo.Check.Readability.Specs, false},
         {Credo.Check.Readability.SinglePipe, false},
+        {Credo.Check.Readability.WithToken, false},
         {Credo.Check.Refactor.ABCSize, false},
         {Credo.Check.Refactor.AppendSingleItem, false},
         {Credo.Check.Refactor.DoubleBooleanNegation, false},
--- a/lib/credo/check/readability/with_tagged_tuple.ex
+++ b/lib/credo/check/readability/with_tagged_tuple.ex
@@ -1,8 +1,8 @@
-defmodule Credo.Check.Readability.WithToken do
+defmodule Credo.Check.Readability.WithTaggedTuple do
   @moduledoc false
 
   @checkdoc """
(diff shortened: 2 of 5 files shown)


whatyouhide:
@sasa1977 what is the check arguing against? Not really sure from the PR text and I'm lazy to look at the code :P

sasa1977:
@whatyouhide the check is saying that you should replace this code:

```elixir
with {:resource, {:ok, resource}} <- {:resource, Resource.fetch(user)},
     {:authz, :ok} <- {:authz, Resource.authorize(resource, user)} do
  do_something_with(resource)
else
  {:resource, _} -> {:error, :not_found}
  {:authz, _} -> {:error, :unauthorized}
end
```

with something like

```elixir
with {:ok, resource} <- fetch_resource(user),
     :ok <- authorize(resource, user),
     do: do_something_with(resource) # should return {:ok, result} | {:error, :reason}
```

I'm pretty sure not everyone is going to like it (which is why the check is not enabled by default), but the team of my client uses it and they are happy with it (and they have been heavily using the former pattern before).

You can also take a look at [this twitter thread](https://twitter.com/devoncestes/status/1226189034806009858) for some opinions.

whatyouhide:
Personally I am strongly 👎 on this check since the former is a pattern that I like and use sometimes. Regardless of personal preference it seems like a fairly aggressive check but up to you folks :)

sasa1977:
As I said, I'm aware not everyone is going to like it, so I'm not going to debate personal tastes :-)

The more important question is how we regard Credo. I personally like to think of it as a styleguide toolkit. What I mean by this is that credo ships with some styleguide proposal (i.e. the checks which are turned on by default), and then there are some extra opt-in checks. And of course the framework allows building custom checks. In that spirit, I submitted this check as an opt-in (so turned off by default). I think it would be a bit nicer to have this check turned over to community, rather than manag it ourselves, but I'm basically fine either way.

whatyouhide:
@sasa1977 fair point :)

rrrene:
I also like it as a valuable opt-in.

Can we call it `WithTaggedTuples`? I recall these constructs being called "tagged tuples" ([and Google seems to back up my memory](url)).

sasa1977:
I'm not quite sold on `WithTaggedTuple`, because `{:ok, res} | {:error, reason}` is also a tagged tuple, so I think this is somewhat misleading. However, I don't have a better proposal, so I renamed it.

rrrene:
@sasa1977 Thx and good point! I thought about it and went with `WithCustomTaggedTuples` :+1:

PR #745, opened 2020-02-13
Update credo version in readme

progsmile:


rrrene:
Ha! Good catch, thx :+1:

PR #747, opened 2020-02-16
Add module layout check

sasa1977:
This is the first stab at a check which allows us to enforce order of different module parts (e.g. moduledoc should appear before anything else). For context, see [this issue](https://github.com/rrrene/credo-proposals/issues/22#issuecomment-584068632) (the link takes you to the first comment where this particular check has been discussed).

The check is marked with low priority and included in the list of controversial checks.
I wasn't sure which `:category` to pick, so I left it at default.

As discussed in the issue, the default order is: `shortdoc moduledoc behaviour use import alias require`, while other elements can appear in arbitrary order, but they must appear after these elements (or any other elements specified in the `:order` option).

The PR also brings an internal helper which I named `Credo.Check.Module` which can be use to extract various module parts. Currently it's quite elementary (only returns locations of each part), but I believe it could be quite useful for other checks. For example, one of my pet peeves is `Credo.Check.Readability.Specs` which requires specs for callback funs, as well as funs marked with `@doc false`. Using `Credo.Check.Module`, we could distinguish between true API functions and these other functions.

My clients have been using this check in various small projects for some 6 months, so the thing is tested in production. However, I made some changes today to make the check more friendly to the general population, so it's possible that I introduced some bugs.


sasa1977:
> Would it be possible to include a wildcard placeholder that basically means "any part we did not list explicitly"?

After sleeping it over, I like this idea. I propose we use `:_` b/c this is already used with ETS matchspecs.

rrrene:
> I propose we use `:_` b/c this is already used with ETS matchspecs.

Nice! Agreed. :+1:

sasa1977:
After implementing wildcards, I'm starting to think that this interface has its shortcomings. Consider the following config:

```elixir
[:_, :public_fun, :_, :private_fun, :_]
```

So here we're saying public funs must appear before private ones. Anything else can be placed before, between, or after. However, this pattern will break on the following code:

```elixir
def foo
@bar 1
def baz
```

You see, `@bar 1` moves us to the wildcard region after `def`, and now `def` is not allowed anymore. I could place some if/else to handle this, but then we'll end up with another problem in the default config:

```elixir
import foo
def bar
import baz
```

Basically, if we allow wildcards to appear in between the elements of the same type (like e.g. module attributes between functions of the same type), then it will sometimes work as expected, and sometimes won't.

I toyed with a few alternative ideas about the interface, but I'm currently not convinced about any of those. It seems that we need a richer language which would allow us to specify which elements can appear in a group like e.g. 

```elixir
[
  :_, 
  {:public_fun, :module_attribute},  # these can appear intertwined
  :_, 
  {:private_fun, :module_attribute},  # these can appear intertwined
  :_
]
```

But I feel that this is becoming too complex too fast, and I propose we scrape this idea for some later version, and stick with the current implementation. It solves some things, though it definitely has room for improvements. But we can spend some time thinking about it, and maybe we'll come up with a solid idea, rather than some ad-hoc clumsy interface. 

WDYT?

WDYT?

rrrene:
> I feel that this is becoming too complex too fast

You're probably right. Let me sleep on it for a day or two and then use the strict layout check if we don't have a better idea. :+1:

rrrene:
@sasa1977 Thx for your work on this!

I concluded that you are right and this was moving too quickly. I renamed the check `StrictModuleLayout`, so that we can add a `FuzzyModuleLayout` in the future.

It's live in `v1.3.0-rc2`

:+1:

sasa1977:
Cool, thanks for letting me know!
I wonder if https://github.com/rrrene/credo-proposals/issues/22 is then resolved? With the defaults we provided in this check, we also verify the use/import/alias/require order.

PR #756, opened 2020-03-01
Add example to explanation - negated is nil check - Closes #748

### LINKED ISSUE #748: Add example to `negated_is_nil` check
### Environment

* Credo version (`mix credo -v`): master
* Erlang/Elixir version (`elixir -v`): n/a
* Operating system: n/a

After creating the `negated_is_nil` check a couple weeks ago, I realized there is a better example that shows even more intent than the current one. When one of my coworkers saw it, I gave the example and I thought it was better than the original one.

```
   # current
      We should avoid negating the `is_nil` predicate function.
      Here are a couple of examples:
      The code here ...
          def fun(%{external_id: external_id, id: id}) when not is_nil(external_id) do
             ...
          end
      ... can be refactored to look like this:
          def fun(%{external_id: nil, id: id}) do
            ...
          end
          def fun(%{external_id: external_id, id: id}) do
            ...
          end
      Similar to negating unless blocks, the reason for this check is not
      technical, but a human one. If we can use the positive, more direct and human
      friendly case, we should.
```

add an example: 
```
  # proposed
      We should avoid negating the `is_nil` predicate function.
      Here are a couple of examples:
      The code here ...
          def fun(%{external_id: external_id, id: id}) when not is_nil(external_id) do
             ...
          end

      ... can be refactored to look like this:
          def fun(%{external_id: nil, id: id}) do
            ...
          end
          def fun(%{external_id: external_id, id: id}) do
            ...
          end

      ... or even better, can match on what you were expecting on the first place:
          def fun(%{external_id: external_id, id: id}) when is_binary(external_id) do
            ...
          end
          def fun(%{external_id: nil, id: id}) do
            ...
          end
          def fun(%{external_id: external_id, id: id}) do
            ...
          end

      Similar to negating unless blocks, the reason for this check is not
      technical, but a human one. If we can use the positive, more direct and human
      friendly case, we should.
```

### Expected outcome
Add another example to the `negated_in_nil` check.

pdgonzalez872:
Closes #748

### DIFF
--- a/lib/credo/check/refactor/negated_is_nil.ex
+++ b/lib/credo/check/refactor/negated_is_nil.ex
@@ -4,21 +4,27 @@ defmodule Credo.Check.Refactor.NegatedIsNil do
     explanations: [
       check: """
       We should avoid negating the `is_nil` predicate function.
-
       Here are a couple of examples:
-
       The code here ...
-
           def fun(%{external_id: external_id, id: id}) when not is_nil(external_id) do
              ...
           end
 
       ... can be refactored to look like this:
-
           def fun(%{external_id: nil, id: id}) do
             ...
           end
+          def fun(%{external_id: external_id, id: id}) do
+            ...
+          end
 
+      ... or even better, can match on what you were expecting on the first place:
+          def fun(%{external_id: external_id, id: id}) when is_binary(external_id) do
+            ...
+          end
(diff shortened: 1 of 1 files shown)


rrrene:
@pdgonzalez872 Thx :+1:

PR #761, opened 2020-03-06
Fix mismatch in base_priority type

TheFirstAvenger:
Getting warnings on master about mismatched types. Updated types to reflect values in use.


rrrene:
@TheFirstAvenger Thx for your continued support :+1: It is greatly appreciated. 🎉

PR #763, opened 2020-03-13
Ignore module attributes in UnsafeToAtom

### LINKED ISSUE #749: Credo warns about creating atoms at runtime in compile time code
### Precheck

### Environment

* Erlang/Elixir version (`elixir -v`): Elixir 1.10.1 (compiled with Erlang/OTP 22)
* Operating system: archlinux

I load a JSON file into a module attribute at compile time, like this:
```
@lut Jason.decode!(File.read!("priv/data/prime_lut.json"), keys: :atoms)
```

### Expected outcome

Credo should understand that this code is only run once, at compile time, and not complain

### Actual outcome

Credo warns:
```
┃ [W] ↗ Prefer Jason.decode!(..., keys: :atoms!) over Jason.decode!(..., keys: 
┃       :atoms) to avoid creating atoms at runtime
```

TheFirstAvenger:
Module attributes are determined at compile time, so are safe to use without fear of overloading the runtime atom table. This PR ignores module attributes for the UnsafeToAtom check.

Fixes #749

### DIFF
--- a/lib/credo/check/warning/unsafe_to_atom.ex
+++ b/lib/credo/check/warning/unsafe_to_atom.ex
@@ -41,6 +41,10 @@ defmodule Credo.Check.Warning.UnsafeToAtom do
     Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
   end
 
+  defp traverse({:@, _, _}, issues, _) do
+    {nil, issues}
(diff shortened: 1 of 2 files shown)


rrrene:
@TheFirstAvenger Thx!

PR #769, opened 2020-04-06
Improve strict module layout

sasa1977:
This PR brings in a few minor changes/additions to the `StrictModuleLayout` check. These additions are motivated by the real usage in the project of my client (which is where the original implementation of this check was done).

Note that these changes weren't discussed in any issue so far. However, to better understand the issues, I experimented with a fix, and so I ended up with the solution here, and then I figured I may as well submit it :-) That said, I'm not insisting on these changes, and so if you have some other ideas feel free to propose them.

Anyway, here are the changes:

1. Callback macros and callback functions are bundled under the same key `callback_impl`. In other words, with this change we can't enforce that e.g. callback macros must preceed callback functions. The reason is because a single module could implement multiple behaviours, and then such order becomes counterproductive, because we typically want to group callbacks by the module (e.g. first all callbacks of `Foo`, then all callbacks of `Bar`).
2. Added the `:ignore` option which allows us to ignore some module parts (i.e. allow some parts to appear anywhere in the module). For my client's projects, the concrete example would be private macros.

### DIFF
--- a/lib/credo/check/readability/strict_module_layout.ex
+++ b/lib/credo/check/readability/strict_module_layout.ex
@@ -68,14 +68,36 @@ defmodule Credo.Check.Readability.StrictModuleLayout do
     ]
 
   alias Credo.Code
+  alias Credo.CLI.Output.UI
 
   @doc false
   def run(source_file, params \\ []) do
+    params = normalize_params(params)
+
     source_file
     |> Code.ast()
     |> Credo.Code.Module.analyze()
     |> all_errors(params, IssueMeta.for(source_file, params))
     |> Enum.sort_by(&{&1.line_no, &1.column})
+  end
+
+  defp normalize_params(params) do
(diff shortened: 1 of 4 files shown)


rrrene:
I am very much in favor of this.

How do we ensure some kind of backwards compatibility? I do understand that there can't 100% backwards compatibility in this case. But could do something that makes this change less "breaking"?

Would it make sense to include a warning for existing configs that include `:callback_fun`? Would it be okay to rewrite existing `:callback_fun` values to `:callback_impl` + print a warning about it?

sasa1977:
> Would it be okay to rewrite existing `:callback_fun` values to `:callback_impl` + print a warning about it?

Yeah, that's a good suggestion! I'll add it.

rrrene:
@sasa1977 :+1: Thank you!

PR #770, opened 2020-04-16
Add note to end of execution noting `mix credo explain`

### LINKED ISSUE #751: Unable to find configuration options for 'Function body is nested too deep'
### Environment
* Erlang/Elixir version (`elixir -v`): 1.9.1
* Operating system: Mac Os

Disable or change the rule that outputs "Function body is nested too deep".

### Expected outcome
Change max depth from 2 to 3 or disable this rule.

### Actual outcome
Not being able to do so.

I could not find anywhere how to change configuration for this specific case in .credo.exs.

Suggestion: maybe create a dictionary that links all messages with their respective NAME_OF_CONFIG and configuration options. If this already exists, I could not find it in the docs or issues.

TheFirstAvenger:
Adds note to end of execution noting the `mix credo explain` feature. Only appears when at least one issue is found. Fixes #751.

Screen Shot 2020-04-16 at 12 07 12 PM

### DIFF
--- a/lib/credo/cli/output/summary.ex
+++ b/lib/credo/cli/output/summary.ex
@@ -10,6 +10,7 @@ defmodule Credo.CLI.Output.Summary do
     {:readability, "code readability issue", "code readability issues"},
     {:design, "software design suggestion", "software design suggestions"}
   ]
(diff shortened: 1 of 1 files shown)


rrrene:
We need something like this. Not sure if we will rephrase the message, but this is definitely a win! :+1:

PR #771, opened 2020-04-20
Add explicit default to Refactor.VariableRebinding check's allow_bang option

stewart:
This was incorrectly surfaced as a warning in an app I work on after upgrading to `v1.4.0`:

    $ mix credo
    ** (config) Credo.Check.Refactor.VariableRebinding: unknown param `allow_bang`.


rrrene:
@stewart Thx! :+1:

PR #773, opened 2020-04-23
Check for `===` as well as `==` in empty enum check

jonnystoten:
At the moment, credo warns on 
```elixir
length(some_list) == 0
```

After this change, it will also warn on
```elixir
length(some_list) === 0
```


NickNeck:
That is true. Credo recommends using `Enum.empty?/1` or `list == []`.
https://github.com/rrrene/credo/blob/ea1d73607e9edcf96264a59563fc6592134298ac/lib/credo/check/warning/expensive_empty_enum_check.ex#L64

rrrene:
@jonnystoten Thx! :+1:

PR #774, opened 2020-05-08
Correctly deduplicate multiclauses

sasa1977:
In the previous version, module part analysis would treat multiclauses as distinct functions. That cause problems with the `StrictModuleLayout`.

For example:

```elixir
@impl
def handle_call(...), do:
def handle_call(...), do:
```

The second clause would be incorrectly interpreted as a public function.


rrrene:
@sasa1977 Thx :+1:

PR #775, opened 2020-05-08
Improve performance

sasa1977:
On a larger test file from a real world project (3600 LOC) this reduces the credo check time from what seems like an infinity (I didn't have the patience to let it finish) to 4.7 seconds.

I feel there's some more room for improvements, but this is already a solid gain, while the change is simple enough, so I'm submitting it now. 

Time-permitting I might analyze perf further, but I can't commit. I advise opening up an ongoing issue for this, and establishing some benchmarks, especially for situations when the input file is large, and when there are a lot of issues.

My impression (based on a very shallow walk through a small part of the code) is that credo might not handle such scenarios quite well. I've see some `++` appends performed during traversals, and random-access writes performed on lists, which might significantly affect performance with larger inputs.

### DIFF
--- a/lib/credo/priority.ex
+++ b/lib/credo/priority.ex
@@ -71,6 +71,7 @@ defmodule Credo.Priority do
 
   defp make_base_map(priority_list, %SourceFile{} = source_file) do
     ast = SourceFile.ast(source_file)
+    scope_info_list = Scope.scope_info_list(ast)
 
     priority_list
     |> Enum.with_index()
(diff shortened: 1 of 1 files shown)


rrrene:
My goal for the next version of Credo was to provide speed improvements, so this is a perfect match.

I changed the way the runner executes between `v1.4.0` and `master` leading to a significant speed increase for codebases of all sizes. Here's an example screenshot I took last week (running against Elixir's source):

Credo_1_4_vs_1_5_preview

Your PR is more than welcome and any further analysis or insights into how we could measure/monitor performance in the future are greatly appreciated! 🎉 :+1:

PR #778, opened 2020-05-31
use created expanded filename path

starbelly:
- private fn get_config_file/2 in Credo.ConfigBuilder created the
   var `filename` for the expanded path of `config_file` but was not
   used and resulted in a warning.


rrrene:
Ha! Good catch! :+1:

PR #779, opened 2020-05-31
Block pipe check

starbelly:
Closes https://github.com/rrrene/credo-proposals/issues/37

There is still a good bit of TODO, in particularly around documentation and more specifically terminology, which I felt would be best served by discussion in a draft.

### DIFF
--- a/lib/credo/check/readability/expr_pipe.ex
+++ b/lib/credo/check/readability/expr_pipe.ex
@@ -6,7 +6,43 @@ defmodule Credo.Check.Readability.ExprPipe do
       check: """
       Pipes (`|>`) should not be used with case, if, or unless expressions.
 
-      # TODO: Better docs here.
+      The code in this example ...
+
+          list
+          |> Enum.take(5)
+          |> Enum.sort()
+          |> case do 
+            [[_h|_t]|_] -> true
+            _ -> false
+          end
+
+      ... should be refactored to look like this:
+
+          maybe_nested_lists = list
+                               |> Enum.take(5)
+                               |> Enum.sort()
+
+          case maybe_nested_lists do 
+            [[_h|_t]|_] = true
+            _->  false
+          end
+
+
+      ... or create a new function
+
+          list
+          |> Enum.take(5)
+          |> Enum.sort()
+          |> contains_nested_list?()
+
+
+      Piping to expressions is harder to read because it may obscure intention, increase cognitive load on the 
+      reader, and suprising to the reader per not following basic syntax principles set forth by all other 
+      expressions. Instead, prefer introducing variables to your code or new functions when it may be a sign that 
+      your function is getting too big and/or has too many concerns. 
+
+      Like all `Readability` issues, this one is not a technical concern, but you can improve the odds of others reading
+      and understanding the intent of your code by making it easier to follow. 
       """
     ]
 
--- a/lib/credo/check/readability/expr_pipe.ex
+++ b/lib/credo/check/readability/expr_pipe.ex
@@ -1,0 +1,44 @@
+defmodule Credo.Check.Readability.ExprPipe do
+  use Credo.Check,
+    base_priority: :high,
+    tags: [:controversial],
+    explanations: [
+      check: """
+      Pipes (`|>`) should not be used with case, if, or unless expressions.
+
+      # TODO: Better docs here.
+      """
+    ]
+
+  @doc false
+  @impl true
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    {_continue, issues} =
+      Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), {true, []})
+
+    issues
+  end
+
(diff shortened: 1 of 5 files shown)


starbelly:
Expression is probably too generic here, suggestions?

lib/credo/check/readability/expr_pipe.ex
```diff
@@ -0,0 +1,44 @@
...
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    {_continue, issues} =
+      Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), {true, []})
+
+    issues
+  end
+
+  defp traverse({:|>, meta, [_, {expr, _, _}]} = ast, {true, issues}, issue_meta)
```
starbelly, line 24: Could break this into a multi-clause functions, not sure it's worth it 🤔

lib/credo/check/readability/expr_pipe.ex
```diff
@@ -0,0 +1,80 @@
...
+      ... or create a new function
+
+          list
+          |> Enum.take(5)
+          |> Enum.sort()
+          |> contains_nested_list?()
+
+
+      Piping to expressions is harder to read because it may obscure intention, increase cognitive load on the 
+      reader, and suprising to the reader per not following basic syntax principles set forth by all other
```
Adzz, line 40: fwiw I don't agree with this at all. `case` and `if` are just functions, so piping into them follows the convention in the rest of the language. And piping into a case statement is not an indication that your function is to big as we can trivially see with this example:

```elixir
thing(stuff)
|> case do
   {:ok, stuff} -> stuff
   {:error, error} -> error
end
```

starbelly: A few points: 

1. You definitely don't have to agree with it. I'd be willing to bet you don't agree with everything in credo as is, I know I don't. These things are all highly subjective. This one has been marked as controversial and not enabled by default to boot.  It's essential that we all agree to disagree. 

2. As to whether we consider `case`, `if`, or `unless` functions... I do not. At best and IMO these are macros. It all decompiles to erlang and in erlang these are keywords and definitely not macros or functions, but even this is arguable depending on the level of abstraction the developer chooses to operate at. However and IMO Elixir is Erlang.  Once again, it's essential that we all agree to disagree. 

3. When I say "big" I am definitely not referring to SLOC.  I'm referring to how much knowledge a function has and further it can be said that every time a conditional expression is introduced in a function it's a branching point and may be (not definitely) time to create a new function that contains said conditional. The wording here definitely needs improving and it's all open to suggestion. But yet again, it's essential that we all agree to disagree.

rrrene:
I am in favor of adding this as a controversial check (just like you proposed). Can we brainstorm on the name once more?  `ExprPipe` seems to tense (imho).

starbelly:
@rrrene Yeah, I definitely don't like `expr` here, was too vague, possibly misnomer... what about `Conditional` ?

rrrene:
Maybe this is more about piping into a function/macro that has a block, rather than about "conditionals" or "expressions".

Opinions?

starbelly:
@rrrene Yeah, so this is much more specific I believe. That is to say, the code didn't try to catch other macros, just a very specific set, that being conditionals (if, unless, case). I could see it being adjusted to prevent piping to a bunch of Kernel functions/macros, but that was never my end goal.

rrrene:
What I'm trying to say is: From a user's perspective, once you want to use some thing that "prevents" this

```elixir
stuff
|> thing()
|> case do
   {:ok, stuff} -> stuff
   {:error, error} -> error
end
```

wouldn't you also want to prevent this?

```elixir
stuff
|> thing()
|> my_macro do
   {:ok, stuff} -> stuff
   {:error, error} -> error
end
```

because the feeling/reasoning that triggered the creation of this check was not about any specific macro (`if`, `with`, `case`) but rather about piping "into" a block.

I'm not sure, but I think this could be the real benefit of a check like this.

starbelly:
@rrrene in that sense, yes I fully agree, as I would prefer not to pipe to a `try` block either. So, perhaps there is more work to be done on in this check, specifically catching anything that is `keyword do`, and the check could be renamed to PipeBlockCheck. Thoughts?

rrrene:
@starbelly Sounds perfect :+1:

starbelly:
@rrrene 👍 Changed the code, module name, added a test for `try do`, need to update the docs a bit... coming 🔜

starbelly:
@rrrene Docs updated, really trimmed them, as it's folly to try and provide a technical rationale here. Unless you or someone has suggestions IMO this is good to go 👌

OvermindDL1:
A `receive` block would be another.

But yeah, this should definitely not be enabled by default.  I pass in to `case` substantially for such ok/error destructuring.

starbelly:
@OvermindDL1 Indeed.

rrrene:
@starbelly Thx, great work! 👍  I added an `:exclude` param so that users like @OvermindDL1 can configure their personal exceptions like this

```elixir
{Credo.Check.Readability.BlockPipe, exclude: [:case]}
```

starbelly:
@rrrene Thanks for adding that, oversight on my end.

PR #780, opened 2020-05-31
fix Elixir formatter asterisk spacing conflict

msz:
The Elixir formatter always removes spaces around the asterisk in
typespecs for binaries by default. Credo shouldn't conflict with the
default Elixir formatter settings, but it raises an issue if the rest of
the file is using spaces around operators.

This PR adds handling of this corner case.


rrrene:
@msz 👍 Thx!

PR #782, opened 2020-06-16
Add check for unsafe Application calls in module attributes 

akoutmos:
This check looks for following calls in module attributes:
- `Application.fetch_env/2`
- `Application.fetch_env!/2`
- `Application.get_all_env/1`
- `Application.get_env/3`
- `Application.get_env/2`

With the introduction of `Application.compile_env` in Elixir 1.10, developers should be leveraging that instead.

I personally think this should be in core Credo versus a 3rd party plugin as this is something I see happen almost weekly by people new to Elixir and even some seasoned Elixir developers.

As a side note, thanks a bunch for Credo! Been using it for years and appreciate all the hard work you have put into it :).

### DIFF
--- a/lib/credo/check/warning/application_config_attribute.ex
+++ b/lib/credo/check/warning/application_config_attribute.ex
@@ -1,0 +1,103 @@
+defmodule Credo.Check.Warning.ApplicationConfigAttribute do
+  use Credo.Check,
+    base_priority: :high,
+    tags: [:controversial],
+    category: :warning,
+    explanations: [
+      check: """
+      Module attributes are evaluated at compile time and not at run time. As
+      a result, certain configuration read calls made in your module attributes
+      may work as expected during local development, but may break once in a
+      deployed context.
+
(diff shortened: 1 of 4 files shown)


rrrene:
Nice work. Could we rename this `ApplicationConfigInModuleAttribute`?

akoutmos:
It has been done! Thanks for taking the time to look at the PR. Appreciate it :)

rrrene:
:+1:

akoutmos:
I just noticed that the default credo config that is generated `mix credo.gen.config` is based on the `.credo.exs` file in Credo itself https://github.com/rrrene/credo/blob/master/lib/credo/cli/command/gen.config.ex#L6.

Should I have added this new rule as an entry in that file or is that something that you do when you cut a new release?

rrrene:
@akoutmos Good catch! I added it. :+1:

akoutmos:
Awesome! Thanks!!

PR #784, opened 2020-06-20
Add Emacs as a supported editor

Sasanidas:
Emacs has support for credo with the flycheck plugin
https://www.flycheck.org/en/latest/languages.html#elixir


rrrene:
@Sasanidas Thx!

PR #785, opened 2020-06-25
add recommended Elixir and Erlang/OTP combinations to CI

adrianomitre:
### Main course

Make sure all recommended Elixir and Erlang/OTP combinations are run on CI

Quoting @josevalim:
> I would still advise folks to go with:
> * For each Elixir version, add a run with earliest supported Erlang/OTP
> * For the last Elixir version, also test the latest supported Erlang/OTP

Please refer to the original discussion on https://github.com/dashbitco/broadway/pull/188 for the full rationale.

### Side dish

This pull request also
- updates [actions/setup-elixir](https://github.com/actions/setup-elixir) and [actions/checkout](https://github.com/actions/checkout) to their latest versions
- restricts compilation with `--warnings-as-errors` to the latest Elixir version, on a separate CI job, following [suggestion](https://github.com/whatyouhide/stream_data/pull/142#discussion_r444396899) by @whatyouhide 
- add CI job to check formatting (`mix format --check-formatted`) on the latest Elixir version

PR #791, opened 2020-07-17
add Erlang/OTP 23 (latest) to CI matrix

adrianomitre:
Add last Erlang/OTP version to CI matrix and also upgrade minor Erlang/OTP 22 version and Elixir 1.7 patch version.

Warning: there are some combinations (none added by this pull request) that are not compatible according to [Compatibility and Deprecations — Elixir v1.10.3](https://hexdocs.pm/elixir/compatibility-and-deprecations.html) (e.g., Elixir 1.9 & Erlang/OTP 23), but @rrrene [disapproved](https://github.com/rrrene/credo/pull/785#issuecomment-659462251) the [non-cartesian-product approach](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations) to Elixir × Erlang/OTP version matrix, so I am waiting for him to suggest a workaround more to his liking. Maybe using the cartesian approach + [exclusion](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix) for filtering?

This pull request was extracted (deformed) from https://github.com/rrrene/credo/pull/785.

PR #793, opened 2020-07-17
disable fail-fast strategy to Erlang/OTP x Elixir version matrix

adrianomitre:
The change in this pull request, which was extracted from https://github.com/rrrene/credo/pull/785, prevents that all in-progress jobs are cancelled if any matrix job fails. The rationale is to save developers time, at the potential expense of computation power, by maximizing the amount of information extracted at once.

Particularly useful for a project that, such as this, often have a broken master:
image

References:
- [Workflow syntax for GitHub Actions - GitHub Docs - jobs.<job_id>.strategy.fail-fast](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)
- [How to avoid broken master with Pipelines for Merged Results and Merge Trains | GitLab](https://about.gitlab.com/releases/2019/09/11/how-to-avoid-broken-master-with-pipelines-for-merge-requests/)

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     name: "[${{matrix.otp}}/${{matrix.elixir}}] CI Tests on Credo [OTP/Elixir]"
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


rrrene:
Thx, please refer to #791 :+1:

PR #798, opened 2020-07-31
Add name of check to message when printing issues

TheFirstAvenger:
For a number of the checks, it is not clear which specific check you are dealing with when reading through the output in Credo. I have a few times had to copy the error message, then search for it in credo source to figure out which one it is. This PR simply notes the name of the check in parenthesis after the message.

Example:
Screen Shot 2020-07-31 at 12 47 37 PM

### DIFF
--- a/lib/credo/cli/command/suggest/output/default.ex
+++ b/lib/credo/cli/command/suggest/output/default.ex
@@ -173,7 +173,7 @@ defmodule Credo.CLI.Command.Suggest.Output.Default do
            priority: priority
          } = issue,
          source_file,
-         %Execution{format: _} = exec,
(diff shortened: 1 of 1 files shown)


rrrene:
Credo's default output is well curated to provide newcomers with just the right amount of information to not overwhelm them, but make them engage with Credo's reasoning and approach. So I am very hesitant to change the default output.

I think we could add this if `--verbose` is given. WDYT?

TheFirstAvenger:
Sounds good, I'll update the PR to only add it on `--verbose`

TheFirstAvenger:
@rrrene updated, it now appears only on `--verbose`

rrrene:
@TheFirstAvenger :+1:

PR #801, opened 2020-08-06
Add SkipTestWithoutComment check to warn if no comment precedes `@tag :skip` in a test

TheFirstAvenger:
"temporarily" skipped tests can become permanent fixtures in a codebase because the reason for the skip can be forgotten. Ideally skips should not be long-lived, so if you do have to skip a test, you should document why it is being skipped so the skip can be cleared as soon as possible. This check warns if there is not a comment on the line preceding the skip.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -83,6 +83,7 @@
         #
         {Credo.Check.Design.AliasUsage,
          [priority: :low, if_nested_deeper_than: 2, if_called_more_often_than: 0]},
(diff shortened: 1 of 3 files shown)


rrrene:
This sounds like a great idea :+1:

rrrene:
@TheFirstAvenger Thx! I toyed around with making this more forgiving where the comment *could* be put, but think your initial design was right all along. 👏

PR #802, opened 2020-08-11
Add check for @impl true to change to @impl MyBehaviour

TheFirstAvenger:
Add check for @impl true to change to @impl MyBehaviour

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -95,6 +95,7 @@
         #
         {Credo.Check.Readability.AliasOrder, []},
         {Credo.Check.Readability.FunctionNames, []},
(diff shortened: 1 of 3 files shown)


rrrene:
Also an interesting idea. Is this in any way "controversial" in that people tend to favor one way over the other or is there a broader consent on the "right way" to use `@impl` today?

TheFirstAvenger:
I have not heard pushback on going with the behaviour name instead of true, more often than not I hear "Oh, I didn't know you could do that", hence my thoughts that it could be a good thing for Credo to teach people about.

PR #807, opened 2020-09-24
Add check for capitalized start of the code comments

kanmaniselvan:
A controversial check which enforces comments longer than a word needs to be capitalized.

This works for both single-line and multi-line comments.

**A single-line comment should:**
```
      - start with ` or  '  or " or 0-9 or A-Z
      - end with . or ? or ! 
```

 Examples for valid case:
```elixir
      # This is a valid comment.

      # `code` block!

      # "wow", this is also valid.

      # 'also', this is valid?

      # 9 is also valid.

      # * starting with other symbol is valid.

      # - this is also valid.

      # valid_single_word_comment

      ############################
      ## This is a valid comment #
      ############################
```

**A multi-line comment should:**
```
      - Comment should start with ` or ' or " or 0-9 or A-Z
      - The following comment sentence can start with non-capitalized letters
      - Comment sentence ending with . or ? or ! at the end is considered as the end of the multi-line comments.
```
Examples for valid cases:
```
      # This is a valid comment
      # so, no issues.

      # `code` block!
      # New comment follows.

      # "yes", this is also valid:
      # def code(x) do
      #   IO.inspect x
      # end

      # 'also', this is valid?
      # - yes
      # - and yes

      # 9 is also valid,
      # but you need to be more
      # clear what 9 means.
```

I'm using this check with quite a big codebase; it seems stable and no issues. There could be some missed cases which I might not have seen, so far, it does well.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -163,6 +163,7 @@
         {Credo.Check.Design.DuplicatedCode, false},
         {Credo.Check.Readability.AliasAs, false},
         {Credo.Check.Readability.BlockPipe, false},
+        {Credo.Check.Readability.CapitalizedComments, false},
         {Credo.Check.Readability.ImplTrue, false},
         {Credo.Check.Readability.MultiAlias, false},
         {Credo.Check.Readability.SinglePipe, false},
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -31,7 +31,7 @@ defmodule Credo.Check do
       ```
 
   - `:param_defaults` Sets the default values for the check's params (e.g. `[param1: 42, param2: "offline"]`)
-  - `:run_on_all`     Sets whether the check runs on all source files at once or each source file separatly.
+  - `:run_on_all`     Sets whether the check runs on all source files at once or each source file separately.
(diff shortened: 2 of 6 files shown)


rrrene:
I think there might be value in this for some, notably yourself, otherwise you would not have put this together. Thank you for that! :clap:

On the one hand, I think a generalized check that let's users define arbitrary regexes that can enforce idiosyncratic comment formatting rules would be beneficial. I tried to [prototype something like that cased on your PR](https://github.com/rrrene/credo/tree/kanmaniselvan-capitalized_code_comments), but did not get a good feel for it yet.

On the other hand, I've been torn for some time about this, as your PR clearly shows how complex the problem can become. The check's is 30 lines (good!) which do not lay out an argument in favor of the enforced comment rule, but rather it takes 30 lines to describe said rule.

> A controversial check which enforces comments [...]

The reason I am not merging this is two-fold: Enforcing correct punctuation in sentences where appropriate seems like a very big problem , prone to false positives and negatives. And on a meta-level, I think that this seems to be the "wrong" kind of "controversial".

Thanks again for putting this together! 👍

kanmaniselvan:
All good! Thank you for having a look 👍

Issue #777, opened 2020-05-25
--min-priority not working (as I would expect)


JMurphyWeb:
### Environment

* Credo version: 1.4.0
* Erlang/Elixir version: 1.10.1
* Operating system: mac os

### What were you trying to do?
I am looking to add credo to my existing projects CI. 
Running `mix credo --strict` lists too many issues, so I want to initially only configure it to run on the 'higher' priority issues (and return a non-zero exit code if it finds a 'higher' priority issue)

I am assuming I should use the `--min-priority` flag: `mix credo suggest --min-priority higher`

### Expected outcome
I would expect this to only print checks where priority is 'higher'.

### Actual outcome
![Screenshot 2020-05-25 at 18 09 04](https://user-images.githubusercontent.com/12638261/82829800-6db37200-9eb4-11ea-82c1-407173d21783.png)
![Screenshot 2020-05-25 at 18 19 52](https://user-images.githubusercontent.com/12638261/82829809-72782600-9eb4-11ea-9bac-f2b558d856e7.png)

_______
I've also not found the documentation on `--min-priority` too clear (maybe I've missed something)
___
Am I doing something wrong?

rrrene:
Sounds like a bug, thx for reporting :+1:

PR #808, opened 2020-09-25
add min_priority after strict to avoid overwrites

grrrisu:
this commit changes the order of the `add_switch_` functions for `strict` and `min_priority`. 
As `add_switch_strict` overwrites in `Execution.set_strict`the min_priority value, it needs to be before the `add_switch_min_priorty`.
See https://github.com/rrrene/credo/blob/master/lib/credo/execution.ex#L298

the last time the order has been changed was with this commit:
https://github.com/rrrene/credo/commit/9d7b62e90b85fc07820ed0c81c371d9c48b0f1c1

Fixes https://github.com/rrrene/credo/issues/777


rrrene:
@grrrisu Thx!

PR #811, opened 2020-10-02
Fix `Strings.replace_with_spaces` with charlist

### LINKED ISSUE #809: "Space missing after comma" in String.split
### Environment

* Credo version (`mix credo -v`): 1.5.0-rc.2
* Erlang/Elixir version (`elixir -v`): 1.10.4
* Operating system: Ubuntu 18.04

I've run into another "space missing after comma" warning in this code block:
```elixir
  def process_value(name, value) do
    case name do
      'To' -> String.split(value, ",")   # <- here
      _ -> value
    end
  end
```
Perhaps exclude `","` from the regex, as that seems to be a common pattern in split/join operations? Cheers!

guitsaru:
There was a bug where a charlist before a string would cause the `replace_with_spaces` function to skip the rest of the line.

fixes #809

### DIFF
--- a/lib/credo/code/strings.ex
+++ b/lib/credo/code/strings.ex
@@ -145,7 +145,7 @@ defmodule Credo.Code.Strings do
   defp parse_charlist(str, acc, replacement) when is_binary(str) do
     {h, t} = String.next_codepoint(str)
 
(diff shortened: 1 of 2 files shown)


rrrene:
Thx @guitsaru! :+1:

PR #812, opened 2020-10-04
adds in missing < operator from this check

ityonemo:
`<` was missing from the non-special-form-operators list.


rrrene:
Thx @ityonemo! :+1:

PR #821, opened 2020-11-02
Add check for *_test.ex files

egze:
A check that enforces all test files in `test` folder to end with `_test.exs`.

This is something that is often done by mistake: a developer creates a test file but names it with `.ex` at the end. Then the file will be skipped by ExUnit and broken tests may get unnoticed.

### DIFF
--- a/lib/credo/check/warning/test_exs.ex
+++ b/lib/credo/check/warning/test_exs.ex
@@ -1,0 +1,67 @@
+defmodule Credo.Check.Warning.TestExs do
+  use Credo.Check,
+    base_priority: :high,
+    param_defaults: [excluded_paths: []],
+    explanations: [
+      check: """
+      Invoking mix test from the command line will run the tests in each file
+      matching the pattern `*_test.exs` found in the test directory of your project.
+
+      (from the `ex_unit` docs)
+
+      This check ensures that test files are not ending with `.ex` (which would cause them to be skipped).
+
+      """,
+      params: [
+        excluded_paths: "List of paths or regex to exclude from this check"
+      ]
+    ]
+
+  @test_files_with_ex_ending_regex ~r/test\/.*\/.*_test.ex$/
+
+  alias Credo.SourceFile
+
+  @doc false
+  def run(%SourceFile{filename: filename} = source_file, params \\ []) do
+    excluded_paths = Params.get(params, :excluded_paths, __MODULE__)
+    ignored = ignore_path?(filename, excluded_paths)
+
+    process_source_file(source_file, params, ignored)
+  end
+
+  defp process_source_file(_source_file, _params, true = _ignored), do: []
+
+  defp process_source_file(
+         %SourceFile{filename: filename} = source_file,
+         params,
+         false = _ignored
+       ) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    cond do
+      matches?(filename, @test_files_with_ex_ending_regex) ->
+        issue_for(issue_meta)
+        |> List.wrap()
+
+      :else ->
+        []
+    end
+  end
+
+  defp issue_for(issue_meta) do
+    format_issue(
+      issue_meta,
+      message: "Test files should end with .exs"
+    )
+  end
+
+  # Check if analyzed module path is within ignored paths
+  defp ignore_path?(filename, excluded_paths) do
+    directory = Path.dirname(filename)
+
+    Enum.any?(excluded_paths, &matches?(directory, &1))
+  end
+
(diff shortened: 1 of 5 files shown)


egze:
@rrrene What about this failing `Reproducing Test-Case Detector (experimental)`? I can ignore it, right?

lib/credo/check/warning/test_exs.ex
```diff
@@ -0,0 +1,67 @@
...
+      ]
+    ]
+
+  @test_files_with_ex_ending_regex ~r/test\/.*\/.*_test.ex$/
+
+  alias Credo.SourceFile
+
+  @doc false
+  def run(%SourceFile{filename: filename} = source_file, params \\ []) do
+    excluded_paths = Params.get(params, :excluded_paths, __MODULE__)
```
rrrene, line 26: You can remove this functionality as it is provided by Credo since the latest update: https://hexdocs.pm/credo/changelog.html#new-general-check-param-files

lib/credo/check/warning/test_exs.ex
```diff
@@ -0,0 +1,67 @@
...
+    param_defaults: [excluded_paths: []],
+    explanations: [
+      check: """
+      Invoking mix test from the command line will run the tests in each file
+      matching the pattern `*_test.exs` found in the test directory of your project.
+
+      (from the `ex_unit` docs)
+
+      This check ensures that test files are not ending with `.ex` (which would cause them to be skipped).
+
```
rrrene, line 13:
```suggestion
```

lib/credo/check/warning/test_exs.ex
```diff
@@ -0,0 +1,67 @@
...
+         false = _ignored
+       ) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    cond do
+      matches?(filename, @test_files_with_ex_ending_regex) ->
+        issue_for(issue_meta)
+        |> List.wrap()
+
+      :else ->
```
rrrene, line 46: We're using `true` as the catch all condition in the rest of the codebase.

```suggestion
      true ->
```

lib/credo/check/warning/test_exs.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Credo.Check.Warning.TestExs do
+  use Credo.Check,
+    base_priority: :high,
+    param_defaults: [excluded_paths: []],
```
rrrene, line 4: As mentioned in a sibling comment, Credo 1.5 allows you to specify which files to run your check on:

```suggestion
    param_defaults: [included: ["test/**/*_test.ex"]],
```

lib/credo/check/warning/test_exs.ex
```diff
@@ -0,0 +1,67 @@
+defmodule Credo.Check.Warning.TestExs do
```
rrrene, line 1: We find a better name. This is not warning on test files ending in exs.

Maybe `WrongTestFileExtension`?

rrrene (commented):
This is already looking pretty good.

Coincidentally, Credo just gained the ability to filter the enabled files per check, which means that we can remove most of the file-detection logic from this check.

lib/credo/check/warning/test_exs.ex
egze: removed

lib/credo/check/warning/test_exs.ex
egze: I like it. Done

lib/credo/check/warning/test_exs.ex
egze: This parameter seems to only add additional includes. All files from the project are still sent to the check. This is intentional, right? So I still need to check the source_file against my regex?

egze:
@rrrene After the changes, the CI is failing and I din't know why 😞 Works locally when I run `mix test`.

Can you have a look please?

egze:
@rrrene All green now, besides the `Reproducing Test-Case Detector (experimental)`

rrrene:
@egze This looks great. I'll finish work von Credo `1.5.1` and then merge this. Thx for the PR! :+1:

egze:
Hi @rrrene

Sorry to nag you, but is it possible to merge it? 😃

rrrene:
@egze Will merge as soon as I begin work on Credo 1.6 since this will become default-enabled check ([as per your suggestion](https://github.com/rrrene/credo/pull/821/files#diff-c44b474280f99dde56e6b80c90554ced12592de34aaa8d42186984126d5521d9R144)) which are only introduced in minor versions.

Stay tuned :+1:

PR #825, opened 2020-11-18
Fix ExpensiveEmptyEnumCheck when length is a var

### LINKED ISSUE #824: Incorrectly reports a variable named `length` as an expensive use of `length()`
The closest I've found to this bug is #720.

### Environment

* Credo version (`mix credo -v`): 1.5.1
* Erlang/Elixir version (`elixir -v`): Elixir 1.11.2 (compiled with Erlang/OTP 23)
* Operating system: Arch Linux

Consider this toy module:
```Elixir
defmodule Foo do
  def credo_you_are_wrong do
    length = 0
    if length == 0 do
      IO.puts "credo pls"
    end
  end
end
```
`length` is a variable here, not a list / not being used as a function, and is being compared to `0` in a way that makes sense to do so. credo reports this as an expensive use of `length()` instead.

### Expected outcome

credo understands that this is a variable and not a function call.

### Actual outcome

```
  Warnings - please take a look
┃ 
┃ [W] ↗ length == 0 is expensive. Prefer Enum.empty?/1 or list == []
┃       lib/foo.ex:4:8 #(Foo.credo_you_are_wrong)
```

w-sanches:
Fix #824

In cases where a local variable called `length` was checked for zero value, credo incorrectly assumed this to be a call to Kernel.length/1 and showed a warning that this was an expensive operation.
This is fixed by making sure pattern match checks length's AST for a single parameter at the last item in the tuple, meeting Kernel.length/1's AST.

This also removes warnings from length function calls with more than a single parameter.
I'm not sure this is desired, but since a length/2+ function can be anything other than checking the size of an enumerable and the suggestion might not be valid, I think it is only fair we don't warn people creating and using a function like that.
Please LMK if you think I should add a test for this scenario or if we should keep warning for length/2+ as it feels it was in some sort of unchecked limbo.

### DIFF
--- a/lib/credo/check/warning/expensive_empty_enum_check.ex
+++ b/lib/credo/check/warning/expensive_empty_enum_check.ex
@@ -31,7 +31,7 @@ defmodule Credo.Check.Warning.ExpensiveEmptyEnumCheck do
                               _,
                               _
                             }
-  @length_pattern quote do: {:length, _, _}
+  @length_pattern quote do: {:length, _, [_]}
   @comparisons [
     {@enum_count_pattern, 0},
     {0, @enum_count_pattern},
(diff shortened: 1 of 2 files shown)


rrrene:
@w-sanches Thx! :+1:

PR #830, opened 2020-12-02
Add check for specs containing structs to cut down on compile-time dependencies

andersonmcook:
I was recently working on a project that would recompile ~70 files each time it was recompiled.  I did some digging and figured out that it was a result of using structs eg `%SomeStruct{}` in specs. Switching them out to `SomeStruct.t()` (and making that type) cut down on the modules that would be recompiled each time. Structs create compile-time dependencies and this will warn about structs being used in specs.

I'm happy to turn this off by default or tag it as experimental :)

Thanks for the wonderful library!

### DIFF
--- a/lib/credo/check/warning/spec_with_struct.ex
+++ b/lib/credo/check/warning/spec_with_struct.ex
@@ -19,6 +19,8 @@ defmodule Credo.Check.Warning.SpecWithStruct do
       """
     ]
 
+  alias Credo.Code.Name
+
   @doc false
   @impl true
   def run(%SourceFile{} = source_file, params) do
@@ -26,21 +28,24 @@ defmodule Credo.Check.Warning.SpecWithStruct do
     Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
   end
 
-  # `::` and `when` are used in specs
-  defp traverse({:spec, meta, [{atom, _, _} | _] = args}, issues, issue_meta)
-       when atom in ~w(:: when)a do
-    case Macro.prewalk(args, nil, &find_structs/2) do
-      {ast, nil} ->
+  defp traverse({:@, meta, [{:spec, _, args}]}, issues, issue_meta) do
+    case Macro.prewalk(args, [], &find_structs/2) do
+      {ast, []} ->
         {ast, issues}
 
-      {ast, struct} ->
-        options = [
-          message: "Struct %#{struct}{} found in @spec",
-          trigger: struct,
-          line_no: meta[:line]
-        ]
+      {ast, structs} ->
+        issues =
+          Enum.reduce(structs, issues, fn curr, acc ->
+            options = [
+              message: "Struct %#{curr}{} found in @spec",
+              trigger: "%#{curr}{}",
+              line_no: meta[:line]
+            ]
 
-        {ast, [format_issue(issue_meta, options) | issues]}
+            [format_issue(issue_meta, options) | acc]
+          end)
+
+        {ast, issues}
     end
   end
 
@@ -48,8 +53,8 @@ defmodule Credo.Check.Warning.SpecWithStruct do
     {ast, issues}
   end
 
-  defp find_structs({:%, _, [{:__aliases__, _, modules} | _]} = ast, _acc) do
-    {ast, Enum.join(modules, ".")}
+  defp find_structs({:%, _, [{:__aliases__, _, _} = aliases | _]} = ast, acc) do
+    {ast, [Name.full(aliases) | acc]}
   end
 
   defp find_structs(ast, acc) do
--- a/lib/credo/check/warning/spec_with_struct.ex
+++ b/lib/credo/check/warning/spec_with_struct.ex
@@ -12,10 +12,10 @@ defmodule Credo.Check.Warning.SpecWithStruct do
       Example:
 
           # preferred
-          @spec a_function(%MyModule{}) :: any
+          @spec a_function(MyModule.t()) :: any
 
           # NOT preferred
-          @spec a_function(MyModule.t()) :: any
+          @spec a_function(%MyModule{}) :: any
       """
     ]
 
--- a/lib/credo/check/warning/spec_with_struct.ex
+++ b/lib/credo/check/warning/spec_with_struct.ex
@@ -9,10 +9,12 @@ defmodule Credo.Check.Warning.SpecWithStruct do
 
       It is preferable to define and use `MyModule.t()` instead of `%MyModule{}` in specs.
 
-      Avoid:
+      Example:
+
+          # preferred
           @spec a_function(%MyModule{}) :: any
 
-      Prefer:
+          # NOT preferred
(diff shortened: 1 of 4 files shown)


test/credo/check/warning/spec_with_struct_test.exs
```diff
@@ -0,0 +1,154 @@
+defmodule Credo.Check.Warning.SpecWithStructTest do
+  use Credo.Test.Case
+
+  @described_check Credo.Check.Warning.SpecWithStruct
+
+  @my_struct_module """
+  defmodule MyApp.MyStruct do
+    @type t :: %__MODULE__{id: integer, name: String.t()}
+    defstruct [:id, name]
```
rrrene, line 9: Shouldn't this be `:name`?

test/credo/check/warning/spec_with_struct_test.exs
```diff
@@ -0,0 +1,154 @@
...
+  end
+  """
+
+  @a_struct_module """
```
rrrene, line 13: Why do we need two structs for the tests?

lib/credo/check/warning/spec_with_struct.ex
```diff
@@ -0,0 +1,54 @@
...
+      Structs create compile-time dependencies between modules.  Using a struct in a spec
+      will cause the module to be recompiled whenever the struct's module changes.
+
+      It is preferable to define and use `MyModule.t()` instead of `%MyModule{}` in specs.
+
+      Avoid:
+          @spec a_function(%MyModule{}) :: any
+
+      Prefer:
+          @spec a_function(MyModule.t()) :: any
```
rrrene, line 16: Please use the style and wording found in the other checks:

https://github.com/rrrene/credo/blob/bd85e260d87bb177c1ee63cb61fba1561150347a/lib/credo/check/warning/raise_inside_rescue.ex#L9-L31

lib/credo/check/warning/spec_with_struct.ex
```diff
@@ -0,0 +1,54 @@
...
+    ]
+
+  @doc false
+  @impl true
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:spec, meta, args}, issues, issue_meta) do
```
rrrene, line 27: Doesn't this also get called for macros, functions and variables of the name "spec"?

rrrene (changes requested):
This seems like a really great addition for Credo.

A few requests:

test/credo/check/warning/spec_with_struct_test.exs
andersonmcook: Thanks!  Fixed in https://github.com/rrrene/credo/pull/830/commits/32c6e19d212695cdf71e3fa57e46bc2d93430362

test/credo/check/warning/spec_with_struct_test.exs
andersonmcook: I wanted to test that I could catch multiple issues for different structs and that I was formatting issues correctly.

lib/credo/check/warning/spec_with_struct.ex
andersonmcook: Thanks! Fixed in https://github.com/rrrene/credo/pull/830/commits/32c6e19d212695cdf71e3fa57e46bc2d93430362

lib/credo/check/warning/spec_with_struct.ex
andersonmcook: Great catch!  I made the check a little more specific by looking at `::` and `:when` to ignore functions/macros and variables named `spec` in https://github.com/rrrene/credo/pull/830/commits/32c6e19d212695cdf71e3fa57e46bc2d93430362

lib/credo/check/warning/spec_with_struct.ex
```diff
@@ -0,0 +1,58 @@
...
+  @doc false
+  @impl true
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  # `::` and `when` are used in specs
+  defp traverse({:spec, meta, [{atom, _, _} | _] = args}, issues, issue_meta)
+       when atom in ~w(:: when)a do
```
rrrene, line 31: What I meant was, that we should be looking for a module attribute here:

```elixir
  defp traverse({:@, _, [{:spec, meta, args}]}, issues, issue_meta) do
```

lib/credo/check/warning/spec_with_struct.ex
```diff
@@ -0,0 +1,58 @@
...
+        {ast, [format_issue(issue_meta, options) | issues]}
+    end
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp find_structs({:%, _, [{:__aliases__, _, modules} | _]} = ast, _acc) do
+    {ast, Enum.join(modules, ".")}
```
rrrene, line 52: Missed this in the first pass: This should use `Credo.Code.Name.full/1`.

test/credo/check/warning/spec_with_struct_test.exs
rrrene: > I wanted to test that I could catch multiple issues for different structs

That is the right goal. Unfortunately, the check does not catch multiple Structs if they are part of the same `@spec`.

test/credo/check/warning/spec_with_struct_test.exs
```diff
@@ -0,0 +1,188 @@
...
+  test "it should report multiple issues" do
+    [
+      """
+      defmodule Offender do
+        @spec f(my_struct :: %MyApp.MyStruct{}) :: any
+        def f(_) do
+          "oops"
+        end
+
+        @spec g(a_struct :: %AStruct{}) :: any
```
rrrene, line 118: The test suite is missing a case, where we test multiple structs in a single `@spec`:

```elixir
        @spec g(a_struct :: %AStruct{}, my_struct :: %MyApp.MyStruct{}) :: any
```

rrrene (changes requested):
Almost there. This will be a nice addition to Credo's list of standard checks :+1:

test/credo/check/warning/spec_with_struct_test.exs
andersonmcook: Thanks! I updated the code and test to cover this case.

test/credo/check/warning/spec_with_struct_test.exs
andersonmcook: Thanks!  I added a test on line 132 to cover this.

rrrene:
LGTM! :+1: Thx!

Issue #831, opened 2020-12-02
Exit with error status when fail to load config


Danielwsx64:
### Environment

* Credo version (`mix credo -v`): 1.5.1
* Erlang/Elixir version (`elixir -v`): Elixir 1.10.3 (compiled with Erlang/OTP 21)
* Operating system: Ubuntu 20.04.1 LTS

### What were you trying to do?
Run `mix credo` with a syntax error in `.credo.exs`


### Expected outcome
`Credo` exit with an error status.


### Actual outcome
```bash
➜  mix credo                                                                                                                                                                                                                            
** (config) Error while loading config file!

  file: /home/user/project/credo.exs
  line: 173

  syntax error before: '{'

  171          # Custom checks can be created using `mix credo.gen.check`.
  172          #
  173          {CredoNaming.Check.Consistency.ModuleFilename}
  174          ]
  175        }


➜  echo $?                                                                                                                                                                                                                              
0
```

### Context

A syntax error was added by the team, but it tooks some time to be noticed because CI pipeline was ok, since credo was not exiting with an error status.

rrrene:
Thanks for reporting this 😀  It is now fixed on `master` and will be part of Credo `1.6.0` (as we discussed in the linked PR).

You can try this by setting the Credo dep to

```elixir
{:credo, github: "rrrene/credo"}
```

Please report back if your issue is solved! 👍

Danielwsx64:
Sorry for the delay. I tested and it's working. 👍🏽 
![image](https://user-images.githubusercontent.com/17304947/120091127-80234580-c0de-11eb-8ddf-50eda4ca0d95.png)

rrrene:
This is part of v1.6.1 💚

PR #832, opened 2020-12-02
Exit with error status when load config fail

Danielwsx64:
As reported in #831 when `Credo` fails to load the config file the exit status is `0` (success). This behavior generates a false positive on CI pipelines.

This PR adds a status code `128` when config load fails.

Obs. I didn't find an appropriate test file to add a spec for this new behavior.

### DIFF
--- a/lib/credo/execution/task/convert_cli_options_to_config.ex
+++ b/lib/credo/execution/task/convert_cli_options_to_config.ex
@@ -15,6 +15,7 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
   def halt_on_error({:error, error}, exec) do
     exec
     |> Execution.put_assign("#{__MODULE__}.error", error)
+    |> Execution.put_assign("credo.exit_status", 128)
     |> Execution.halt()
   end
 
--- a/guides/introduction/exit_statuses.md
+++ b/guides/introduction/exit_statuses.md
@@ -17,11 +17,14 @@ This way you can reason about the encountered issues right from the exit status.
 
 Default values for the checks are based on their category:
 
-    consistency:  1
-    design:       2
-    readability:  4
-    refactor:     8
-    warning:     16
+    consistency:         1
+    design:              2
+    readability:         4
+    refactor:            8
+    warning:             16
+    <custom category>:   32
+    <custom category>:   64
+    runtime errors:      128 (and above)
 
 Let's see what this means using an example:
 


RafaelCamarda (approved):

fernandoalmeida (approved):

lib/credo/execution/task/convert_cli_options_to_config.ex
```diff
@@ -15,6 +15,7 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
   def halt_on_error({:error, error}, exec) do
     exec
     |> Execution.put_assign("#{__MODULE__}.error", error)
+    |> Execution.put_assign("credo.exit_status", 1)
```
fernandoalmeida, line 18: According to the exit statuses documentation, 1 means that consistency issues were found:

https://github.com/rrrene/credo/blob/master/guides/introduction/exit_statuses.md

How in this case we have a "runtime error" and the execution fails, what do you think about using `-1`?

fernandoalmeida (changes requested):

lib/credo/execution/task/convert_cli_options_to_config.ex
Danielwsx64: Good point @fernandoalmeida. I found this similar case where the code `1` is used too:

https://github.com/rrrene/credo/blob/7f224fedc6686b89b0b06cd8c481664eb8c03ccf/lib/credo/execution/task/validate_options.ex#L26

Maybe @rrrene could help us with this.

rrrene:
This is certainly a great oversight on Credo's part. Thx for reporting and fixing it!

@fernandoalmeida is right: we cannot use `1` (for historical reasons) and I'm pretty sure we can not use `-1` either (for technical reasons).

These are the current exit statuses ([the reasoning is outlined here](https://github.com/rrrene/credo/blob/7f224fedc6686b89b0b06cd8c481664eb8c03ccf/guides/introduction/exit_statuses.md)):

    consistency:        1
    design:             2
    readability:        4
    refactor:           8
    warning:            16

The documentation does not make this explicit, but it means that we implicitly have three more exit statuses, which fall under the bitwise-combining-scheme:

    <custom category>:  32
    <custom category>:  64
    <custom category>:  128

Regarding our predicament, I think explicitly reserving `128` (and above) for Credo runtime errors is our best and most sustainable option. 

We would have to cut the possible custom categories from 3 to 2, but being between a rock and a hard place, this seems like our best bet:

    consistency:        1
    design:             2
    readability:        4
    refactor:           8
    warning:            16
    <custom category>:  32
    <custom category>:  64
    runtime errors      128 (and above)

Danielwsx64:
@rrrene I made the changes according to your comment. Tks for the answer.

fernandoalmeida (approved):

lib/credo/execution/task/convert_cli_options_to_config.ex
```diff
@@ -13,10 +13,7 @@ defmodule Credo.Execution.Task.ConvertCLIOptionsToConfig do
   end
 
   def halt_on_error({:error, error}, exec) do
-    exec
-    |> Execution.put_assign("#{__MODULE__}.error", error)
-    |> Execution.put_assign("credo.exit_status", 128)
```
Danielwsx64: Hi @rrrene . 

I noticed you removed the issue solution when you merged the master into this branch. The solution would be adding an exit status in addition to the halt message. Without the exit status we still have a error message with a success exit code.

I'm not sure if you did it intentionally or it was an oversight.

rrrene: Credo `1.5.x` can not change this long-standing behaviour in a patch release (although it's a bug). 

The next minor release `1.6` will change it and highlight it in the release notes.

If you are in need of a workaround before `1.6` is out, you can look for output to `stderr`, which does not happen on a normal run.

PR #833, opened 2020-12-08
Fix Git check for systems without git

matthew-myers-bose:
This PR fixes `git_present?` for systems that don't have git.

On a system without git available, the previous implementation raises an ErlangError:
```
** (ErlangError) Erlang error: :enoent
    (elixir 1.11.2) lib/system.ex:815: System.cmd("git", ["--help"], [])
```
This is solved by rescuing the error.


rrrene:
@matthew-myers-bose Thx! `v1.5.3` contains your fix :+1:

PR #834, opened 2020-12-08
Adding support for spec checks on private functions

coryodaniel:
I like writing specs for my private functions! Adding an option to enable checking on private functions. Its disabled by default.

To enable:
```elixir
{Credo.Check.Readability.Specs, [include_defp: true]}
```

Added tests for asserting and refuting errors


rrrene:
@coryodaniel :+1: Thx!

PR #839, opened 2020-12-24
Add :exit_status to the list of valid options

TBK145:
There seems to be support for this option (it's even added by default in the generated custom check), but using Credo.Check will fail as it's not in the list op valid options.


rrrene:
Ouch. Thx! :+1:

rrrene:
This is release in `v1.5.4`. Thx again!

PR #848, opened 2021-02-19
Enum.count/2 needs to execute condition, the alternatives do not apply

tverlaan:
I made the `Enum.count/1-2` pattern more strict so it only matches the `/1` and not the `/2`. For reference here is the documentation: https://hexdocs.pm/elixir/Enum.html#count/2. I took that example code in the tests :)

Adding this case was a breeze. Thanks for making it so easy 😄


tverlaan:
Sorry, I changed the PR. Instead of not flagging `Enum.count/2` I added an explanation for how to improve.

Credits go to @TBK145

rrrene:
Great 👍 

I think the optimal solution lies somewhere in between.

Could you refactor the check to provide a different message in line 71, depending on whether we have encountered `Enum.count/1` or `Enum.count/2`?

rrrene:
@tverlaan Thx!

PR #849, opened 2021-02-20
Add check for useless with clause

NickNeck:
Hello,

sometimes, after rewriting my code, it looks like this:
```elixir
with {:ok, map} <- check(input),
     {:ok, result} <- do_something(map) do
  {:ok, result}
end
```
And I have to refactor it to:
```elixir
with {:ok, map} <- check(input) do
  do_something(map)
end
```
However, I don't always notice it right away. So, I have created a check for this, and maybe that could help someone else too.

I have lend me some code from `with_clauses`.

Thanks for Credo!


rrrene:
Will have to have a deeper look, but this is looking good :+1:

lib/credo/check/refactor/useless_with_clause.ex
```diff
@@ -0,0 +1,90 @@
...
+    # If clauses_and_body is a list with at least two elements in it, we think
+    # this might be a call to the special form "with". To be sure of that,
+    # we get the last element of clauses_and_body and check that it's a keyword
+    # list with a :do key in it (the body).
+
+    # We can hard-match on [maybe_body] here since we know that clauses_and_body
+    # has at least two elements.
+    {maybe_clauses, [maybe_body]} = Enum.split(clauses_and_body, -1)
+
+    if Keyword.keyword?(maybe_body) and Keyword.has_key?(maybe_body, :do) do
```
rrrene, line 52: You should be able to refactor this using `Credo.Code.Block.do_block?/1`.

lib/credo/check/refactor/useless_with_clause.ex
```diff
@@ -0,0 +1,90 @@
+defmodule Credo.Check.Refactor.UselessWithClause do
+  use Credo.Check,
+    base_priority: :high,
+    explanations: [
+      check: ~S"""
+      `with` statements are useful when you need to chain a sequence
+      of pattern matches, stopping at the first one that fails.
+
+      If the `match` of the last clause in a with equal to the expression in the
```
rrrene, line 9:
```suggestion
      If the match of the last clause in a `with` statement is identical to the expression in the
```

lib/credo/check/refactor/useless_with_clause.ex
```diff
@@ -0,0 +1,90 @@
...
+      body, the code should be refactored.
```
rrrene, line 10:
```suggestion
      in its body, the code should be refactored to remove the redundant expression.
```

lib/credo/check/refactor/useless_with_clause.ex
```diff
@@ -0,0 +1,90 @@
+defmodule Credo.Check.Refactor.UselessWithClause do
```
rrrene, line 1: I am not a native speaker, but somehow "useless" carries a bad, accusing vibe. While calling a piece of code "useless" is technically the same as calling it "redundant" or "pleonastic", it somehow feels less educational.

Maybe `RedundantWithClauseResult` would be a better name?

NickNeck: I'm right there with you. I'am fine with `RedundantWithClauseResult`.

lib/credo/check/refactor/useless_with_clause.ex
NickNeck: That makes things easier.

PR #852, opened 2021-02-26
Add check for using apply

NickNeck:
Inspired by https://github.com/elixir-lang/elixir/pull/10748

> If the number of arguments is known at compile time, prefer
  `fun.(arg_1, arg_2, ..., arg_n)` as it is clearer than
  `apply(fun, [arg_1, arg_2, ..., arg_n])`.


rrrene:
Will have to have a deeper look, but this is looking good 👍

PR #855, opened 2021-02-27
Add ForbiddenModule Check

spencerdcarlson:
# Proposal
Add `Credo.Check.Warning.ForbiddenModule` that can be configured to prohibit a list of modules.

### Example .credo.exs usage
```elixir
...
checks: [
  {Credo.Check.ForbiddenModule, [modules: [Ecto.Adapters.SQL]]}
]
```

### mix credo explain Credo.Check.ForbiddenModule
Screen Shot 2021-02-26 at 5 45 11 PM

### Example Violation
Screen Shot 2021-02-26 at 5 47 47 PM


NickNeck:
Hello @spencerdcarlson 

that looks great and interesting.

One problem could be that the list of modules depends on the project. I understand Credo more as a checker for general code styles.

If you want to have such boundaries in your project, you could check out https://github.com/sasa1977/boundary

Maybe you can add predefined lists of modules for well-known libs. Like:
`modules: [:ecto, :phoenix]`.

Another feature could be to add all hidden modules to the list of forbidden modules: 
```elixir
:code.all_loaded() |> Enum.filter(fn {m,_} -> match?({_,_,_,_,:hidden,_,_}, Code.fetch_docs(m)) end)
```
In this case, you must also check the root module.
```
ok   : MyApp.Worker using MyApp.HiddenUtil 
error: MyAppWeb.Foo using MyApp.HiddenUtil
```

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Credo.Check.Warning.ForbiddenModule do
+  @moduledoc """
+    Checks for use of a list of modules that are not allowed
+  """
+
```
rrrene, line 5: Please remove the `@moduledoc` as checks do not have them (they are generated).

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+  alias Credo.Code
+
```
rrrene, line 7: Please re-order the `alias` below the `use` statements (as seen in all other checks in Credo).

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+  use Credo.Check,
+    base_priority: :high,
```
rrrene, line 9: Why would this check be of high priority to the regular user?

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+    category: :warning,
```
rrrene, line 10: Why is this a `warning`?

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+    param_defaults: [modules: []],
+    explanations: [
+      check: """
+      These modules are are forbidden.
```
rrrene, line 14: Please put a bit more effort into the explanation of the check's existence. Why would anyone want to deny usage of a certain list of modules?

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+      """,
+      params: [modules: "This check warns about modules that can not be used."]
```
rrrene, line 16: This is a description of the parameter, not the check. Please describe what the parameter does.

Also: Not a native speaker, but shouldn't this be "must" instead of "can"?

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+    ]
+
+  @impl Credo.Check
+  def run(source_file = %SourceFile{}, params) do
+    modules =
+      params
+      |> Params.get(:modules, __MODULE__)
+      |> List.wrap()
```
rrrene, line 24: Please use `Params.get/3` here instead.

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+
+    Code.prewalk(source_file, &traverse(&1, &2, modules, IssueMeta.for(source_file, params)))
+  end
+
+  defp traverse(ast = {:__aliases__, meta, module}, issues, forbidden_modules, issue_meta) do
+    # Catches import and alias statements for a forbidden module
```
rrrene, line 30: Please don't write comments literally describing what happens in the next line.

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+      {ast, [issue_for(issue_meta, meta[:line], module) | issues]}
+    else
+      {ast, issues}
+    end
+  end
+
+  defp traverse(ast, issues, _, _), do: {ast, issues}
+
+  defp forbidden_module?(module, forbidden_modules) do
+    Enum.all?(module, &is_atom/1) && Enum.member?(forbidden_modules, Module.concat(module))
```
rrrene, line 41: Why do we need the `Enum.all?(module, &is_atom/1)`?

lib/credo/check/warning/forbidden_module.ex
```diff
@@ -0,0 +1,56 @@
...
+
+    format_issue(
+      issue_meta,
+      message: "The `#{trigger}` module is not allowed.",
+      trigger: trigger,
+      line_no: line_no
+    )
+  end
+
+  defp mod_string(module), do: ~r/Elixir\./ |> Regex.replace(to_string(module), "")
```
rrrene, line 55: Please use `Credo.Code.Module.name/1`.

test/credo/check/warning/forbidden_module_test.exs
```diff
@@ -0,0 +1,73 @@
...
+    |> to_source_file
+    |> run_check(@described_check)
+    |> refute_issues()
+  end
+
+  #
+  # cases raising issues
+  #
+
+  test "it should report on aliases" do
```
rrrene, line 27: Your missing the most basic case here:

```elixir
      def some_function do
        CredoSampleModule.ForbiddenModule.another_function()
      end
    end
```

rrrene (changes requested):
@spencerdcarlson This seems like a worthwhile endeavour. 👍 

Could please go into more detail re: "Which are the scenarios where you needed/used this?"

test/credo/check/warning/forbidden_module_test.exs
spencerdcarlson: That test case is here on [L63-L72](https://github.com/rrrene/credo/pull/855/files#diff-2dedc3870ab83967d2ae9c6aa3a4d19f542f72d90728b7e56ff9a753060180e2R63-R72). I'll move it up to be first in the list.

lib/credo/check/warning/forbidden_module.ex
spencerdcarlson: I removed this, it was left over from a prior approach to catch inline usage.

lib/credo/check/warning/forbidden_module.ex
spencerdcarlson: When I have used this Check on projects, it is usually critical that a given module is forbidden because it could cause issues. I have no preference here, what would you recommend?

lib/credo/check/warning/forbidden_module.ex
spencerdcarlson: When I have used this Check on projects, it is usually critical that a given module is forbidden because it could cause issues. I have no preference here, what would you recommend?

lib/credo/check/warning/forbidden_module.ex
spencerdcarlson: Updaed.

spencerdcarlson:
> @spencerdcarlson This seems like a worthwhile endeavour. 👍
> 
> Could please go into more detail re: "Which are the scenarios where you needed/used this?"

Hopefully my updated explanation makes it more clear:

>Some modules that are included by a package may be hazardous
      if used by your application. Use this check to allow these modules in
      your dependencies but forbid them to be used in your application.

>Examples:
      The `:ecto_sql` package includes the `Ecto.Adapters.SQL` module,
      but direct usage of the `Ecto.Adapters.SQL.query/4` function, and related functions, may
      cause issues when using Ecto's dynamic repositories.


Because internally [Ecto.Adapters.SQL.query/4](https://github.com/elixir-ecto/ecto_sql/blob/v3.5.4/lib/ecto/adapters/sql.ex#L420-L426) does not call [Repo.get_dynamic_repo/0](https://hexdocs.pm/ecto/Ecto.Repo.html#c:get_dynamic_repo/0) before executing the query, usage the adapter's functions could be hazardous. Instead we need to require developers to use [Ecto.Repo.query/3](https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/sql.ex#L512-L545) instead.

lib/credo/check/warning/forbidden_module.ex
rrrene: I'd just leave it for the user to decide :+1:

lib/credo/check/warning/forbidden_module.ex
rrrene: You're right. Thinking about it, `warning` is probably the most sensible choice here :+1:

rrrene (commented):
After a longer-than-anticipated absence, this check proposal has really grown on me :+1:

One thing I wished this had is the ability to provide a custom message for a forbidden module:

```elixir
# .credo.exs

# ...
    {Credo.Check.Warning.ForbiddenModule, modules: [{Ecto.Adapters.SQL, "jesus, team, please stop using this!!!11"}]},
# ...
```

WDYT?

rrrene:
@spencerdcarlson Any news on this? If you could add the ability to provide a custom message for a forbidden module, we'll make this part of Credo v1.6 :+1:

spencerdcarlson:
> @spencerdcarlson Any news on this? If you could add the ability to provide a custom message for a forbidden module, we'll make this part of Credo v1.6 👍

Thanks, I added the custom message feature. Let me know if there is anything you want me to change.

rrrene (approved):

rrrene:
@spencerdcarlson Thx! 🎉  :+1:

PR #858, opened 2021-03-08
Check UnsafeToAtom along pipe chains

thalesmg:
Currently, if a call to `String.to_atom/1`, `List.to_atom/1`,
`Module.concat/{1,2}`, `:erlang.list_to_atom/1` and
`:erlang.binary_to_atom/2` is in a pipe (`|>`) chain, it is not
detected by the `UnsafeToAtom` check.

This change alters the prewalk check to unpipe those chains and insert
a phantom argument in the chain to make the arity "correct" and allow
detection.

### DIFF
--- a/lib/credo/check/warning/unsafe_to_atom.ex
+++ b/lib/credo/check/warning/unsafe_to_atom.ex
@@ -48,8 +48,6 @@ defmodule Credo.Check.Warning.UnsafeToAtom do
   end
 
   defp traverse({:|>, _loc, _args} = ast, issues, _issue_meta) do
-    unpiped = Macro.unpipe(ast)
-
     case Macro.unpipe(ast) do
       [first | rest] ->
         unpiped =
--- a/lib/credo/check/warning/unsafe_to_atom.ex
+++ b/lib/credo/check/warning/unsafe_to_atom.ex
@@ -47,16 +47,22 @@ defmodule Credo.Check.Warning.UnsafeToAtom do
     {nil, issues}
   end
 
-  defp traverse({:|>, loc, args}, issues, _issue_meta) do
-    {args, _} = Enum.reduce(args, {[], first?: true}, &mark_as_in_pipe/2)
-    ast = {:|>, loc, Enum.reverse(args)}
-    {ast, issues}
+  defp traverse(
+         {:|>, _meta1, [_lhs, {{:., _meta2, call}, meta, args}]} = ast,
+         issues,
+         issue_meta
+       ) do
+    case get_forbidden_pipe(call, args) do
+      {bad, suggestion} ->
+        {ast, issues_for_call(bad, suggestion, meta, issue_meta, issues)}
+
+      nil ->
+        {ast, issues}
+    end
   end
 
   defp traverse({{:., _loc, call}, meta, args} = ast, issues, issue_meta) do
-    in_pipe? = Keyword.get(meta, :in_pipe?, false)
-
-    case get_forbidden_call(call, args, in_pipe?: in_pipe?) do
+    case get_forbidden_call(call, args) do
       {bad, suggestion} ->
         {ast, issues_for_call(bad, suggestion, meta, issue_meta, issues)}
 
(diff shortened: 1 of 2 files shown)


brunorafa (approved):
nice and good.

rrrene:
You're right, this is a bug.

But we can not solve it this way. :-(

> [...] alters the prewalk check to unpipe those chains and insert
> a phantom argument [...] 

We won't *alter* an AST while we are analysing it, just so we don't have to refactor the existing code.

Let's find another solution :+1:

thalesmg:
> You're right, this is a bug.
> 
> But we can not solve it this way. :-(
> 
> > [...] alters the prewalk check to unpipe those chains and insert
> > a phantom argument [...]
> 
> We won't _alter_ an AST while we are analysing it, just so we don't have to refactor the existing code.
> 
> Let's find another solution +1

How about we only change the AST metadata for nodes inside a pipe chain, but not the structure of the AST?
I thought of adding a simple flag value indicating that a call (`{{:., _loc, call}, meta, args}`) is inside a pipe, and forward that info to `get_forbidden_call` to check the true arity.

What do you think?

thalesmg:
I've made the change [here](https://github.com/rrrene/credo/pull/858/commits/5380b9780cc839edf1f34a1074635eb8b16e571d). Is this more acceptable?

rrrene:
You're onto something here, but it is still too complex. The logic needs to be safe, and the first step towards safety is simplicity.

We have two different use cases: One with pipes and one without. Both use cases have different implementations, because the arity we have to match is different.

We can find all the pipes like this:

```elixir
  defp traverse(
         {:|>, _meta1, [_lhs | [{{:., _meta2, call}, meta, args} = ast]]},
         issues,
         issue_meta
       ) do
    case get_forbidden_pipe(call, args) do
      {bad, suggestion} ->
        {ast, issues_for_call(bad, suggestion, meta, issue_meta, issues)}

      nil ->
        {ast, issues}
    end
  end
```

So then we can match on those right hand sides of the pipes:

```elixir
  defp get_forbidden_pipe([:erlang, :list_to_atom], []) do
    {":erlang.list_to_atom/1", ":erlang.list_to_existing_atom/1"}
  end

  defp get_forbidden_pipe([:erlang, :binary_to_atom], [_]) do
    {":erlang.binary_to_atom/2", ":erlang.binary_to_existing_atom/2"}
  end

  defp get_forbidden_pipe([{:__aliases__, _, [:String]}, :to_atom], []) do
    {"String.to_atom/1", "String.to_existing_atom/1"}
  end

  defp get_forbidden_pipe([{:__aliases__, _, [:List]}, :to_atom], []) do
    {"List.to_atom/1", "List.to_existing_atom/1"}
  end

  defp get_forbidden_pipe([{:__aliases__, _, [:Module]}, :concat], []) do
    {"Module.concat/1", "Module.safe_concat/1"}
  end

  defp get_forbidden_pipe([{:__aliases__, _, [:Module]}, :concat], [_]) do
    {"Module.concat/2", "Module.safe_concat/2"}
  end

  defp get_forbidden_pipe([{:__aliases__, _, [:Jason]}, decode] = call, args)
       when decode in [:decode, :decode!] do
    get_forbidden_call(call, args)
  end

  defp get_forbidden_pipe(_, _) do
    nil
  end
```

The question here is: Why would we want to carry around implicit state when we can make it explicit?

P.S.: In your edit you're using pattern matches on keyword lists in function heads. Never do that. Keyword lists are order dependent, so your `in_pipe: true` only matches if `{:in_pipe, true}` is the first element in the list.

thalesmg:
Like [so](https://github.com/rrrene/credo/pull/858/commits/4512b91786aaa70fac43781c8b7357fb059ba17f)?

rrrene:
Did you skip the `get_forbidden_pipe/2` function head for matching `Jason.decode/2` on purpose?

thalesmg:
> Did you skip the `get_forbidden_pipe/2` function head for matching `Jason.decode/2` on purpose?

Yes, because the same clause in `get_forbidden_call/2` is not restricted by arity, unlike the other forbidden clauses.

Also, you'll notice that I did not include a `get_forbidden_pipe([{:__aliases__, _, [:Module]}, :concat], [_])` clause, because it overlaps with `get_forbidden_call([{:__aliases__, _, [:Module]}, :concat], [_])` and was generating duplicate warnings. Since `get_forbidden_call/2` always triggers for those cases, I omitted those clauses.

rrrene:

PR #859, opened 2021-03-10
Allow to select only required modules for AliasUsage check

ostap0207:
Currently this check runs on all modules, however, sometimes it's useful
to only run it on a subset of modules. For example, one might want to
only include modules with project top-level module
`ProjectModule.<Module>`.

This was originally proposed in https://github.com/rrrene/credo-proposals/issues/2

### DIFF
--- a/lib/credo/check/design/alias_usage.ex
+++ b/lib/credo/check/design/alias_usage.ex
@@ -12,7 +12,8 @@ defmodule Credo.Check.Design.AliasUsage do
                           Stream String StringIO Supervisor System Task Time
                           Tuple URI Version],
       if_nested_deeper_than: 0,
-      if_called_more_often_than: 0
+      if_called_more_often_than: 0,
+      only: nil
     ],
     explanations: [
       check: """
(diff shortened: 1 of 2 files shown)


rrrene:
Does the [`:files` param](https://hexdocs.pm/credo/check_params.html#files) provide the same utility?

ostap0207:
It was your original proposal that I followed https://github.com/rrrene/credo-proposals/issues/2#issuecomment-278592895.
Files specifies on which files to run the check.
I still want to run the check on all files but within a file I want to only take into account specific aliases and ignore others.

ostap0207:
Perhaps my description `Regex or a list of regexes that specifies which modules to include for this check.` is not clear enough and its too similar to `files` param.

rrrene:
I misread your initial description. This PR is fine, I just have to find some time to fix GitHub Actions :+1:

rrrene:
@ostap0207 Thx! This will be part of Credo 1.6 :+1:

PR #861, opened 2021-03-15
Add useless block pipe check

NickNeck:
This check looks for code like
```elixir
          list
          |> length()
          |> case do
            0 -> :none
            1 -> :one
            _ -> :many
          end
```
and suggesting to change in
```elixir
          case length(list) do
            0 -> :none
            1 -> :one
            _ -> :many
          end
```


rrrene:
Repeating my impression from another PR: I think "useless" sounds to harsh, especially since the presented is not redundant or ineffective. This seems like a personal readability preference more than anything (which is fine, but again, calling something "useless" sounds a bit immature in my book).

Can we find a wording that conveys the suboptimal nature of the code in question and differentiates this check from the `BlockPipe` check?

NickNeck:
You are right. Maybe we can call it `SingleBlockPipe`, because the check has something roughly in common with `SinglePipe` and `BlockPipe`. The explanation could become: `A single pipe (`|>`) should not be used to pipe into blocks.`

rrrene:
`SingleBlockPipe` sounds good to me! 👍

rrrene:
@NickNeck Could you update this PR? 

1. `.credo.exs` has a merge conflict
2. the check is still called `Credo.Check.Refactor.UselessBlockPipe` in the conflict
3. please move the check to `readability` since it's suggestion improves comprehension, but does not impact the correctness of the code

rrrene:
@NickNeck Thx for your contributions :+1:

PR #863, opened 2021-03-19
Update example of checking properties of issue list to assert_issues

J3RN:
The documentation has an example of validating a property of a list of issues with `assert_issue`.

The anonymous function for `assert_issue` only receives a single issue, I think this example was supposed to use `assert_issues`, which takes a list of issues, instead.


rrrene:
@J3RN Thx! :+1:

PR #865, opened 2021-03-23
Fix CI by switching to erlef/setup-elixir action

begedin:
We had a similar issue with our CI. Some interaction between Ubuntu 20.04 and the crypto library caused all our builds to fail.

The easiest solution is to switch to the action's new home. It's been adopted by the The Erlang Ecosystem Foundation recently.

This is me just blindly shooting as an attempt to do a quick fix, so feel free to close if it doesn't work.

### DIFF
--- a/.github/workflows/bugfix-reproducer.yml
+++ b/.github/workflows/bugfix-reproducer.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.1
 
-      - uses: actions/setup-elixir@v1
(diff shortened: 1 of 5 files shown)


begedin:
It seems to have move past the initial build step for every check, but CI is still failing, so not sure If I'm doing something wrong, or if it needs to run manually in some way.

rrrene:
I think `erlef/setup-elixir`  has since been renamed to  `erlef/setup-beam`. I made the necessary changes.

Thanks for highlighting this :+1:

PR #867, opened 2021-03-31
Changing predicate to arity

panoramix360:
Changing predicate rules to check funcions with arity greater than 0.

I created three more tests too


rrrene:
@panoramix360 Thx :+1:

PR #868, opened 2021-04-01
Add new check for bad ecto migration timestamps

bglusman:
Hopefully this is a welcome PR and not out of scope since it's targeting an issue from a specific hex library?  Just thought it would be useful as I'm guilty of occasionally hand editing these files and have been bit by this once or twice!  Thanks, let me know if I can do anything further to help/improve!

### DIFF
--- a/lib/credo/check/warning/bad_migration_timestamp.ex
+++ b/lib/credo/check/warning/bad_migration_timestamp.ex
@@ -1,0 +1,50 @@
+defmodule Credo.Check.Warning.BadMigrationTimestamp do
+  use Credo.Check,
+    base_priority: :high,
+    param_defaults: [included: ["priv/repo/migrations/*.exs"]],
+    explanations: [
+      check: """
+      If a migration file name is hand edited it is possible to have an aberrant timestamp
+      which will always be applied last no matter how many later migrations are added
+
+      This check ensures that migration files do not have such timestamps.
+      """
+    ]
+
+  @test_files_containing_timestamp ~r/(?<timestamp>\d{14}\d*).*.exs$/
+
+  alias Credo.SourceFile
+
+  @doc false
+  def run(%SourceFile{filename: filename} = source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    if map = Regex.named_captures(@test_files_containing_timestamp, filename) do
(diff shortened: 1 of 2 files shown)


bglusman:
I see that tests are failing, but that seems to be the case on master and for all the PR's...  not sure if anything I can do to help there, but my new tests are passing in any case locally.

lib/credo/check/warning/bad_migration_timestamp.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Credo.Check.Warning.BadMigrationTimestamp do
+  use Credo.Check,
+    base_priority: :high,
+    param_defaults: [included: ["priv/repo/migrations/*.exs"]],
+    explanations: [
+      check: """
+      If a migration file name is hand edited it is possible to have an aberrant timestamp
```
dmarcoux, line 7:
```suggestion
      If a migration file name is manually edited, it is possible to have a wrong timestamp
```

lib/credo/check/warning/bad_migration_timestamp.ex
```diff
@@ -0,0 +1,50 @@
...
+      which will always be applied last no matter how many later migrations are added
```
dmarcoux, line 8:
```suggestion
      In this case, it will always be applied last, even if other migrations were created afterwards.
```

dmarcoux, line 8:
```suggestion
      If a migration file name is manually edited, it could have a wrong timestamp.
      In this case, it will always be applied last, even if other migrations were created afterwards.
```

rrrene:
I apologize for the age/inactivity on this. I should have done a better job at resolving this properly. 😥 

Please feel free to re-open this at your discretion.

bglusman:
Hmm, I don't seem to have permission to reopen this.  I still think it would be a useful check to add if you're open to?

rrrene:
Maybe I misunderstood the code, especially the code comments  (why tolerate a year and/or a month of difference?).

All in all, this seemed to me like this is a solution to manual errors that occur in your team/codebase. Since this concerns itself with timestamps and detecting errors in hand-edited timestamps, I am not sure if we can solve this in a manner that covers all use-cases, which is exactly what custom checks were invented for :+1:

PR #869, opened 2021-04-02
Allow ignoring guards in CyclomaticComplexity calculation

rodrigues:
Hello credo team! 💜

With this change, credo allows the user to ignore guards when calculating functions CC.

When you have public functions receiving and validating a complex input, pattern matching on head, adding guards can very often exhaust easily the CC calculation, leaving no room for the body.

So I thought this could be a setting on this check, for the projects that want to use guards feature more freely on the function head, but still would like to have max complexity strict for the function body. You think that makes sense?

Thanks!


rrrene:
@rodrigues Thx :+1:

PR #870, opened 2021-04-17
Add GitHub output format (for GitHub Actions)

arnodirlam:
This PR adds a new output `--format github` for adding checks in [GitHub Actions](https://docs.github.com/actions) (via [workflow commands](https://docs.github.com/actions/reference/workflow-commands-for-github-actions)).

The checks are annotated in the workflow logs, grouped by file, as well as inline in the "compare files" view.

Please let me know if this is interesting and whether it needs any changes 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ Some IDEs and editors are able to run Credo in the background and mark issues in
 * [flycheck](https://www.flycheck.org/en/latest/languages.html#elixir) - Emacs syntax checking extension
 * [kakoune](https://github.com/mawww/kakoune/wiki/Lint#elixir) - Config for linting support in Kakoune editor
 
+### Continuous Integration
+
+* [GitHub Actions](https://docs.github.com/actions) - runs workflows within GitHub repositories (run `mix credo --format github` to create checks)
+
 ### Automated Code Review
 
 * [Codacy](https://www.codacy.com/) - checks your code from style to security, duplication, complexity, and also integrates with coverage.
--- a/guides/commands/suggest_command.md
+++ b/guides/commands/suggest_command.md
@@ -129,7 +129,7 @@ $ mix credo --files-excluded ./test/**/*.exs
 
 ### `--format`
 
-Display the list in a specific format (json, flycheck, or oneline)
+Display the list in a specific format (json, flycheck, github, or oneline)
 
 ```bash
 $ mix credo --format json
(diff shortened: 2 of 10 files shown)


arnodirlam:
Is this something of interest for Credo? @rrrene 
Should I have opened an issue first?

rrrene:
@arnodirlam Hi, sorry for the late reply. I am sympathetic to this in principle, but I was asking myself if we really need to have a formatter which is this GitHub specific or if we should strive for a generic solution which is also recognized by GitHub Actions (e.g. doing sth. like printing to `stderr`  and/or prefixing the issues with `Warning:`).

Doesn't GH Actions pick up warning by other linters "automatically"?

WDYT?

btkostner:
Just want to drop by and say this would be a welcomed addition for my team's CI process.

As for answering your question, GitHub can do this in a lot of different ways. One way is for third party libraries to add a formatter like this when the library is being ran in a GitHub action (detectable with env values.) Alternatively, most of the `setup-*` actions like `setup-elixir` include ["problem matchers"](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) that use regex to parse output and display it inline.

The `setup-elixir` problem matchers: https://github.com/erlef/setup-beam/blob/main/.github/elixir-matchers.json
The `eslint` problem matchers in `setup-node`: https://github.com/actions/setup-node/blob/main/.github/eslint-stylish.json

It might be worth talking upstream to @erlef about adding problem matchers for common tooling like Credo.

btkostner:
I was working on regex for this, and it seems we can get this parsing from the normal Credo output (though it doesn't look to have all of the features this PR has).

https://github.com/btkostner/setup-beam-action-test/commit/827fa95ce04bfd150af38f0adae252ef450f49ad

arnodirlam:
That's a nice approach indeed. I didn't know about problem matchers, really nice!

Although the approach depends on the GitHub Action being used, this should be fine, as [erlef/setup-beam](https://github.com/erlef/setup-beam) seems to be the go-to Action by now. Adding it there makes a lot of sense then 👍 

Would you mind opening an issue in their repo and referencing this PR? I'd close this PR then, if you agree 🙂 @btkostner

btkostner:
Sounds good. I'm currently waiting for one of my PRs to get merged for https://github.com/erlef/setup-beam before I start adding other problem matchers, but I have a working one [here](https://github.com/btkostner/setup-beam-action-test/blob/master/.github/problem-matchers.json#L46) if you want to test and make sure it covers everything!

btkostner:
@arnodirlam The PR has been merged in `setup-beam` and will be available when they cut a new release :confetti_ball:

PR #875, opened 2021-06-11
Revert #867 and #869

rrrene:
The contents of these PRs somehow broke the test suite, i.e.

    mix test

times out during random integration tests.

It is unclear at this point, what caused this:

1. Reverting only #867 *or* #869 did not fix it.
2. Reverting only the changes introduced in test/ or lib/ by those PRs did not fix it.

### DIFF
--- a/lib/credo/check/readability/predicate_function_names.ex
+++ b/lib/credo/check/readability/predicate_function_names.ex
@@ -77,7 +77,7 @@ defmodule Credo.Check.Readability.PredicateFunctionNames do
 
   defp issues_for_definition(op, body, issues, issue_meta) do
     case Enum.at(body, 0) do
-      {name, meta, _} ->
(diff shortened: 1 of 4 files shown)


rrrene:
@panoramix360 I have no idea why your changes are causing the integration tests to time out (spent a week of evenings trying to figure it out, to no avail). Thus I am reverting the changes 😭 

If you can figure out a way to include the changes without the described side effect, I will gladly accept a PR for that :+1:

panoramix360:
Strange, I'll try to fix this in this weekend. Are ou talking about the integration tests on github?

You reverted on master but I can get the commit again to make another PR?

panoramix360:
@rrrene I tried to run tests on master on my machine, but 58 tests are failing for no reason, do you happen to know why?

It seems something to do with encoding

PR #886, opened 2021-08-03
Fix RedundantWithClauseResult

### LINKED ISSUE #885: RedundantWithClauseResult false positive
### Environment

* Credo version (`mix credo -v`): 1.6.0-dev-ref.master.6d195932
* Erlang/Elixir version (`elixir -v`): 1.12.2 (OTP 24.0..5)

```elixir
defmodule Test do
  def test(conn) do
    with ["Basic " <> token] <- get_req_header(conn, "authorization"),
         {:ok, decoded_token} <- token |> String.trim() |> Base.decode64(),
         [client_id, client_secret] <- String.split(decoded_token, ":"),
         {:ok, client} <- authorize(id: client_id, secret: client_secret) do
      {:ok, client}
    else
      _ -> :error
    end
  end

  def authorize(_params) do
    {:error, "ouch"}
  end
end
```

### Expected outcome
Don't apply `RedundantWithClauseResult` if `else` block is present

### Actual outcome
```
┃ [F] ↗ the last clause in `with` is redundant
┃       lib/test.ex:3 #(Test.test)
```

NickNeck:
Closes #885

### DIFF
--- a/lib/credo/check/refactor/redundant_with_clause_result.ex
+++ b/lib/credo/check/refactor/redundant_with_clause_result.ex
@@ -57,7 +57,7 @@ defmodule Credo.Check.Refactor.RedundantWithClauseResult do
   end
 
   defp split(clauses_and_body) do
-    case Block.do_block?(clauses_and_body) do
+    case Block.do_block?(clauses_and_body) and not Block.else_block?(clauses_and_body) do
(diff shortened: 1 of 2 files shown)


rrrene:
@NickNeck :+1:

PR #889, opened 2021-08-12
Add a check for IO.puts

mosic:
I was surprised to see that there was a check for `IO.inspect` and not for `IO.puts`, so I added one. It's almost an exact copy of the `IO.inspect` code.

Btw, thanks for your hard work on Credo! Saved me a bunch of time over the years.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -182,6 +182,7 @@
           {Credo.Check.Refactor.AppendSingleItem, []},
           {Credo.Check.Refactor.DoubleBooleanNegation, []},
           {Credo.Check.Refactor.FilterReject, []},
+          {Credo.Check.Refactor.IoPuts, []},
           {Credo.Check.Refactor.MapMap, []},
           {Credo.Check.Refactor.ModuleDependencies, []},
           {Credo.Check.Refactor.NegatedIsNil, []},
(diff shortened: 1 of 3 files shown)


rrrene:
The difference between `IO.puts` and `IO.inspect` is that the former is used to output strings to `stdout` while the latter is an expensive operation outputting raw Elixir terms to `stdout`, which is why we assume that this is not used for regular program output in most cases.

I do not really get the rationale behind warning on usage of `IO.puts`. How do you output information to the console when using this check?

mosic:
That's a fair assesment, I meant to add it as an optional check. For example, in the project I'm working on, we use `Logger` statements anywhere, which automatically include the metadata etc.

`IO.puts` and `IO.inspect` obviously don't have all of the benefits logging statements do - but they're still useful for quick debugging, and sometimes they get left behind and caught in the PR review process.

Does that make sense? If it's not a common enough use-case, it's fine not to merge this.

rrrene:
> Does that make sense?

Absolutely. :+1: 

I would suggest the following:

1. Let's reflect your reasoning in the explanation for the check.
2. Move the check from the `Warning` category to the `Refactor` category.

WDYT?

mosic:
Sorry for the late response - I was on vacation. That sounds good!

I updated the PR with the changes you proposed. Feel free to change the wording, wasn't sure if I should add more to the check description.

rrrene:
@mosic Thx for your contribution! :+1:

PR #897, opened 2021-10-01
Fix Credo.Check.Refactor.Apply

### LINKED ISSUE #892: Credo.Check.Refactor.Apply shouldn't apply if the function name is sourced from a variable
Our function name is sourced from a variable like this
```
apply(stub_module, function, [channel, request])
```
Which means we can't call `stub_module.function(channel, request)` 

Maybe there is a way but I can't seem to find a way. If there is we should add it to the docs. Or if it isn't possible we should have this check ignore this scenario.

NickNeck:
closes #892

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-erlang 24.0
+erlang 24.1
 elixir 1.12.1-otp-24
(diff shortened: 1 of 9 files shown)


rrrene:
@NickNeck Thx! :+1:

PR #904, opened 2021-10-25
Clarify verbiage for Readability.MultiAlias check

J3RN:
The statement "Avoid multi aliases ... consider removing" confused several members of our team, who assumed that this was indicating that a single module was aliased multiple times.

With this PR, I am trying to clarify what the actual issue is and how best to remediate it.


rrrene:
@J3RN Thx! :+1:

PR #905, opened 2021-11-01
Add new check for one-arity functions in pipes

NickNeck:
Some style guides propose:
> Use parentheses for one-arity functions when using the pipe operator (|>)
> ```elixir
> # not preferred
> some_string |> String.downcase |> String.trim
> 
> # preferred
> some_string |> String.downcase() |> String.trim()
> ```
This PR provides a check for that.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -115,6 +115,7 @@
           {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
           {Credo.Check.Readability.VariableNames, []},
           {Credo.Check.Readability.WithSingleClause, []},
+          {Credo.Check.Readability.OneArityFunctionInPipe, []},
 
           #
           ## Refactoring Opportunities
--- a/.credo.exs
+++ b/.credo.exs
@@ -115,7 +115,6 @@
           {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
           {Credo.Check.Readability.VariableNames, []},
           {Credo.Check.Readability.WithSingleClause, []},
-          {Credo.Check.Readability.OneArityFunctionInPipe, []},
 
           #
           ## Refactoring Opportunities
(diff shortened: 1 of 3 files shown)


lib/credo/check/readability/one_arity_function_in_pipe.ex
```diff
@@ -0,0 +1,39 @@
...
+    explanations: [
+      check: """
+      Use parentheses for one-arity functions when using the pipe operator (|>).
+
+          # not preferred
+          some_string |> String.downcase |> String.trim
+
+          # preferred
+          some_string |> String.downcase() |> String.trim()
+      ```
```
rands0n, line 13: Did you open this ``` in someplace before?

.credo.exs
```diff
@@ -115,6 +115,7 @@
           {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
           {Credo.Check.Readability.VariableNames, []},
           {Credo.Check.Readability.WithSingleClause, []},
+          {Credo.Check.Readability.OneArityFunctionInPipe, []},
```
rrrene, line 118: We should not add checks as default, esp. when it's rules that "some style guides" propose :+1:

lib/credo/check/readability/one_arity_function_in_pipe.ex
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  defp traverse(ast, issues, issue_meta) do
+    case issue(ast, issue_meta) do
+      nil -> {ast, issues}
+      issue -> {ast, [issue | issues]}
+    end
+  end
+
+  defp issue({:|>, _meta, [_left, {name, meta, nil}]}, issue_meta) when is_atom(name) do
```
rrrene, line 29: This should be matched in `traverse/3` not in the issue-creating function (see every other check).

rrrene (changes requested):
This seems good, just a few minor nitpicks :+1:

rrrene:

PR #906, opened 2021-11-03
Support elixir 1.13 tokenizer results

v0idpwn:
There are some failing tests that seem related to 1.13, but I'm not confident whether fixing them is required.


rrrene:
@v0idpwn Thx!

PR #914, opened 2021-11-18
fix refactoring example

woylie:
Just saw this in the check output.


rrrene:
It's pretty amazing that nobody, including me, ever noticed this. Thank you!

PR #920, opened 2021-11-25
Fix #919: add `:in_op` to parentheses in conditon

### LINKED ISSUE #919: False positive: "The condition of `if` should not be wrapped in parentheses."
### Precheck

Similar closed issues I came across:
* https://github.com/rrrene/credo/issues/164
* https://github.com/rrrene/credo/issues/461

### Environment

* Credo version (`mix credo -v`): `1.6.1-ref.main.cb6c59b+uncommittedchanges`
* Erlang/Elixir version (`elixir -v`): `Elixir 1.12.3 (compiled with Erlang/OTP 24)`
* Operating system (`sw_vers`): `macOS 11.6`

Checking membership of an element in a [`MapSet`](https://hexdocs.pm/elixir/1.12/MapSet.html) using [`in/2`](https://hexdocs.pm/elixir/1.12/Kernel.html#in/2) rather than [`MapSet.member?/2`](https://hexdocs.pm/elixir/1.12/MapSet.html#member?/2).

Contrived example:

```elixir
defmodule Example do
  @moduledoc false

  def example() do
    if (5 - 4) in MapSet.new([1]) do
      :ok
    else
      :error
    end
  end

  def a(), do: nil
end
```

`mix format` will add parentheses around the expression `5 - 4`. 

Regarding the trailing `a/0` function:
For some reason, I had trouble triggering the warning _without_ a trailing function. Once added, it seems to consistently trigger the warning.

### Expected outcome

No `Readability` issue.

### Actual outcome

from `mix credo explain`:

```
...
┃   [R] Category: readability
┃    ↗  Priority: high
┃
┃       The condition of `if` should not be wrapped in parentheses.
┃       lib/credo_example.ex:5:5 (Example.example)
┃
┃    __ CODE IN QUESTION
┃
┃     3
┃     4   def example() do
┃     5     if (5 - 4) in MapSet.new([1]) do
┃           ^^
┃     6       :ok
┃     7     else
┃
┃    __ WHY IT MATTERS
...
```

ed-flanagan:
Fixes #919

### DIFF
--- a/lib/credo/check/readability/parentheses_in_condition.ex
+++ b/lib/credo/check/readability/parentheses_in_condition.ex
@@ -106,6 +106,12 @@ defmodule Credo.Check.Readability.ParenthesesInCondition do
     false
   end
 
+  # matches:  if( something ) in something_else do
+  #                           ^^
+  defp check_for_closing_paren(_start, {:")", _}, [{:in_op, _, _} | _tail], _prev_head) do
+    false
+  end
+
   # matches:  if( 1 + foo ) / bar > 0 do
   #                         ^
   defp check_for_closing_paren(_start, {:")", _}, [{:mult_op, _, _} | _tail], _prev_head) do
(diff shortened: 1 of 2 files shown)


rrrene:
Looking good 👍  Will merge soon!

PR #923, opened 2021-12-04
Extend Credo.Check.Warning.IoInspect to consider Elixir.IO.inspect in addition to IO.inspect

juhalehtonen:
As originally proposed in https://github.com/rrrene/credo-proposals/issues/72, this PR extends the `Credo.Check.Warning.IoInspect` check to also consider calls to `Elixir.IO.inspect` under the same rule, as the two calls are functionally the same.

### DIFF
--- a/test/credo/check/warning/io_inspect_test.exs
+++ b/test/credo/check/warning/io_inspect_test.exs
@@ -63,4 +63,17 @@ defmodule Credo.Check.Warning.IoInspectTest do
     |> run_check(@described_check)
     |> assert_issue()
   end
+
+  test "it should report a violation /4" do
+    """
+    defmodule CredoSampleModule do
+      def some_function(parameter1, parameter2) do
+        Elixir.IO.inspect parameter1 + parameter2
+      end
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check)
+    |> assert_issue()
+  end
 end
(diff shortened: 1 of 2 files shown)


test/credo/check/warning/io_inspect_test.exs
```diff
@@ -63,4 +63,17 @@ defmodule Credo.Check.Warning.IoInspectTest do
     |> run_check(@described_check)
     |> assert_issue()
   end
+
+  test "it should report a violation /4" do
```
juhalehtonen, line 67: The Checks tab on GitHub suggests that all tests pass, yet there is a failure report here 🤔 What would that mean?

rrrene: Where do you see this?

juhalehtonen: On the diff view at https://github.com/rrrene/credo/pull/923/files, per the screenshot :)

gh

rrrene: Ah, yeah. That's a special GitHub actions workflow I created to automatically check if added tests actually fail in case one reverts the changes to `lib/`.

So this fail is actually a good thing :+1:

juhalehtonen: Ohhh, gotcha! Thanks for clearing that up :)

rrrene:
@juhalehtonen Thx!

PR #926, opened 2021-12-07
Detect end of ~A""" style sigils correctly when searching for config comments

paulstatezny:
# Summary

This fixes a bug that presents itself with the following behavior:

The existence of a heredoc-style sigil (with `"""` delimiters) causes credo to ignore all config comments appearing after the sigil.

# Impact

This affects code written with any library that defines a sigil and is commonly used in heredoc style, such as [LiveView](https://hex.pm/packages/phoenix_live_view) (with HEEx `~H` sigils) and [Surface](https://hex.pm/packages/surface) (with `~F` sigils).

# Explanation of Fix

Internally, there is a mechanism that replaces the contents of sigils with spaces in order to avoid misinterpreting their contents as config comments. This mechanism was unable to properly detect the end of the sigil. As a result, code like this

```elixir
defmodule MyModule do
  def render do
    ~F"""
    """
  end

  # BUG: The following config comment will be ignored by Credo
  # credo:disable-for-next-line
  def foo, do: :ok
end
```

would be replaced with

```elixir
defmodule MyModule do
  def render do
    ~F"""
       
     

                                                              
                               
                  
   
```

before searching it for `# credo:`-style config comments. (Note, the last 7 lines were replaced with space characters)

### DIFF
--- a/lib/credo/code/sigils.ex
+++ b/lib/credo/code/sigils.ex
@@ -271,17 +271,6 @@ defmodule Credo.Code.Sigils do
       parse_removable_sigil(t, acc <> replacement <> replacement, unquote(sigil_end), replacement)
     end
 
-    if sigil_end != "\"" do
-      defp parse_removable_sigil(
-             <<"\""::utf8, t::binary>>,
-             acc,
-             unquote(sigil_end),
-             replacement
-           ) do
-        parse_removable_sigil(t, acc <> replacement, unquote(sigil_end), replacement)
-      end
-    end
-
     defp parse_removable_sigil(
            <<unquote("\\#{sigil_end}")::utf8, t::binary>>,
            acc,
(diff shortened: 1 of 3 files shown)


lib/credo/code/sigils.ex
```diff
@@ -271,17 +271,6 @@ defmodule Credo.Code.Sigils do
...
-             acc,
-             unquote(sigil_end),
-             replacement
-           ) do
-        parse_removable_sigil(t, acc <> replacement, unquote(sigil_end), replacement)
-      end
-    end
-
     defp parse_removable_sigil(
            <<unquote("\\#{sigil_end}")::utf8, t::binary>>,
```
paulstatezny, line 275: This `parse_removable_sigil` clause and the one below it are designed to catch the `"""` sigil end marker.

The `parse_removable_sigil` clause above matches on a single `"` character. By having it above the "sigil end" clauses, it blocked `"""` sigil end markers from being detected. This is because a string starting with `"""` will match on both `<<"\""::utf8, t::binary>>` (a single `"`) and <<"\"\"\""::utf8, t::binary>> (a triple `"""`).

By moving the clause matching `"` below these "sigil end" clauses, we allow them to match correctly when the sigil ending is reached.

paulstatezny:
I see that 4 of 5 CI runs failed, but it appears unrelated to this PR. (Perhaps they failed on `master` when I branched?)

Let me k ow if you want me to do anything to remedy this @rrrene. Thanks again!

rrrene:
@paulstatezny Thx! The failing tests did not have anything to do with your changes :+1:

PR #933, opened 2021-12-31
Add nested function calls check

Stratus3D:
This PR adds a check named `Credo.Check.Readability.NestedFunctionCalls` that checks for nested function calls that could be converted to pipelines. I created this check to catch violations of the rule about nested function calls specified in the Christopher Adam's style guide (https://github.com/christopheradams/elixir_style_guide#pipe-operator)

It tries to be smart enough to handle things like access protocol, string interpolation, and function calls with no arguments. See the tests for examples.

Context:

* https://github.com/christopheradams/elixir_style_guide/issues/143#issuecomment-394191496

### DIFF
--- a/lib/credo/check/readability/nested_function_calls.ex
+++ b/lib/credo/check/readability/nested_function_calls.ex
@@ -111,16 +111,6 @@ defmodule Credo.Check.Readability.NestedFunctionCalls do
   # Module.function_call()
   defp valid_chain_start?({{:., _, _}, _, []}), do: true
 
-  # Elixir <= 1.8.0
-  # '__#{val}__' are compiled to String.to_charlist("__#{val}__")
-  # we want to consider these charlists a valid pipe chain start
-  defp valid_chain_start?({{:., _, [String, :to_charlist]}, _, [{:<<>>, _, _}]}), do: true
-
-  # Elixir >= 1.8.0
-  # '__#{val}__' are compiled to String.to_charlist("__#{val}__")
-  # we want to consider these charlists a valid pipe chain start
-  defp valid_chain_start?({{:., _, [List, :to_charlist]}, _, [[_ | _]]}), do: true
-
   # Kernel.to_string is invoked for string interpolation e.g. "string #{variable}"
   defp valid_chain_start?({{:., _, [Kernel, :to_string]}, _, _}), do: true
 
--- a/lib/credo/check/readability/nested_function_calls.ex
+++ b/lib/credo/check/readability/nested_function_calls.ex
@@ -13,12 +13,12 @@ defmodule Credo.Check.Readability.NestedFunctionCalls do
 
       The code in this example ...
 
-          Enum.shuffle(integers(3))
+          Enum.shuffle(Enum.uniq([1,2,3,3]))
 
       ... should be refactored to look like this:
 
-          3
-          |> integers()
+          [1,2,3,3]
+          |> Enum.uniq()
           |> Enum.shuffle()
 
       Nested function calls make the code harder to read. Instead, break the
--- a/lib/credo/check/readability/nested_function_calls.ex
+++ b/lib/credo/check/readability/nested_function_calls.ex
@@ -1,0 +1,126 @@
+defmodule Credo.Check.Readability.NestedFunctionCalls do
+  use Credo.Check,
+    base_priority: :high,
+    tags: [:controversial],
+    param_defaults: [min_pipeline_length: 2],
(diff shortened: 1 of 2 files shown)


lib/credo/check/readability/nested_function_calls.ex
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  # A call with arguments
+  defp traverse({{:., _loc, call}, meta, args} = ast, {_, issues}, issue_meta, min_pipeline_length) do
+    if valid_chain_start?(ast) do
+      {ast, {min_pipeline_length, issues}}
+    else
+      case length_as_pipeline(args) + 1 do
+        potential_pipeline_length when potential_pipeline_length >= min_pipeline_length ->
+          {ast, {min_pipeline_length, issues ++ [issue_for(issue_meta, meta[:line], "something")]}}
```
Stratus3D, line 65: What should the value of the `trigger` parameter be?

lib/credo/check/readability/nested_function_calls.ex
```diff
@@ -0,0 +1,126 @@
...
+
+  # Elixir <= 1.8.0
+  # '__#{val}__' are compiled to String.to_charlist("__#{val}__")
+  # we want to consider these charlists a valid pipe chain start
+  defp valid_chain_start?({{:., _, [String, :to_charlist]}, _, [{:<<>>, _, _}]}), do: true
+
+  # Elixir >= 1.8.0
+  # '__#{val}__' are compiled to String.to_charlist("__#{val}__")
+  # we want to consider these charlists a valid pipe chain start
+  defp valid_chain_start?({{:., _, [List, :to_charlist]}, _, [[_ | _]]}), do: true
```
Stratus3D, line 120: I'm not sure if these two charlist clauses are needed or not. I copied them from `Credo.Check.Refactor.PipeChainStart`.

lib/credo/check/readability/nested_function_calls.ex
```diff
@@ -0,0 +1,126 @@
...
+
+  # Kernel.to_string is invoked for string interpolation e.g. "string #{variable}"
+  defp valid_chain_start?({{:., _, [Kernel, :to_string]}, _, _}), do: true
```
Stratus3D, line 123: This one is needed as it does actually handle the `"string #{variable}"` case.

Stratus3D:
`Reproducing Test-Case Detector (experimental)` has failed and I am not sure why. Any suggestions?

lib/credo/check/readability/nested_function_calls.ex
rrrene: @Stratus3D The trigger is the literal string that "triggered" the issue.

In this example:

```
Enum.shuffle(Enum.uniq([1,2,3,3]))
```

the trigger should probably be `"Enum.uniq"` :+1:

rrrene:
> `Reproducing Test-Case Detector (experimental)` has failed and I am not sure why. Any suggestions?

It is a pipeline that does not matter for new checks :+1:

lib/credo/check/readability/nested_function_calls.ex
rrrene: This comment does irritate me a bit as it sounds like you made copy-pasta and are asking me if you did it correctly? Maybe I'm misunderstanding the question?

lib/credo/check/readability/nested_function_calls.ex
rrrene: I would assume that by default PRs should not contain unneeded functions. Why are you pointing this out?

lib/credo/check/readability/nested_function_calls.ex
Stratus3D: PR updated! I used a function from the `Credo.Code.Name` module for this.

lib/credo/check/readability/nested_function_calls.ex
Stratus3D: Yes, I did copy-and-paste this code from the `Credo.Check.Refactor.PipeChainStart` module. I have removed the unnecessary clauses from the function and added some additional tests to validate the remaining code is sufficient.

lib/credo/check/readability/nested_function_calls.ex
Stratus3D: It does now.

Stratus3D:
I'm not sure why the Github workflows haven't run, but I have updated the PR and as far as I know everything is good to merge.

PR #935, opened 2022-01-05
Create CondFinalCondition readability check

Stratus3D:
This PR adds a check named `Credo.Check.Readability.CondCatchallTrue` that checks for `cond`s that end in a catchall clause that use a literal value other than `true`. I created this check to catch violations of the rule about catchall `cond` clauses specified in the Christopher Adam's style guide (https://github.com/christopheradams/elixir_style_guide#true-as-last-condition). See the tests for examples.

Context:

* https://github.com/christopheradams/elixir_style_guide/issues/143#issuecomment-394191496

### DIFF
--- a/lib/credo/check/readability/cond_catchall_true.ex
+++ b/lib/credo/check/readability/cond_catchall_true.ex
@@ -1,0 +1,78 @@
+defmodule Credo.Check.Readability.CondCatchallTrue do
+  use Credo.Check,
+    explanations: [
+      check: """
+      If a cond expresion ends in an "always true" statement. That last
+      statement should be simply `true`. Other literal truthy values (such as
+      `:else`, `:always`, etc... aren't allowed.
+
+      Example:
+
+          cond do
+            x == y -> 0
+            x > y -> 0
+            :else -> 1
+          end
+
+          # should be written as
+
+          cond do
+            x == y -> 0
+            x > y -> 0
+            true -> 1
+          end
+      """
+    ]
+
+  @doc false
+  @impl true
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:cond, meta, arguments} = ast, issues, issue_meta) do
+    conditions =
+      arguments
+      |> Credo.Code.Block.do_block_for!()
+      |> List.wrap()
+
+    if conditions
+      |> List.last()
+      |> catchall_other_than_true?() do
+      {ast, issues ++ [issue_for(issue_meta, meta[:line], :cond)]}
+    else
+      {ast, issues}
+    end
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp catchall_other_than_true?({:->, _meta, [[true], _args]}), do: false
+  # Integer literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, ['{', _args]}), do: true
+  # Binary literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[binary], _args]}) when is_binary(binary), do: true
+  # List literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[list], _args]}) when is_list(list), do: true
+  # Map literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[{:%{}, _meta2, []}], _args]}), do: true
+  # Tuple literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[{:{}, _meta2, _values}], _args]}), do: true
+  # Atom literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[name], _args]}) when is_atom(name), do: true
+  defp catchall_other_than_true?(_), do: false
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue(
+      issue_meta,
+      message:
+      "Cond statements that end with an \"always true\" condition should use `true` instead of some other literal.",
+      trigger: trigger,
+      line_no: line_no
+    )
+  end
+end
--- a/lib/credo/check/readability/cond_final_condition.ex
+++ b/lib/credo/check/readability/cond_final_condition.ex
@@ -1,5 +1,8 @@
 defmodule Credo.Check.Readability.CondFinalCondition do
   use Credo.Check,
+    # Default to the value specified here:
+    # https://github.com/christopheradams/elixir_style_guide#true-as-last-condition
+    param_defaults: [final_condition_value: true],
     explanations: [
       check: """
       If a cond expresion ends in an "always true" statement. That last
@@ -21,7 +24,10 @@ defmodule Credo.Check.Readability.CondFinalCondition do
             x > y -> 0
             true -> 1
           end
-      """
+      """,
+      params: [
+        final_condition_value: "Set the expected value for the final condition"
+      ]
     ]
 
   @doc false
@@ -29,10 +35,12 @@ defmodule Credo.Check.Readability.CondFinalCondition do
   def run(%SourceFile{} = source_file, params) do
     issue_meta = IssueMeta.for(source_file, params)
 
-    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+    final_condition_value = Params.get(params, :final_condition_value, __MODULE__)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta, final_condition_value))
   end
 
-  defp traverse({:cond, meta, arguments} = ast, issues, issue_meta) do
+  defp traverse({:cond, meta, arguments} = ast, issues, issue_meta, final_condition_value) do
     conditions =
       arguments
       |> Credo.Code.Block.do_block_for!()
@@ -40,31 +48,31 @@ defmodule Credo.Check.Readability.CondFinalCondition do
 
     if conditions
       |> List.last()
-      |> catchall_other_than_true?() do
+      |> catchall_other_than_value?(final_condition_value) do
       {ast, issues ++ [issue_for(issue_meta, meta[:line], :cond)]}
     else
       {ast, issues}
     end
   end
 
-  defp traverse(ast, issues, _issue_meta) do
+  defp traverse(ast, issues, _issue_meta, _min_pipeline_length) do
     {ast, issues}
   end
 
-  defp catchall_other_than_true?({:->, _meta, [[true], _args]}), do: false
+  defp catchall_other_than_value?({:->, _meta, [[value], _args]}, value), do: false
   # Integer literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, ['{', _args]}), do: true
+  defp catchall_other_than_value?({:->, _meta, ['{', _args]}, _value), do: true
   # Binary literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, [[binary], _args]}) when is_binary(binary), do: true
+  defp catchall_other_than_value?({:->, _meta, [[binary], _args]}, _value) when is_binary(binary), do: true
   # List literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, [[list], _args]}) when is_list(list), do: true
+  defp catchall_other_than_value?({:->, _meta, [[list], _args]}, _value) when is_list(list), do: true
   # Map literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, [[{:%{}, _meta2, []}], _args]}), do: true
+  defp catchall_other_than_value?({:->, _meta, [[{:%{}, _meta2, []}], _args]}, _value), do: true
   # Tuple literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, [[{:{}, _meta2, _values}], _args]}), do: true
+  defp catchall_other_than_value?({:->, _meta, [[{:{}, _meta2, _values}], _args]}, _value), do: true
   # Atom literal catch-all clause
-  defp catchall_other_than_true?({:->, _meta, [[name], _args]}) when is_atom(name), do: true
-  defp catchall_other_than_true?(_), do: false
+  defp catchall_other_than_value?({:->, _meta, [[name], _args]}, _value) when is_atom(name), do: true
+  defp catchall_other_than_value?(_clause, _value), do: false
 
   defp issue_for(issue_meta, line_no, trigger) do
     format_issue(
--- a/lib/credo/check/readability/cond_final_condition.ex
+++ b/lib/credo/check/readability/cond_final_condition.ex
@@ -53,8 +53,10 @@ defmodule Credo.Check.Readability.CondFinalCondition do
       |> List.last()
 
     case is_catchall_clause_with_invalid_value?(last_cond_clause, config_catchall_value) do
-      {true, other_literal} ->
-        {ast, issues ++ [issue_for(issue_meta, config_catchall_value, meta[:line], other_literal)]}
+      {true, ast} ->
+        expression = Macro.to_string(ast)
+        {ast, issues ++ [issue_for(issue_meta, config_catchall_value, meta[:line], expression)]}
+
       _ ->
         {ast, issues}
     end
@@ -66,7 +68,10 @@ defmodule Credo.Check.Readability.CondFinalCondition do
 
   defp is_catchall_clause_with_invalid_value?({:->, _meta, [[value], _args]}, value), do: false
   # Integer literal catch-all clause
-  defp is_catchall_clause_with_invalid_value?({:->, _meta, ['{', _args]}, _value), do: {true, :integer}
+  defp is_catchall_clause_with_invalid_value?({:->, _meta, [[integer], _args]}, _value)
+       when is_integer(integer),
+       do: {true, integer}
+
   # Binary literal catch-all clause
   defp is_catchall_clause_with_invalid_value?({:->, _meta, [[binary], _args]}, _value)
        when is_binary(binary),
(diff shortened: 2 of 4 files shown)


lib/credo/check/readability/cond_catchall_true.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Credo.Check.Readability.CondCatchallTrue do
```
Stratus3D, line 1: Open to suggestions on what to name this check. I had a hard time picking a name and I'm not sure this name is best.

lib/credo/check/readability/cond_catchall_true.ex
```diff
@@ -0,0 +1,78 @@
...
+  defp catchall_other_than_true?({:->, _meta, [[binary], _args]}) when is_binary(binary), do: true
+  # List literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[list], _args]}) when is_list(list), do: true
+  # Map literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[{:%{}, _meta2, []}], _args]}), do: true
+  # Tuple literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[{:{}, _meta2, _values}], _args]}), do: true
+  # Atom literal catch-all clause
+  defp catchall_other_than_true?({:->, _meta, [[name], _args]}) when is_atom(name), do: true
+  defp catchall_other_than_true?(_), do: false
```
Stratus3D, line 67: Returns true of the clause condition is a catchall that isn't the literal value `true`.

rrrene:
Great :+1: 

Since the Elixir website & documentation call this the "final condition", we should name this `CondFinalCondition`.

Also, I do think the `true` should be a config parameter for this check. I can't think of a reason why we would exclude people using `:else` from the ability to check for it (although `true` should be the default to match the linked styleguide). :+1:

Stratus3D:
> Great 👍
> 
> Since the Elixir website & documentation call this the "final condition", we should name this `CondFinalCondition`.
> 
> Also, I do think the `true` should be a config parameter for this check. I can't think of a reason why we would exclude people using `:else` from the ability to check for it (although `true` should be the default to match the linked styleguide). 👍

PR updated!

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition)
+    |> assert_issue()
+  end
+
+  test "it should report conds that with a last condition that is binary literal" do
```
rkledesma, line 69: suggestion```
test "it should report conds with a last condition that is binary literal" do
```

File: test/credo/check/readability/cond_final_condition_test.exs [hunk 4]
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition)
+    |> assert_issue()
+  end
+
+  test "it should report conds that with a last condition that is integer literal" do
```
rkledesma, line 91:
```suggestion
  test "it should report conds with a last condition that is integer literal" do
```

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition)
+    |> assert_issue()
+  end
+
+  test "it should report conds that with a last condition that is list literal" do
```
rkledesma, line 113:
```suggestion
  test "it should report conds with a last condition that is a list literal" do
```

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition)
+    |> assert_issue()
+  end
+
+  test "it should report conds that with a last condition that is tuple literal" do
```
rkledesma, line 135:
```suggestion
  test "it should report conds with a last condition that is a tuple literal" do
```

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition)
+    |> assert_issue()
+  end
+
+  test "it should report conds that with a last condition that is map literal" do
```
rkledesma, line 157:
```suggestion
  test "it should report conds with a last condition that is a map literal" do
```

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
...
+        end
+      end
+    end
+    """
+    |> to_source_file()
+    |> run_check(CondFinalCondition, final_condition_value: :else)
+    |> assert_issue()
+  end
+
+  test "it should NOT report conds that with a last condition that match config" do
```
rkledesma, line 201:
```suggestion
  test "it should NOT report conds with a last condition that match the config" do
```

Stratus3D:
@rrrene this is ready for review again!

Stratus3D:
@rrrene any chance you can take a look at this PR again?

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  defp traverse({:cond, meta, arguments} = ast, issues, issue_meta, final_condition_value) do
+    conditions =
+      arguments
+      |> Credo.Code.Block.do_block_for!()
+      |> List.wrap()
+
+    if conditions
+      |> List.last()
+      |> catchall_other_than_value?(final_condition_value) do
```
rrrene, line 51: Please extract this condition into a variable.

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+          # should be written as
+
+          cond do
+            x == y -> 0
+            x > y -> 0
+            true -> 1
+          end
+      """,
+      params: [
+        final_condition_value: "Set the expected value for the final condition"
```
rrrene, line 29: Please rename this to `value`, since we don't repeat the check's name in param names for any other check.

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+  defp catchall_other_than_value?({:->, _meta, [[{:{}, _meta2, _values}], _args]}, _value), do: true
+  # Atom literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, [[name], _args]}, _value) when is_atom(name), do: true
+  defp catchall_other_than_value?(_clause, _value), do: false
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue(
+      issue_meta,
+      message:
+      "Cond statements that end with an \"always true\" condition should use `true` instead of some other literal.",
```
rrrene, line 81: The message should not specify `true` as the value if another value has been set.

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+  defp catchall_other_than_value?({:->, _meta, [[binary], _args]}, _value) when is_binary(binary), do: true
+  # List literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, [[list], _args]}, _value) when is_list(list), do: true
+  # Map literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, [[{:%{}, _meta2, []}], _args]}, _value), do: true
+  # Tuple literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, [[{:{}, _meta2, _values}], _args]}, _value), do: true
+  # Atom literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, [[name], _args]}, _value) when is_atom(name), do: true
+  defp catchall_other_than_value?(_clause, _value), do: false
```
rrrene, line 75: If the clauses "above" this one are "catch alls" for all cases, what are examples where we need this clause?

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+    else
+      {ast, issues}
+    end
+  end
+
+  defp traverse(ast, issues, _issue_meta, _min_pipeline_length) do
+    {ast, issues}
+  end
+
+  defp catchall_other_than_value?({:->, _meta, [[value], _args]}, value), do: false
```
rrrene, line 62: Please rename this function to something that gives a hint at what it does.

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+  # Integer literal catch-all clause
+  defp catchall_other_than_value?({:->, _meta, ['{', _args]}, _value), do: true
```
rrrene, line 64: Why does this always return `true`, no matter what I specified in the config?

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
+defmodule Credo.Check.Readability.CondFinalConditionTest do
+  use Credo.Test.Case
+
+  alias Credo.Check.Readability.CondFinalCondition
+
+  test "it should NOT report conds with a last condition of true" do
```
rrrene, line 6: Could you please group the tests which are not raising issues at the top of the test file? For examples, look at any standard check.

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
+defmodule Credo.Check.Readability.CondFinalCondition do
+  use Credo.Check,
```
rrrene, line 2: Please add

```
    tags: [:controversial],
```

here as this will not be a standard check since it is implementing a community style guide which not everybody agrees upon.

test/credo/check/readability/cond_final_condition_test.exs
```diff
@@ -0,0 +1,222 @@
+defmodule Credo.Check.Readability.CondFinalConditionTest do
+  use Credo.Test.Case
+
+  alias Credo.Check.Readability.CondFinalCondition
```
rrrene, line 4: Please use a module attribute for the check instead of an alias (see any standard check's test file for an example).

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+          end
+
+          # should be written as
+
+          cond do
+            x == y -> 0
+            x > y -> 0
+            true -> 1
+          end
+      """,
```
rrrene, line 27: If this is a Readability check, then this needs the readability disclaimer at the end of the explanation that all other standard readability checks have.

rrrene (changes requested):
Apologies, I'm having the first cold in 2.5 years and am no longer used to that *%&/!.

Did you use a formatter on the code?

lib/credo/check/readability/cond_final_condition.ex
```diff
@@ -0,0 +1,86 @@
...
+  defp traverse({:cond, meta, arguments} = ast, issues, issue_meta, final_condition_value) do
+    conditions =
+      arguments
+      |> Credo.Code.Block.do_block_for!()
+      |> List.wrap()
+
+    if conditions
+      |> List.last()
+      |> catchall_other_than_value?(final_condition_value) do
+      {ast, issues ++ [issue_for(issue_meta, meta[:line], :cond)]}
```
rrrene, line 52: Why is the `trigger` for the issue always "cond"? The trigger is the word in the code that "triggers" the issue and that should be the "wrong" final condition, e.g. using `:else` when using `true` is expected.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: This is for a non catch-all clause. A contrived example:

```
 cond do
   n > 5 ->
     "greater than 5"
   n >= 0 ->
     "positive"
   n < 0 ->
     "negative"
end
```

The last clause is not a catchall and is not composed of a single literal. I will add a comment for this final clause. Let me know if you think I should rename this function.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: This function returns false when the `value` variable bound by the match of the first argument (`{:->, _meta, [[value], _args]}`) matches the second argument `value`. The second argument passed to this function is value specified in the config.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated. Let me know if the new name is acceptable.

test/credo/check/readability/cond_final_condition_test.exs
Stratus3D: PR updated.

test/credo/check/readability/cond_final_condition_test.exs
Stratus3D: PR updated.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated. The trigger is now the literal catchall value, or in the case of maps, tuples, and integers, the atom `:map`, `:tuple` or `:integer`. Let me know if this is acceptable.

lib/credo/check/readability/cond_final_condition.ex
Stratus3D: PR updated.

Stratus3D:
> Apologies, I'm having the first cold in 2.5 years and am no longer used to that *%&/!.
> 
> Did you use a formatter on the code?

Sorry to hear you have been sick. Thanks for the review and thank you for all your work on credo!

I forgot to run the formatter when I opened the PR. I ran it and pushed the reformatted code. Everything is ready to review again.

lib/credo/check/readability/cond_final_condition.ex
rrrene: As I mentioned before

> The trigger is the word in the code that "triggers" the issue

which means the actual string that should get squiggly lines in your editor.

Stratus3D: I'm sorry about that. I don't know how to do that but I'll see if I can figure it out. Are there any other checks that I could refer to that do this?

Stratus3D: PR updated. I have added assertions to the tests for this check to verify that the clause expression that triggers issue is what is set for the issue's `trigger` parameter.

Stratus3D:
I have updated the PR again based on your reply to one of my comments about the issue trigger value. The trigger is now set to the string of code that triggered the issue. The PR is ready to review again.

Stratus3D:
Any chance this can get merged soon?

rrrene:
> I have added assertions to the tests for this check to verify that the clause expression that triggers issue is what is set for the issue's trigger parameter.

This is good in principle, but it does not work if I modify your tests to something that might very well happen in real life (a dangling space somewhere):

```
*..................................................................................................................................................................................................................................................................................................................................

  1) test it should report conds with a last condition that is a list literal (Credo.Check.Readability.CondFinalConditionTest)
     test/credo/check/readability/cond_final_condition_test.exs:141
     Assertion with == failed
     code:  assert "[:else ]" == trigger
     left:  "[:else ]"
     right: "[:else]"
     stacktrace:
       (credo 1.6.2) lib/credo/test/assertions.ex:30: Credo.Test.Assertions.assert_issue/2
       test/credo/check/readability/cond_final_condition_test.exs:160: (test)

.

  2) test it should report conds with a last condition that is a tuple literal (Credo.Check.Readability.CondFinalConditionTest)
     test/credo/check/readability/cond_final_condition_test.exs:165
     Assertion with == failed
     code:  assert "{ :else }" == trigger
     left:  "{ :else }"
     right: "{:else}"
     stacktrace:
       (credo 1.6.2) lib/credo/test/assertions.ex:30: Credo.Test.Assertions.assert_issue/2
       test/credo/check/readability/cond_final_condition_test.exs:184: (test)

.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

Stratus3D:
@rrrene I copied what was done in the existing checks. I used `Macro.to_string(ast)` to convert the AST back into the original expression and passed that into the issue as the trigger value.

Can you provide a suggestion on how to improve this? Happy to change things but not sure how.

Stratus3D:
@rrrene, I'd like to get this merged, can you show me how to address this in credo? I'm not sure I can get the cond clause with whitespace in credo.

> This is good in principle, but it does not work if I modify your tests to something that might very well happen in real life (a dangling space somewhere):

> Can you provide a suggestion on how to improve this? Happy to change things but not sure how.

Thanks!

Stratus3D:
Re-opening this PR as I closed it in error. @rrrene happy to make any changes to get this merged!

Stratus3D:
@rrrene you've merged several of my checks but not this one. If you can provide any info as to why this hasn't been accepted yet I will be happy to make any necessary changes (or close it permanently if that is needed).

rrrene:
> If you can provide any info as to why this hasn't been accepted yet

Hi, sorry for being a slow responder. This check won't get merged as long as it assumes that people write code the same way that `Macro.to_string/1` translates ASTs. This is just an unrealistic expectation.

I do not have an "obvious" solution for your specific problem here (a.k.a. I'm not trolling you).

rrrene:
I am closing this for age/inactivity. 

If you find a way to make the final changes, please feel free to re-open this issue at your discretion.

Stratus3D:
Hi @rrrene , sorry for the late reply here.

> Hi, sorry for being a slow responder. This check won't get merged as long as it assumes that people write code the same way that Macro.to_string/1 translates ASTs. This is just an unrealistic expectation.

Are you referring to this line? https://github.com/rrrene/credo/pull/935/files#diff-cbac4e2c2b102d2eba670658d03634fb4b21022007e98f39f5b7bccd674e52daR57 The `Credo.Check.Warning.ExpensiveEmptyEnumCheck` already uses `Macro.to_string/1` to translate AST into code for the error (see https://github.com/rrrene/credo/blob/master/lib/credo/check/warning/expensive_empty_enum_check.ex#L66).

Is there a reason this is problematic? I understand the code may not be formatted exactly as the user typed it in, but I'm not sure this will be a big problem as we do output the line number as well, so the user should be able to locate the problem without issue. Remember, this AST in this case is only ever going to be a catchall clause **literal** value. So for example, `:else`, `%{random: value}`, `[]`, `[1,2,3]`, etc... it'll never be an anything other than a literal. While it's true the user may have formatted the literal differently, it I don't think it'll be a big problem in practice. The only alternative to I see to this would be to remove the `trigger` from reported issues entirely, which probably is not desired.

PR #929, opened 2021-12-24
Check plugin module loaded fixed

juancgalvis:
Closes #928 replacing `function_exported?/3` by `Code.ensure_loaded?/1` as well as indicated in documentation:

> Note that this function does not load the module in case it is not loaded. Check Code.ensure_loaded/1 for more information.

[https://hexdocs.pm/elixir/Kernel.html#function_exported?/3](https://hexdocs.pm/elixir/Kernel.html#function_exported?/3)

PR #937, opened 2022-01-10
InitializePlugins: verify that callback init/1 of the plugin module i…

redmaner:
In addition to PR #929, the `Code.ensure_loaded?/1` function only makes sure that the plugin is loaded but does check if the plugin module also exports the required callback function `init/1`. This PR adds an extra check to verify that the the required callback function `init/1` is exported in the plugin module. If it is not exported we can safely assume the module doesn't implement the required behavior. 

I have tested the commit with one of our own applications that uses `CredoSonarqube` as a plugin, which works as expected. I also tested it with another module that doesn't have an `init/1` function, and with this false plugin the error is raised properly.

### DIFF
--- a/lib/credo/execution/task/initialize_plugins.ex
+++ b/lib/credo/execution/task/initialize_plugins.ex
@@ -13,13 +13,21 @@ defmodule Credo.Execution.Task.InitializePlugins do
 
   defp init_plugin(exec, {mod, _params}) do
     module_loaded? = Code.ensure_loaded?(mod)
+    is_plugin? = function_exported?(mod, :init, 1)
 
     if module_loaded? do
-      exec
-      |> Execution.set_initializing_plugin(mod)
-      |> mod.init()
-      |> Execution.ensure_execution_struct("#{mod}.init/1")
-      |> Execution.set_initializing_plugin(nil)
+      if is_plugin? do
+        exec
+        |> Execution.set_initializing_plugin(mod)
+        |> mod.init()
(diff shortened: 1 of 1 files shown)


lib/credo/execution/task/initialize_plugins.ex
```diff
@@ -12,19 +12,19 @@ defmodule Credo.Execution.Task.InitializePlugins do
...
     else
-      Execution.halt(
-        exec,
-        "Plugin module `#{Credo.Code.Module.name(mod)}` is not available and could not be initialized."
-      )
+      false ->
+        Execution.halt(
+          exec,
+          "Plugin module `#{Credo.Code.Module.name(mod)}` is not available and could not be initialized or is not a valid credo plugin."
+        )
```
rrrene, line 27: Great 👍  We need a nested `if` here so we can give a concise error message. "It is this or this" is not a great user experience.

redmaner: Thanks for your review, good point! I've updated the PR to include separation of error messages for a better user experience

rrrene: Hi, please update the PR according to my suggestion. You forced and unnecessary usage of `with` makes the code less readable with every iteration.

redmaner: Done

rrrene:
@redmaner Thx! :+1:

PR #939, opened 2022-01-13
fix: sanitize map before converting to json

woylie:
This fixes an issue that can occur when the credo config contains values that cannot be serialized by Jason.

To reproduce the issue:

1. `mix credo gen.config`
2. change the `ModuleDoc` check config to:

```elixir
          {Credo.Check.Readability.ModuleDoc,
           [
             ignore_names: [
               ~r/(\.\w+Controller|\.Endpoint|\.Repo|\.Router|\.\w+Socket|\.\w+View|\.\w+Components|\.\w+Live\.\w+)$/
             ]
           ]},
```

3. run `mix credo info --format json --verbose`

This will result in:

```
** (Protocol.UndefinedError) protocol Jason.Encoder not implemented for ~r/(\.\w+Controller|\.Endpoint|\.Repo|\.Router|\.\w+Socket|\.\w+View|\.\w+Components|\.\w+Live\.\w+)$/ of type Regex (a struct), Jason.Encoder protocol must always be explicitly implemented.

If you own the struct, you can derive the implementation specifying which fields should be encoded to JSON:

    @derive {Jason.Encoder, only: [....]}
    defstruct ...

It is also possible to encode all fields, although this should be used carefully to avoid accidentally leaking private information when new fields are added:

    @derive Jason.Encoder
    defstruct ...

Finally, if you don't own the struct you want to encode to JSON, you may use Protocol.derive/3 placed outside of any module:

    Protocol.derive(Jason.Encoder, NameOfTheStruct, only: [...])
    Protocol.derive(Jason.Encoder, NameOfTheStruct)

    lib/jason.ex:150: Jason.encode!/2
    lib/credo/cli/output/formatter/json.ex:15: Credo.CLI.Output.Formatter.JSON.print_map/1
    lib/credo/cli/command/info/info_command.ex:32: Credo.CLI.Command.Info.InfoCommand.PrintInfo.call/2
    lib/credo/execution/task.ex:133: Credo.Execution.Task.do_run/3
    (elixir 1.13.0) lib/enum.ex:2396: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/credo/execution/task.ex:133: Credo.Execution.Task.do_run/3
    (elixir 1.13.0) lib/enum.ex:2396: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/credo.ex:30: Credo.run/1
```

Noticed because this happened with a certain VS code plugin.

### DIFF
--- a/lib/credo/cli/output/formatter/json.ex
+++ b/lib/credo/cli/output/formatter/json.ex
@@ -26,7 +26,7 @@ defmodule Credo.CLI.Output.Formatter.JSON do
   def prepare_for_json([]), do: []
   def prepare_for_json([h | t]), do: [prepare_for_json(h) | prepare_for_json(t)]
 
-  def prepare_for_json(%Regex{} = regex), do: "~r/#{Regex.source(regex)}/"
+  def prepare_for_json(%Regex{} = regex), do: inspect(regex)
 
   def prepare_for_json(%{} = term) do
     Enum.into(term, %{}, fn {key, value} ->
--- a/lib/credo/cli/output/formatter/json.ex
+++ b/lib/credo/cli/output/formatter/json.ex
@@ -13,34 +13,34 @@ defmodule Credo.CLI.Output.Formatter.JSON do
 
   def print_map(map) do
     map
-    |> sanitize()
+    |> prepare_for_json()
     |> Jason.encode!(pretty: true)
     |> UI.puts()
   end
 
-  def sanitize(term)
+  def prepare_for_json(term)
       when is_atom(term) or is_number(term) or is_binary(term) do
     term
   end
 
-  def sanitize([]), do: []
-  def sanitize([h | t]), do: [sanitize(h) | sanitize(t)]
+  def prepare_for_json([]), do: []
+  def prepare_for_json([h | t]), do: [prepare_for_json(h) | prepare_for_json(t)]
 
-  def sanitize(%Regex{} = regex), do: "~r/#{Regex.source(regex)}/"
+  def prepare_for_json(%Regex{} = regex), do: "~r/#{Regex.source(regex)}/"
 
-  def sanitize(%{} = term) do
+  def prepare_for_json(%{} = term) do
     Enum.into(term, %{}, fn {key, value} ->
-      {sanitize(key), sanitize(value)}
+      {prepare_for_json(key), prepare_for_json(value)}
     end)
   end
 
-  def sanitize(term) when is_tuple(term) do
+  def prepare_for_json(term) when is_tuple(term) do
     term
     |> Tuple.to_list()
-    |> sanitize()
+    |> prepare_for_json()
   end
 
-  def sanitize(term) do
+  def prepare_for_json(term) do
     inspect(term)
   end
 
--- a/lib/credo/cli/output/formatter/json.ex
+++ b/lib/credo/cli/output/formatter/json.ex
@@ -30,7 +30,7 @@ defmodule Credo.CLI.Output.Formatter.JSON do
 
   def prepare_for_json(%{} = term) do
     Enum.into(term, %{}, fn {key, value} ->
-      {prepare_for_json(key), prepare_for_json(value)}
+      {prepare_key_for_json(key), prepare_for_json(value)}
     end)
   end
 
@@ -42,6 +42,14 @@ defmodule Credo.CLI.Output.Formatter.JSON do
 
   def prepare_for_json(term) do
     inspect(term)
+  end
+
+  defp prepare_key_for_json(k) when is_atom(k) or is_binary(k) or is_number(k) do
+    k
+  end
+
+  defp prepare_key_for_json(k) do
+    inspect(k)
   end
 
   def issue_to_json(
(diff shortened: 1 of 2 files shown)


woylie:
I didn't consider structs here, I don't think they will ever occur in the config?

woylie:
Hey @rrrene,

if you have a better function name, I can change it. I'm trying to come up with something else, but... I don't know. I mean, `to sanitize` just means `säubern, hygienisch machen, keimfrei machen`, so although it is often used in the context of input validation, it's still kind of fitting. Maybe just `format`? `clean`? Or we can make it even more dramatic: `decontaminate`.

`defimpl` would implement the protocol in any application using credo. Even if the library is only used in dev or test, I wouldn't like that.

rrrene:
Your reasoning against `defimpl` is solid 👍 Did not think of that.

> it's still kind of fitting

Well, I don't think it is as there is literally nothing wrong with the inputs except their format.

What about `term_to_json`?

woylie:
> > it's still kind of fitting
> 
> Well, I don't think it is as there is literally nothing wrong with the inputs except their format.

Well, in the scope of JSON, a Regex struct is very wrong 😄 

> What about `term_to_json` or `cast_for_json`?

What comes out of there is not JSON yet, so `for` is better than `to`. Maybe `convert_for_json` or `prepare_for_json`?

rrrene:
> Well, in the scope of JSON, a Regex struct is very wrong 😄

I don't think we see eye to eye on this, but neither of our lives will be touched by the outcome, so ...

> What comes out of there is not JSON yet, so `for` is better than `to`. Maybe `convert_for_json` or `prepare_for_json`?

If it is this important to you, you can call it Susan if you like 👍

lib/credo/cli/output/formatter/json.ex
```diff
@@ -12,7 +12,36 @@ defmodule Credo.CLI.Output.Formatter.JSON do
...
+
+  def sanitize(term)
+      when is_atom(term) or is_number(term) or is_binary(term) do
+    term
+  end
+
+  def sanitize([]), do: []
+  def sanitize([h | t]), do: [sanitize(h) | sanitize(t)]
+
+  def sanitize(%Regex{} = regex), do: "~r/#{Regex.source(regex)}/"
```
rrrene, line 29: Why aren't we using `inspect/1` here? Conducting your own string operation magic just to arrive at the same result seems unnecessary.

lib/credo/cli/output/formatter/json.ex
```diff
@@ -12,7 +12,36 @@ defmodule Credo.CLI.Output.Formatter.JSON do
...
+
+  def sanitize(%{} = term) do
+    Enum.into(term, %{}, fn {key, value} ->
+      {sanitize(key), sanitize(value)}
+    end)
+  end
+
```
rrrene, line 36: If (!) this is done this way to catch incompatible keys then we are missing a test for that.

lib/credo/cli/output/formatter/json.ex
```diff
@@ -12,7 +12,36 @@ defmodule Credo.CLI.Output.Formatter.JSON do
...
+  def sanitize(term) when is_tuple(term) do
+    term
+    |> Tuple.to_list()
+    |> sanitize()
+  end
+
+  def sanitize(term) do
```
rrrene, line 43: I think you are missing a case for atom keys, unless you want

```elixir
%{a: "b"}
```

to result in

```json
{":a": "b"}
```

test/credo/cli/output/formatter/json_test.exs
```diff
@@ -0,0 +1,29 @@
+defmodule Credo.CLI.Output.Formatter.JsonTest do
+  use Credo.Test.Case
+
+  alias Credo.CLI.Output.Formatter.JSON
+
+  test "print_map/1 does not raise when map contains a regex" do
+    assert JSON.print_map(%{"option" => ~r/foo/}) == nil
```
rrrene, line 7: Why is this asserting the return value of `UI.puts/1`?

lib/credo/cli/output/formatter/json.ex
woylie: There's a match on atoms in line 21.

test/credo/cli/output/formatter/json_test.exs
woylie: Before the change, this line was raising an exception, after the change it's fine. If you mean the `== nil`, I can remove that, if you prefer.

woylie: But then it would have to be a `refute`, which looks a bit weird.

woylie: Or I can remove the `assert ` _and_ the ` == nil`. Let me know what you prefer.

rrrene: What I am saying is: you are not testing the implementation of `JSON.print_map/1` here. You are testing the result of `UI.puts/1`.

If `UI.puts/1` would be refactored to return `:ok`, this test would break for no benefit, since it is not testing the function it says it is testing.

woylie: Makes sense. Updated to remove the assertion.

rrrene:
@woylie Thx! :+1:

PR #945, opened 2022-02-02
Add null-ls diagnostic to README

jc00ke:
image

I'm seriously impressed with `null-ls` and it's `credo` integration, thought it was worth documenting for others. Thanks!


rrrene:
@jc00ke That's a great addition! Thank you!

PR #955, opened 2022-03-10
bugfix: diff fixed issue return correctly

calvin-kargo:
bugfix: diff fixed issue return not correct fixed issue, because previous issue has different directory & filename than old issues, thus `--` doesn't work


rrrene:
Hi, sorry for the late response. 😭 

This PR changes core functionality of `mix credo diff` but does neither mention, document or test the faulty behaviour it is fixing. 🤷‍♂️ 

If the bug is still present today, please re-open this PR and provide a test that reproduces the original problem and makes sure that it is fixed :+1:

PR #960, opened 2022-03-29
Make Credo.Check.Refactor.NegatedIsNil work for multi clause guard

### LINKED ISSUE #958: Credo.Check.Refactor.NegatedIsNil ignores issues when there is more than one guard
### Environment

* Credo version (`mix credo -v`): 1.6.4
* Erlang/Elixir version (`elixir -v`): 1.13.0
* Operating system: osx monterrey

Run Credo.Check.Refactor.NegatedIsNil on a module with function
```
def foo(%{item_map: item_map, item_type: item_type})
    when not is_nil(item_type) and is_map(item_map) do
  ...
end
```

### Expected outcome

Credo should find an issue from `when not is_nil(item_type)`

### Actual outcome

No issue is found.  However, if the second guard is removed, then the issue will be properly found and reported by Credo.

ak3nji:
Fixes #958
 
Make `Credo.Check.Refactor.NegatedIsNil` work for multi clause guard like:

```elixir
def foo(%{item_map: item_map, item_type: item_type})
    when not is_nil(item_type) and is_map(item_map) do
  ...
end
```

### DIFF
--- a/lib/credo/check/refactor/negated_is_nil.ex
+++ b/lib/credo/check/refactor/negated_is_nil.ex
@@ -63,5 +63,12 @@ defmodule Credo.Check.Refactor.NegatedIsNil do
     {ast, [issue | issues]}
   end
 
+  defp traverse({:when, meta, [fun, {_ , _, [first_op | second_op]}]} = ast, issues, issue_meta) do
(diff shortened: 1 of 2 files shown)


rrrene:
@ak3nji Thx!

PR #961, opened 2022-04-02
Ignore Phoenix.LiveView modules by default

btkostner:
This change will ignore module docs for phoenix live views by default. According to [this code sample](https://github.com/phoenixframework/phoenix_live_view/blob/master/lib/phoenix_live_view/router.ex#L415) live views should end with a `Live` suffix like so `MyApp.ExampleLive`.

### DIFF
--- a/lib/credo/check/readability/module_doc.ex
+++ b/lib/credo/check/readability/module_doc.ex
@@ -2,7 +2,7 @@ defmodule Credo.Check.Readability.ModuleDoc do
   use Credo.Check,
     param_defaults: [
       ignore_names: [
-        ~r/(\.\w+Controller|\.Endpoint|\.Repo|\.Router|\.\w+Socket|\.\w+View)$/
+        ~r/(\.\w+Controller|\.Endpoint|\.\w+Live|\.Repo|\.Router|\.\w+Socket|\.\w+View)$/
       ]
     ],
     explanations: [


cblavier:
it looks like that none of the readability checks are run against live_views? is this intended?

rrrene:
@cblavier What do you mean?

@btkostner I think this does not work as intended. When generating a new Phoenix LiveView app and running `mix phx.gen.live Accounts User users name:string` you get modules called `DemoWeb.UserLive.Index` and `DemoWeb.UserLive.Show`, which your regex do not cover.

Could you re-iterate on your solution?

btkostner:
Hmmm that must have changed recently (since the PR). I'll work on opening another PR to update the regex.

rrrene:
Wouldn't we have to cover both `DemoWeb.PageLive` and `DemoWeb.UserLive.Show`?

> Hmmm that must have changed recently (since the PR).

I don't think the case where one generates via `phx.gen.live` changed since April 😉

cblavier:
> @cblavier What do you mean?

What I mean is that I expected this PR to disable `Credo.Check.Readability.ModuleDoc` for LiveViews. Still, I tested with `Credo.Check.Readability.AliasOrder`, and it also seems to be disabled in LiveViews.

I also checked:
- that other credo rules (for instance `Refactor`) are still run against my LiveViews
- that `Readability` checks are still run against other modules

rrrene:
@cblavier I can not reproduce this. :(

Generating a new Liveview enabled Phoenix app, adding an "unordered alias" and running `mix credo --strict`, I get the expected line:

```
┃ [R] ↘ The alias `DemoWeb.PageController` is not alphabetically ordered among its group. [Credo.Check.Readability.AliasOrder]
┃       /home/rrrene/projects/demo/lib/demo_web/live/user_live/index.ex:4:9 #(DemoWeb.UserLive.Index)
```

Did you run Credo in strict mode?

cblavier:
Nevermind, my test was flawed. 
Following aliases will never trigger a `Credo.Check.Readability.AliasOrder` warning.

```
alias PhxLiveStorybook.Variation
alias PhxLiveStorybook.{EntryNotFound, EntryTabNotFound}
```

But this code would.

```
alias PhxLiveStorybook.Variation
alias PhxLiveStorybook.EntryNotFound
alias PhxLiveStorybook.EntryTabNotFound
```

So please forget my comment, and thanks for your time!

PR #968, opened 2022-04-28
Load application configuration in mix task

axelson:
There are some custom credo checks that depend on configuration being loaded so most developers would assume that `config/runtime.exs` is loaded. This PR ensures that the projects `config/runtime.exs` is loaded.

For example https://github.com/Artur-Sulej/excellent_migrations uses application configuration so that there is only one place to configure the rules since it can be run from its own mix task, as a migration task or via a credo check.

### DIFF
--- a/lib/mix/tasks/credo.ex
+++ b/lib/mix/tasks/credo.ex
@@ -4,6 +4,9 @@ defmodule Mix.Tasks.Credo do
   @shortdoc "Run code analysis (use `--help` for options)"
   @moduledoc @shortdoc
 
+  # Load application config because some custom checks depend on configuration
+  @requirements ["app.config"]
+
   @doc false
   def run(argv) do
     Credo.CLI.main(argv)


rrrene:
While I get the idea, I am unsure regarding its consequences.

Does this change the runtime behaviour of the credo task for existing users? Can this break their workflow in certain scenarios?

And why don't you just load the app config inside your `.credo.exs`?

Then, you are saying "some checks" and then cite a single one (I might be a bit harsh with this one as I am confronted with this meme at work all the time, i.e. claiming something being ultra common best practise and then being unable to cite five teams/projects using said method/framework/etc). 

Lastly, my main concern would be changing the UX from "all Credo config is in .credo.exs (where you can do whatever you like since it is just regular Elixir code)" to "Credo config can be anywhere" ...

rrrene:
After sleeping on it, maybe I was too harsh yesterday. If we can ensure that no existing workflows are affected, we should add this 👍

axelson:
> Does this change the runtime behaviour of the credo task for existing users? Can this break their workflow in certain scenarios?

It will not change the runtime behavior significantly since the only change is loading the user's `config/runtime.exs` so I think it is highly doubtful to break user's workflows. Although theoretically it is possible for breakage with custom credo checks.

> And why don't you just load the app config inside your .credo.exs?

This did not occur to me. After a quick check `Mix.Task.run("app.config")` does ensure that `config/runtime.exs` is loaded in time to be available for when the custom credo check is run. So I will be using that locally. But I do still think that the current behavior is surprising, I would generally expect `config/runtime.exs` to be loaded in all cases (during runtime) that `config/config.exs` is loaded. So from my point of view, I think it makes sense to load `config/runtime.exs` by default. Although perhaps that should be more of a proposal to change Elixir rather than a proposal to change Credo.

> Lastly, my main concern would be changing the UX from "all Credo config is in .credo.exs (where you can do whatever you like since it is just regular Elixir code)" to "Credo config can be anywhere"

I definitely understand this concern. And 100% for all the core credo checks the configuration should be via `.credo.exs` and not via configuration. But for custom credo checks that run in non-credo contexts as well, I think configuration via the application environment can make sense as a technical direction.

Although one hiccup is that I didn't realize that `@requirements` itself was only added in Elixir 1.11. That definitely means that this PR is not ready to go as is (as evidenced by the failing CI checks). I think there's a few ways forward to this PR:
1. Wrap the `@requirements ["app.config"]` in a `System.version` check
  a. Ugly and maybe brittle but it would work
2. `Mix.Tasks.Credo` could run `Mix.Task.run("app.config")` in code (probably behind a `System.version` check as well)
  a. Will be a bit cleaner than option 1 because it does less meta programming
3. Add a note about `app.config` to Credo's docs
  a. Probably to this page: https://hexdocs.pm/credo/config_file.html
4. Change Elixir's default behavior
  a. Although a PR seems relatively unlikely to be merged
5. Add notes to the configuration of any custom credo check that is impacted by this
6. Close this PR and let the PR itself serve as a form of documentation

I'm thinking that 1, 2 (and probably 3) feel a bit invasive given that this won't be an issue for the vast majority of Credo's user's. So I think I'll focus on 4, 5, and 6 (especially 5). Thank you for your consideration.

PR #970, opened 2022-05-04
fix: __MODULE__ multi alias issue

avinayak:


rrrene:
This looks good. Thx! 👍

Issue #956, opened 2022-03-24
Credo.Check.Design.AliasUsage doesn't warn when unaliased modules are used without function calls


pdgonzalez872:
Hi @rrrene!

First off, thanks for Credo!

### Precheck

Current master/main.

### Expected outcome

Warn when modules are not aliased. More details below.

### Actual outcome

It doesn't warn.

### More details

I wanted to note a curious behavior from the current implementation of
`AliasUsage`: It only works when modules are making function calls. If you
reference the module itself you won't get a warning. This showed up in one of
our implementation switches and I wondered why Credo didn't complain. I imagine
more folks will start to run into this as Mox (I love Mox) is used for testing
and this type of switch is more common. I wrote a test that shows the
behavior.

The real question here is: would you expect it to warn in this case? If so, I'd
be happy to try to fix this. If not, no worries, maybe some docs on
this would help clear things up. I'd be happy either way.

Here is a commit with a failing test, I pushed it up to a branch here if you'd
like to check it out:
- branch: https://github.com/pdgonzalez872/credo/commits/pg-alias-usage-01
- commit: https://github.com/pdgonzalez872/credo/commit/967a663a42e56c5612de1a0e7b8085b6dafa5f8f

Here is the commit for simplicity sake:

```
commit 967a663a42e56c5612de1a0e7b8085b6dafa5f8f
Author: Paulo Daniel Gonzalez <pdgonzalez872@gmail.com>
Date:   Thu Mar 24 10:31:18 2022 -0500

    Add failing test

diff --git a/test/credo/check/design/alias_usage_test.exs b/test/credo/check/design/alias_usage_test.exs
index 189e4ace..496b8d71 100644
--- a/test/credo/check/design/alias_usage_test.exs
+++ b/test/credo/check/design/alias_usage_test.exs
@@ -211,6 +211,23 @@ defmodule Credo.Check.Design.AliasUsageTest do
     |> assert_issue()
   end

+  test "it should report a violation even when there is no function call" do
+    """
+    defmodule CredoSampleModule do
+      def call do
+        impl().something
+      end
+
+      def impl do
+        Application.get_env(:some_third_party, :live_impl, Credo.LiveImpl)
+      end
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check)
+    |> assert_issue()
+  end
+
   test "it should report if configured to complain start at a certain depth" do
     """
     defmodule CredoSampleModule do
```

You will get a failing test when running `mix test
[...]

PR #986, opened 2022-08-03
[956] Add test and maybe a fix for modules passed as arguments

tpitale:
Related to #956 

I want to check in early to make sure I'm on the right path.

When a module is passed as an argument to a function, it should maybe be aliased?

What I'm actually trying to fix is probably when a module is passed to a macro and not aliased. I may try to update the PR if I'm on the right path.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-erlang 24.2
+erlang 24.3.4.2
 elixir 1.13.4-otp-24
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-erlang 24.3.4.2
+erlang 24.2
 elixir 1.13.4-otp-24
--- a/lib/credo/check/design/alias_usage.ex
+++ b/lib/credo/check/design/alias_usage.ex
@@ -61,6 +61,8 @@ defmodule Credo.Check.Design.AliasUsage do
 
   alias Credo.Code.Name
 
+  @keywords [:alias]
+
   @doc false
   @impl true
   def run(%SourceFile{} = source_file, params) do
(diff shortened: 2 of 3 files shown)


.tool-versions
```diff
@@ -1,2 +1,2 @@
-erlang 24.2
+erlang 24.3.4.2
```
tpitale, line 1: I can remove this.

lib/credo/check/design/alias_usage.ex
```diff
@@ -61,6 +61,8 @@ defmodule Credo.Check.Design.AliasUsage do
 
   alias Credo.Code.Name
 
+  @keywords [:alias]
```
tpitale, line 64: This could maybe be a param.

lib/credo/check/design/alias_usage.ex
```diff
@@ -169,40 +171,90 @@ defmodule Credo.Check.Design.AliasUsage do
...
+       excluded_namespaces,
+       excluded_lastnames,
+       only,
+       aliases,
+       mod_deps
+     )}
+  end
+
+  defp find_issues(
+         {fun_atom, _, [{:__aliases__, meta, mod_list}]} = ast,
```
tpitale, line 189: This was the best I think I could do when trying to find a function call 🤔 But, I'm not super-familiar with the AST.

rrrene:
This is really great. Can you introduce a param to the check that allows users to deactivate this behaviour? 

Something like `ignore_plain_module_reference` (please use a better name if you have suggestions ^^)^.

rrrene:
I am closing this for age/inactivity. 

If you find the time to make the final changes, please feel free to re-open this issue at your discretion.

rrrene:
@tpitale I closed this in error. It is now merged. Thx for your contribution! :+1:

PR #988, opened 2022-08-09
Update ModuleDoc regex for LiveView

btkostner:
This is a follow up from https://github.com/rrrene/credo/pull/961#issuecomment-1197071958. Sorry for the delay. This should cover modules like `DemoWeb.TestLive` and `DemoWeb.TestLive.Index`.


rrrene:
@btkostner Thx for following up 👍

PR #995, opened 2022-09-02
Update docs to use import Bitwise

ryvasquez:
I noticed that we're now supporting elixir 1.14, just saw one remaining doc that still using `use Bitwise`, as per deprecation notice we want to use `import Bitwise` instead of `use Bitwise`


rrrene:
Good catch! Thx!

PR #996, opened 2022-09-02
feat: Add Ignore option for Credo.Check.Readability.ModuleNames

BobrImperator:
- Adds :ignore param
  - :ignore is a list of regexes
- Adds tests for the functionality

Given
  ignore: [~r/Sample_Module$/]

Credo.Sample_Module will pass ✅ 
Credo.Another_Module.Sample_Module will fail ❌

Context:
On our project for versioning reasons, we decided to add an API version to a module name, but credo didn't have an option to configure it.

closes https://github.com/rrrene/credo-proposals/issues/83

### DIFF
--- a/lib/credo/check/readability/module_names.ex
+++ b/lib/credo/check/readability/module_names.ex
@@ -25,8 +25,7 @@ defmodule Credo.Check.Readability.ModuleNames do
       it easier to follow.
       """,
       params: [
-        ignore:
-          "List of ignored name segment patterns e.g. `[~r/Sample_Module/, \"Sample_Module\"]`"
+        ignore: "List of ignored module names and patterns e.g. `[~r/Sample_Module/, \"Credo.Sample_Module\"]`"
       ]
     ]
 
@@ -63,17 +62,11 @@ defmodule Credo.Check.Readability.ModuleNames do
   end
 
   defp issues_for_name(name, meta, issues, issue_meta, ignored_patterns) do
-    segments =
+    module_name =
       name
       |> to_string
-      |> String.split(".")
 
-    all_correct? =
-      Enum.all?(segments, fn segment ->
-         Name.pascal_case?(segment) or ignored_segment?(ignored_patterns, segment) 
-      end)
-
-    if all_correct? do
+    if Name.pascal_case?(module_name) or ignored_module?(ignored_patterns, module_name) do
       issues
     else
       [issue_for(issue_meta, meta[:line], name) | issues]
@@ -89,15 +82,14 @@ defmodule Credo.Check.Readability.ModuleNames do
     )
   end
 
-  defp ignored_segment?([], _segment), do: false
+  defp ignored_module?([], _module_name), do: false
 
-  defp ignored_segment?(ignored_patterns, segment) do
+  defp ignored_module?(ignored_patterns, module_name) do
     Enum.any?(ignored_patterns, fn
       %Regex{} = pattern ->
-        String.match?(segment, pattern)
-
-      module_name ->
-        String.equivalent?(segment, to_string(module_name))
+        String.match?(module_name, pattern)
+      name ->
+        module_name == name
     end)
   end
 end
--- a/lib/credo/check/readability/module_names.ex
+++ b/lib/credo/check/readability/module_names.ex
@@ -70,7 +70,7 @@ defmodule Credo.Check.Readability.ModuleNames do
 
     all_correct? =
       Enum.all?(segments, fn segment ->
-         allowed_segment?(allowed_patterns, segment) or Name.pascal_case?(segment)
+         Name.pascal_case?(segment) or allowed_segment?(allowed_patterns, segment) 
       end)
 
     if all_correct? do
--- a/lib/credo/check/readability/module_names.ex
+++ b/lib/credo/check/readability/module_names.ex
@@ -1,6 +1,9 @@
 defmodule Credo.Check.Readability.ModuleNames do
   use Credo.Check,
     base_priority: :high,
+    param_defaults: [
+      allow: []
+    ],
     explanations: [
       check: """
       Module names are always written in PascalCase in Elixir.
@@ -20,7 +23,11 @@ defmodule Credo.Check.Readability.ModuleNames do
       Like all `Readability` issues, this one is not a technical concern.
       But you can improve the odds of other reading and liking your code by making
       it easier to follow.
-      """
+      """,
+      params: [
+        allow:
+          "List of allowed name segment patterns e.g. `[~r/Sample_Module/, \"Sample_Module\"]`"
+      ]
     ]
 
   alias Credo.Code.Name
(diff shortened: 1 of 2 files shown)


lib/credo/check/readability/module_names.ex
```diff
@@ -29,39 +36,61 @@ defmodule Credo.Check.Readability.ModuleNames do
...
+
+                module_name ->
+                  String.equivalent?(segment, to_string(module_name))
+              end)
+
+            _ ->
+              false
+          end
+
+        is_allowed_segment? or is_pascal_case?
```
tomekowal, line 90: Not sure if it has a significant impact, but the code performs this additional work even if the name is pascal case.
If the `is_allowed_segment?` is transformed into a function, you could write
`Name.pascal_case? or is_pascal_case?(segment, allowed_patterns)`
and that would short circuit and not run the code that does Regex or any other expensive stuff.

BobrImperator: Good point 👍 
Applied the feedback

tomekowal (approved):

rrrene:
Why would you open a PR for something and then write a proposal for it 40 minutes later?

BobrImperator:
@rrrene I thought that's the process and that It'd also be nice for discoverability

lib/credo/check/readability/module_names.ex
```diff
@@ -1,6 +1,9 @@
 defmodule Credo.Check.Readability.ModuleNames do
   use Credo.Check,
     base_priority: :high,
+    param_defaults: [
+      allow: []
```
rrrene, line 5: This should be called `:ignore` since the given module names should be exempt from the rule and thus ignored by the check (see several other checks providing an `:ignore` parameter).

lib/credo/check/readability/module_names.ex
```diff
@@ -20,7 +23,11 @@ defmodule Credo.Check.Readability.ModuleNames do
       Like all `Readability` issues, this one is not a technical concern.
       But you can improve the odds of other reading and liking your code by making
       it easier to follow.
-      """
+      """,
+      params: [
+        allow:
+          "List of allowed name segment patterns e.g. `[~r/Sample_Module/, \"Sample_Module\"]`"
```
rrrene, line 29: I absolutely dislike this idea of "segments". Why would I want to define patterns for segments instead of patterns for module names to ignore?

lib/credo/check/readability/module_names.ex
```diff
@@ -29,39 +36,44 @@ defmodule Credo.Check.Readability.ModuleNames do
...
+    segments =
       name
       |> to_string
       |> String.split(".")
-      |> Enum.all?(&Name.pascal_case?/1)
 
-    if all_pascal_case? do
+    all_correct? =
+      Enum.all?(segments, fn segment ->
+         allowed_segment?(allowed_patterns, segment) or Name.pascal_case?(segment)
```
rrrene, line 73: You have not applied @tomekowal's feedback here to put `Name.pascal_case?/1` first.

lib/credo/check/readability/module_names.ex
```diff
@@ -76,4 +88,16 @@ defmodule Credo.Check.Readability.ModuleNames do
...
+
+  defp allowed_segment?([], _segment), do: false
+
+  defp allowed_segment?(allowed_patterns, segment) do
+    Enum.any?(allowed_patterns, fn
+      %Regex{} = pattern ->
+        String.match?(segment, pattern)
+
+      module_name ->
+        String.equivalent?(segment, to_string(module_name))
```
rrrene, line 100: Could please provide cases where we would need `String.equivalent/2` for the comparison?

test/credo/check/readability/module_names_test.exs
```diff
@@ -42,6 +42,46 @@ defmodule Credo.Check.Readability.ModuleNamesTest do
...
+    |> refute_issues()
+  end
+
+  test "it should not raise on multiple allowed segments" do
+    """
+    defmodule Credo.Another_Module.Sample_Module do
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check, allow: ["Another_Module", "Sample_Module"])
```
rrrene, line 81: This is a good example for why I think this "segment approach" is not good.

I want to ignore a module and need to put two patterns to do that.

test/credo/check/readability/module_names_test.exs
```diff
@@ -42,6 +42,46 @@ defmodule Credo.Check.Readability.ModuleNamesTest do
     |> refute_issues()
   end
 
+  test "it should not raise on allowed segment name" do
+    """
+    defmodule Credo.Sample_Module do
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check, allow: ["Sample_Module"])
```
rrrene, line 51: You have no test that covers your "core module".

```
"""
defmodule Sample_Module do
end
"""
|> to_source_file
|> run_check(@described_check, allow: ["Sample_Module"])
```

lib/credo/check/readability/module_names.ex
```diff
@@ -29,39 +36,44 @@ defmodule Credo.Check.Readability.ModuleNames do
...
 
       _ ->
         issues
     end
   end
 
-  defp issues_for_name(name, meta, issues, issue_meta) do
-    all_pascal_case? =
+  defp issues_for_name(name, meta, issues, issue_meta, allowed_patterns) do
+    segments =
```
rrrene, line 66: Again, why would you choose checking "segments" over just listing modules and patterns that should be ignored by the check?

rrrene (changes requested):
I'm not sure if you are serious ... anyway, please find my review attached.

lib/credo/check/readability/module_names.ex
BobrImperator: So the rule could either take a regex `~r/Some_Module/` or simply a string `"Some_Module"`.
I guess if you'd prefer to abandon checking against segments then it'd be not needed.

test/credo/check/readability/module_names_test.exs
BobrImperator: That's incorrect. It doesn't require to provide a path in chunks.

In this example it allows 2 modules.
If it were as you're saying, then the rule would require to type `allow: ["Credo", "Another_Module", "Sample_Module"]`

tomekowal: I feel there is a slight misunderstanding between you two about the goal of that PR.

@BobrImperator use case is having version number in the module name. E.g. for v1.1 some modules would have the version in the name `V1_1`, e.g. `MyApp.V1_1.SomeController`, `MyApp.V1_1.SomeControllerTest`, `MyApp.V1_1.SomeOtherController` and so on... So far, we've specified `# credo:disable-for-this-file Credo.Check.Readability.ModuleNames` for each of those. That is why he wants to specify a "segment" that is considered valid. However, he would still like to catch `My_App.V1_1.Some_Controller` because that would be wrong.

But in Credo, no rules were working on "segments" so far. Segments were just an implementation detail of this particular function. If working on segments is confusing for @rrrene, it would also be confusing to other users.

I think working on module names might be less confusing. If any module "segment" matches the pattern, it is ignored. The downside is that if we ignore `V1_1` then `My_App.V1_1.Some_Controller` will be valid (it will be less strict than the version of @BobrImperator because it will ignore the entire module name). However, it might be easier to grasp.

rrrene: @BobrImperator Tomasz already summarized the main point here: The fact that the original implementation worked on the individual parts of the fullname is just an implementation detail.

For a good user experience, we should think from a user's perspective and I would argue that most users think "I want to ignore `MyProject.Foo_Bar`.

@tomekowal While it is less strict for your example, it is also more powerful. Using Regex, you can easily allow for your exact scenario (`~r/^[^_]+\.V\d+_\d+\.[^_]+$/`). :+1:

lib/credo/check/readability/module_names.ex
```diff
@@ -76,4 +81,13 @@ defmodule Credo.Check.Readability.ModuleNames do
       line_no: line_no
     )
   end
+
+  defp ignored_module?([], _module_name), do: false
+
+  defp ignored_module?(ignored_patterns, module_name) do
+    Enum.any?(ignored_patterns, fn
+      pattern ->
+        String.match?(module_name, pattern)
```
rrrene, line 90: I probably wasn't clear on this one: I liked your idea of supporting binaries, atoms and regexes :+1: 

```
ignore: ["Credo.Code.My_Module", Credo.Code.My_Module, ~r/My_Module/]
```

This would be great for different scenarios of different complexity I outlined in the other comment.

I just wasn't convinced by the usage of `String.equivalent/2` instead of `==` (as equivalent seems to be for other use cases).

rrrene:

PR #997, opened 2022-09-02
feat: warn on dbg calls

ding-an-sich:
Implements a warning for calls to the dbg/0 or dbg/2 Kernel fns.

I used the IO.inspect check as a reference.

### DIFF
--- a/lib/credo/check/warning/dbg.ex
+++ b/lib/credo/check/warning/dbg.ex
@@ -1,5 +1,6 @@
 defmodule Credo.Check.Warning.Dbg do
   use Credo.Check,
+    id: "EX5026",
     base_priority: :high,
     elixir_version: ">= 1.14.0-dev",
     explanations: [
--- a/lib/credo/check/warning/dbg.ex
+++ b/lib/credo/check/warning/dbg.ex
@@ -1,0 +1,79 @@
+defmodule Credo.Check.Warning.Dbg do
+  use Credo.Check,
+    base_priority: :high,
+    elixir_version: ">= 1.14.0-dev",
+    explanations: [
+      check: """
+      Calls to dbg/0 and dbg/2 should mostly be used during debugging sessions.
+
+      This check warns about those calls, because they probably have been committed
+      in error.
+      """
+    ]
+
+  @call_string "dbg"
+
+  @doc false
+  @impl true
+  def run(%SourceFile{} = source_file, params) do
+    issue_meta = IssueMeta.for(source_file, params)
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse(
+         {:dbg, meta, []} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(
+         {:dbg, meta, [_single_param]} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(
+         {:dbg, meta, [_first_param, _second_param]} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(
+         {{:., _, [{:__aliases__, _, [:"Elixir", :Kernel]}, :dbg]}, meta, _args} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(
+         {{:., _, [{:__aliases__, _, [:Kernel]}, :dbg]}, meta, _args} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+
+  defp issues_for_call(meta, issues, issue_meta) do
+    [issue_for(issue_meta, meta[:line], @call_string) | issues]
+  end
+
+  defp issue_for(issue_meta, line_no, trigger) do
+    format_issue(
+      issue_meta,
+      message: "There should be no calls to dbg.",
+      trigger: trigger,
+      line_no: line_no
+    )
+  end
+end
--- a/test/credo/check/warning/dbg_test.exs
+++ b/test/credo/check/warning/dbg_test.exs
@@ -1,0 +1,184 @@
+defmodule Credo.Check.Warning.DbgTest do
+  use Credo.Test.Case
+
+  @described_check Credo.Check.Warning.Dbg
+
+  #
+  # cases NOT raising issues
+  #
+
+  test "it should NOT report expected code" do
+    """
+    defmodule CredoSampleModule do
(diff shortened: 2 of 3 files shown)


test/credo/check/warning/dbg_test.exs
```diff
@@ -0,0 +1,78 @@
...
+  # cases NOT raising issues
+  #
+
+  test "it should NOT report expected code" do
+    """
+    defmodule CredoSampleModule do
+      def some_function(parameter1, parameter2) do
+        parameter1 + parameter2
+      end
+    end
```
rrrene, line 16: I think we are missing a couple of important examples for "expected code" that should not yield an issue:

```suggestion
    defmodule CredoSampleModule do
      @dbg :should_not_raise_an_issue
      
      def dbg(my_param) do
        # ...
      end
      
      def some_function(parameter1, parameter2) do
        dbg(parameter1)
      end
      
      def some_other_function(parameter1, parameter2) do
        dbg = parameter1 + parameter2
      end
    end
```

lib/credo/check/warning/dbg.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Credo.Check.Warning.Dbg do
+  use Credo.Check,
+    base_priority: :high,
+    explanations: [
+      check: """
+      This check warns about calls to the dbg/3 macro
+      or the dbg/2 Kernel fn.
```
rrrene, line 7: Nitpicking for good measure: Let's change this to

```suggestion
      Calls to dbg/2 and dbg/3 should mostly be used during debugging sessions.

      This check warns about those calls, because they probably have been committed
      in error.
```

lib/credo/check/warning/dbg.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Credo.Check.Warning.Dbg do
+  use Credo.Check,
+    base_priority: :high,
```
rrrene, line 3: We could think about deactivating this check for older versions of Elixir:

```suggestion
    base_priority: :high,
    elixir_version: ">= 1.14.0-dev",
```

rrrene (changes requested):
Thanks for putting this together!

A few minor nitpicks below:

test/credo/check/warning/dbg_test.exs
ding-an-sich: Those were a lot trickier so I had to rewrite the check from scratch.

ding-an-sich:
I decided to not check for the use of Macro.dbg/3 because that function could be used as a fallback when building custom backends for dbg().

ding-an-sich:
I have streamlined the way we check if a given :dbg call should be a warning or not. If the user attempts to override the auto-imported dbg fns dbg/0, dbg/1, dbg/2 he'll get a compile error, so we shouldn't worry about those.

lib/credo/check/warning/dbg.ex
```diff
@@ -37,10 +38,15 @@ defmodule Credo.Check.Warning.Dbg do
            issue_meta: issue_meta,
            issues: issues
          } = acc
-       )
-       when length(args) < 3 do
-    issues = issues_for_call(meta, issues, issue_meta)
-    {ast, Map.put(acc, :issues, issues)}
+       ) do
+    is_dbg? = Enum.count_until(args, @dbg_max_arity + 1) < @dbg_max_arity + 1
```
rrrene, line 42: This is unnecessary, you can just match for 

```elixir
{:dbg, meta, [_single_param]} = ast,
```

and 

```elixir
{:dbg, meta, [_first_param, _second_param]} = ast,
```

to match the two cases where you want to raise an issue. Matching is always better than counting :+1:

lib/credo/check/warning/dbg.ex
```diff
@@ -0,0 +1,111 @@
...
+  defp traverse(
+         {:dbg, _meta, _} = ast,
+         %{
+           skip_next_match?: true
+         } = acc
+       ) do
+    {ast, Map.put(acc, :skip_next_match?, false)}
+  end
+
+  defp traverse({:=, _meta, children} = ast, acc) do
```
rrrene, line 73: This does only catch matching a variable, but not using it, e.g.

```elixir
x = dbg + something
```

is not matched as well as


```elixir
some_fun(dbg)
```

You just have to add a case where you match on variable usage to catch all those cases and not raise an issue:

```elixir
  defp traverse(
         {:dbg, _meta, nil} = ast,
         acc
       ) do
    {ast, acc}
  end
```

Next, you can lose the `acc` construct for good 👍

lib/credo/check/warning/dbg.ex
```diff
@@ -0,0 +1,111 @@
...
+    if is_dbg? do
+      issues = issues_for_call(meta, issues, issue_meta)
+      {ast, Map.put(acc, :issues, issues)}
+    else
+      {ast, acc}
+    end
+  end
+
+  defp traverse(
+         {:dbg, meta, nil} = ast,
```
rrrene, line 53: Regarding my earlier comment: Why do we need `acc` if you already match on the `nil` case here?

lib/credo/check/warning/dbg.ex
ding-an-sich: True! Will refactor!

lib/credo/check/warning/dbg.ex
ding-an-sich: We have to differentiate between `dbg = something` and `dbg()` (i.e. the use of dbg/0).

ding-an-sich: Oooo, I did not realize that calling dbg() actually builds a node with an empty child list `[]` and defining/using a variable builds a node with `nil` as child. That significantly simplies things.

ding-an-sich: Should be a lot simpler now that all matches are unambiguous.

rrrene: Great. I also think the code is much more readable now :+1:

lib/credo/check/warning/dbg.ex
```diff
@@ -0,0 +1,71 @@
...
+  defp traverse(
+         {:dbg, meta, [_first_param, _second_param]} = ast,
+         issues,
+         issue_meta
+       ) do
+    {ast, issues_for_call(meta, issues, issue_meta)}
+  end
+
+  defp traverse(
+         {{:., _, [{:__aliases__, _, [:Kernel]}, :dbg]}, meta, _args} = ast,
```
wingyplus, line 48: We should add 1 case for `Elixir.Kernel.dbg`

Ref: https://github.com/rrrene/credo/pull/998/files#diff-5c99577ee3ac0438232258234dece6b69f1649a7f430290e09fbeaf5dec67b0aR34-R40

rrrene: @wingyplus Out of curiosity: Is this for cases were I explicitly spell `Elixir.Kernel.dbg(arg)` in my code?

wingyplus: @rrrene yes. But I think it’s rare case to calling a function like this in elixir.

ding-an-sich:
@rrrene, any chance to get this rolling again? Not sure how to tackle the red CI checks.

rrrene (commented):
Absolutely. We just have to add an ID for the upcoming SARIF support (#1011)

rrrene:

PR #1000, opened 2022-09-11
Add `Credo.Check.Design.AliasOnlyToBoundaries` module

cheerfulstoic:
This is a check which

 * lets you know if you are aliasing too far into other modules
 * lets you know if you should be aliasing deeper into modules

My choosing the level of your alias it is possible to make it clear in your code (i.e. inside of your functions) when a module is a reference to an external or internal module.

Part of this check makes suggestions for aliases similar to the `AliasUsage` check, though `AliasUsage` requires you to alias directly to the module.

This check also allows an `exceptions` option which can be useful for modules which are references very frequently.  For example `MyApp.Blog.Post` may be referenced many, many times and there may only be one module ending in `.Post`, so you can make `MyApp.Blog.Post` an exception so that you could alias directly to it from anywhere.

### DIFF
--- a/lib/credo/check/design/alias_only_to_boundaries.ex
+++ b/lib/credo/check/design/alias_only_to_boundaries.ex
@@ -1,0 +1,315 @@
+defmodule Credo.Check.Design.AliasOnlyToBoundaries do
+  use Credo.Check,
+    base_priority: :normal,
+    explanations: [
+      check: """
+      When defining aliases to other modules you don't need to alias directly to other modules.
+      For example:
+
+          alias MyApp.Accounts.User
+          # From here you would refer to `User`
+
+          alias MyApp.Accounts
+          # From here you would refer to `Accounts.User`
+
+      If you are inside of a particular module namespace, aliasing directly to other modules can be
+      very useful. But when aliasing directly to modules which are outside of the current namespace,
+      you can lose the context of which module is being used in your code
+
+      In this example we are in the `MyApp.Accounts.Access` module. Referencing `User` directly
+      makes sense because it's in the `MyApp.Accounts` namespace, but `Post` is outside of `MyApp.Accounts`:
+
+          defmodule MyApp.Accounts.Access do
+            alias MyApp.Accounts.User
+            alias MyApp.Posts.Post
+
+            def owns?(%User{} = user, %Post{} = post) do
+              # ...
+            end
+
+      If the alias is made at a higher level, you then need to refer to the context in your code, helping
+      to make clear where the module is coming from:
+
+          defmodule MyApp.Accounts.Access do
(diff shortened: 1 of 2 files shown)


gunnar2k (approved):

rrrene:
Hi, sorry for not commenting earlier. 

I find this check interesting, but I am wondering if we really need this in Credo core? Is the interest in having this check really as broad as you make it out to be?

P.S.: There are half a dozen places where you have two spaces (`  `) in the explanation for the check. Why is that?

cheerfulstoic:
No worries 😄 

Easy question first: the double spaces are just the way I learned to write 😆   I don't really have a preference, so I changed them to single spaces (at least, I think I did.  You mean the spaces after periods, right?)

It's your library, so you can make the call on if it belongs.  I'll lay out my case, though 😄 

I'm happy for this to be an experimental and/or controvertial check.  I realize that it might not be for everybody, which is one of the things I love about credo's configurability.  I would say, though, that I've started using it in a big app and there was some reluctance from the team at first, but a couple of people liked it once they started to use it.  It's a bit hard to understand at first and I've taken a couple of stabs at the explination so far.  I'll likely need to keep working on that.

The part of the rule that I would most strongly encourage adopting into `credo` core is the part which recommends pulling-back existing alises so that modules aren't aliased directly (the "You are aliasing too far into another module..." warning).  In my experience so far it very often makes the code that I'm reading clearer. I put in the `exceptions` option because I agree that in many codebases there are modules which are so ubiquitous and unique that it doesn't hurt readability to alias to them directly.

The other part of the rule "Nested modules could be aliased at the top of the invoking module" is something that I implemented later because it seemed that it would be useful to have.  Since the first warning is incompatible with the `AliasUsage` check, the second part takes the place of `AliasUsage` with the goal of a more nuanced suggestion for modules/namespaces to alias.  I would actually say that this second warning shouldn't be considered done yet.  I just recently saw a need for it and I've been slowly adding test cases to build it out.  I'm trying it out on the codebase I mentioned so that I can base my examples on real-world cases.  I'm hoping that will inform my decision on if this second warning should use the `exceptions` option of maybe if there should be two separate `exceptions` options (for example).

Thanks for considering my submission.  I'd love your guidance on how this might be structured (i.e. as two separate checks), and of course I'd be happy to answer any other questions.

rrrene:
I apologize for the age/inactivity on this issue. It seems like there was only a limited amount of interest in getting this check merged. Sometimes people weigh in on things I do not see and persuade me to include in Credo (config comments have been such an instance).

I am closing this for now. Please feel free to re-open this at your discretion.

cheerfulstoic:
No worries, and thanks!  I've been finding this rule useful, so I think I'll still make this a library and see where it goes from there 👍

Issue #993, opened 2022-08-30
False positive on Credo.Check.Refactor.Apply when piped


halostatue:
I had false positives on this when I had changed from `apply(module, fun, [arg1, arg2])` to a piped version (because there was another call around it).

I decided to make a verification in tests rather than try to specify it otherwise.

### Environment

* Credo version (`mix credo -v`): 1.6.6 / `master`
* Erlang/Elixir version (`elixir -v`): 1.13.4 / OTP 24
* Operating system: macOS

### What were you trying to do?

```elixir
  test "it should NOT report violation for piped apply/3 when fun is a var" do
    """
    defmodule Test do
      def some_function(module, fun, arg1, arg2) do
        module
        |> apply(fun, [arg1, arg2])
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> refute_issues()
  end
```

### Expected outcome

`.`

### Actual outcome

```
  1) test it should NOT report violation for piped apply/3 when fun is a var (Credo.Check.Refactor.ApplyTest)
     test/credo/check/refactor/apply_test.exs:110
     There should be no issues, got 1: [
       %Credo.Issue{
         category: :refactor,
         check: Credo.Check.Refactor.Apply,
         column: nil,
         diff_marker: nil,
         exit_status: 8,
         filename: "test-untitled-576460752303404795.ex",
         line_no: 4,
         message: "Avoid `apply/2` and `apply/3` when the number of arguments is known",
         meta: [],
         priority: -7,
         scope: "Test.some_function",
         severity: 1,
         trigger: nil
       }
     ]
     code: |> refute_issues()
     stacktrace:
       (credo 1.6.6) lib/credo/test/assertions.ex:17: Credo.Test.Assertions.refute_issues/1
       test/credo/check/refactor/apply_test.exs:121: (test)
```

rrrene:
Thx for reporting! 👍

PR #1001, opened 2022-09-12
fix: False positive on Credo.Check.Refactor.Apply when piped

cheerfulstoic:
An attempt to fix https://github.com/rrrene/credo/issues/993

### DIFF
--- a/test/credo/check/refactor/apply_test.exs
+++ b/test/credo/check/refactor/apply_test.exs
@@ -107,6 +107,46 @@ defmodule Credo.Check.Refactor.ApplyTest do
     |> refute_issues()
   end
 
+  test "it should NOT report violation for piped apply/3 when fun is a var" do
+    """
+    defmodule Test do
+      def some_function(module, fun, arg1, arg2) do
+        module
+        |> apply(fun1, [arg1, arg2])
+      end
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check)
+    |> refute_issues()
+  end
+
+  test "it should report violation for apply/2 in a pipe" do
+    """
+    defmodule Test do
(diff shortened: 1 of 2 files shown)


test/credo/check/refactor/apply_test.exs
```diff
@@ -107,6 +107,20 @@ defmodule Credo.Check.Refactor.ApplyTest do
...
+        module
+        |> apply(fun, [arg1, arg2])
+      end
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check)
+    |> refute_issues()
+  end
+
```
rrrene, line 123: We need to make sure that the cases where we expect issues are still yielding them after this PR is merged:

```suggestion

  test "it should report violation for apply/2 in a pipe" do
    """
    defmodule Test do
      def some_function(fun, arg1, arg2) do
        fun |> apply([arg1, arg2])
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end

  test "it should report a violation for apply/3 in a pipe" do
    """
    defmodule Test do
      def some_function(module, arg1, arg2) do
        module |> apply( :fun_name, [arg1, arg2])
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end
```

rrrene (commented):
Great! Just a small request :+1:

test/credo/check/refactor/apply_test.exs
cheerfulstoic: Nice catch, thank you!  This led me to a nicer solution where I just called `issue` again with a re-combined ast which looks like an `apply` without a pipe.

rrrene:

Issue #1003, opened 2022-09-23
Custom Check Testing Documentation


LeakyBucket:
### What were you trying to do?

Use well tested Custom Checks without adding `credo` to our production environment

### Expected outcome

Be able to write and test custom checks without needing to add `credo` to `extra_applications`

### Actual outcome

Adding `{:ok, _} = Application.ensure_all_started(:credo)` to `test_helper.exs` appears to work with the exception that credo does not need to be configured for all environments.

I am just wondering if this could be added to the docs (https://github.com/rrrene/credo/blob/v1.6.7/guides/custom_checks/testing_checks.md#testing-custom-checks) as an option for those who don't want to include `credo` everywhere?  Or is there a pitfall here that I am not seeing?

PR #1004, opened 2022-09-26
Application config agnostic testing instructions

LeakyBucket:
The current instructions for testing a custom check do not work if you do not want to add `credo` to your production build.  Adding `:credo` to `extra_applications` causes environments where credo is not included to fail to start.

As an alternative this PR replaces that instruction with an instruction to add `Application.ensure_all_started(:credo)` to the `test_helper.exs` file.  That way credo starts but does not have to be included in all envs.

See https://github.com/rrrene/credo/issues/1003

### DIFF
--- a/guides/custom_checks/testing_checks.md
+++ b/guides/custom_checks/testing_checks.md
@@ -11,15 +11,10 @@ Using this module will:
 * import all the functions from this module
 * make the test case `:async` by default (use `async: false` to opt out of this behaviour)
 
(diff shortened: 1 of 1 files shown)


LeakyBucket:
👋 Any interest in this change.  This is based on a solution we put in place a few months ago after initially following the guides to setup the custom check.

rrrene:
@LeakyBucket Absolutely, thx 👍

PR #1006, opened 2022-10-11
Add test for case-sensitive alias ordering

alecho:
I have an app where this is the case. Vim's `sort` and recode's alias order say that this test should be correct but credo disagrees.

I took a stab at trying to fix this but I only had a limited time and I couldn't get it work work with existing tests. I'll open an issue and link theses. Let me know if there's anything else I can do.

### DIFF
--- a/lib/credo/check/readability/alias_order.ex
+++ b/lib/credo/check/readability/alias_order.ex
@@ -1,6 +1,9 @@
 defmodule Credo.Check.Readability.AliasOrder do
   use Credo.Check,
     base_priority: :low,
+    param_defaults: [
+      sort_method: :alpha
+    ],
     explanations: [
       check: """
       Alphabetically ordered lists are more easily scannable by the read.
@@ -38,7 +41,16 @@ defmodule Credo.Check.Readability.AliasOrder do
       Like all `Readability` issues, this one is not a technical concern.
       But you can improve the odds of others reading and liking your code by making
       it easier to follow.
-      """
+      """,
+      params: [
+        sort_method: """
+        The ordering method to use.
+
+        Options
+        - `:aplha` - Alphabetical case-insensitive sorting.
+        - `:ascii` - Case-sensitive sorting where capitals are ordered earier than their lower case equivilants.
+        """
+      ]
     ]
 
   alias Credo.Code.Name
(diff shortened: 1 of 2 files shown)


rrrene:
The check is supposed to do the opposite of your test.

In the current implementation, we want "natural language sorting", i.e. the way people unaware of ASCII would sort letters manually `a > A > b > B > c`. 

If you want to provide a different implementation, it should be opt-in :+1:

alecho:
@rrrene Thanks for the comment. I was able to get an implementation working more easily than I expected. I added a test for each path, both raising issues and passing. Let me know what else I can do!

alecho:
@rrrene Anything I can do to help move this along?

lib/credo/check/readability/alias_order.ex
```diff
@@ -38,34 +41,44 @@ defmodule Credo.Check.Readability.AliasOrder do
...
       it easier to follow.
-      """
+      """,
+      params: [
+        sort_method: """
+        The ordering method to use.
+
+        Options
+        - `:aplha` - Alphabetical case-insensitive sorting.
+        - `:ascii` - Case-sensitive sorting where capitals are ordered earier than their lower case equivilants.
```
dmarcoux, line 51:
```suggestion
        - `:ascii` - Case-sensitive sorting where upper case characters are ordered before their lower case equivalent.
```

test/credo/check/readability/alias_order_test.exs
```diff
@@ -228,4 +240,16 @@ defmodule Credo.Check.Readability.AliasOrderTest do
       assert issue.trigger == "Sorter"
     end)
   end
+
+  test "it should repot violation with case-sensitive sorting" do
```
rrrene, line 244: This test says that it should report a violation, but then `refute_issues/1` is used at the end.

Should the title read "... should NOT report ...". If so, please move this check up (we always group happy path tests at the top) :+1:

kybishop: The test here is actually incorrect; ASCII sorting should require the uppercase version to be first in the list, so a violation should be reported here. I'll see if I can take a stab at fixing up this PR tonight, as I desperately want this feature 😅

PR #1014, opened 2022-11-25
properly analyze function with guards

pnezis:
In case of a guard the `ast` was not properly traversed and in case of a callback implementation the `@impl` attribute was not propagated to the remaining clauses


rrrene:

PR #1018, opened 2022-12-08
feat: add dbg warning check

hamptokr:
This PR adds a check similar to the `IO.inspect` warning. The new `dbg` macro will also get a ton of use during debugging and it would be great to check for this in a similar way.

This is my first time jumping into the credo source so if I missed anything please let me know, will be happy to fix it! For the `id` I just grabbed the next available `EX50..` number, `EX5026`.

Thanks ahead of time for your review & all your hard work on credo! ❤️


hamptokr:
Ah I see #997 already covers this, I'll close this, sorry for the noise!

PR #1020, opened 2022-12-16
Fix bugs in Credo.Check.Readability.NestedFunctionCalls check

Stratus3D:
While using `Credo.Check.Readability.NestedFunctionCalls` I found two bugs:

### 1. Doesn't ignore function calls already in a pipeline

It would complain about the inner function call here:

``` elixir
[1,2,3,4]
|> Enum.map(SomeMod.some_fun(argument))
```

But this cannot be addressed as `SomeMod.some_fun(argument)` is for the second argument and cannot be extracted out into a pipeline.

### 2. Doesn't recognize local function calls as function calls

It would warn about:

```elixir
Enum.shuffle(Enum.uniq([1,2,2,3,3]))
```

But not warn about:

```elixir
Enum.shuffle(my_uniq([1,2,2,3,3]))
```

### Tests

For the bugs above I wrote additional tests and committed them (see the first commit in this PR).

### Solution

For bug #1 I refactored some code in `NestedFunctionCalls` and added another clause to explicitly ignore function calls already in a pipeline.

For bug #2 I realized the patterns in the code only matched on fully-qualified function calls. `Credo.Check.Refactor.PipeChainStart` was similar to the `NestedFunctionCalls` code in some ways and worked with both fully-qualified and local function calls. I was able to fix the bug by extracting the `PipeChainStart` module's `valid_chain_start?/3` function into a new shared module named `Credo.Check.PipeHelpers` so I can use the same function in both `PipeChainStart` and `NestedFunctionCalls`. This fixed the bug.

I'm not sure how best to organize this code and I wasn't sure where to put the new `PipeHelpers` module. Happy to make any changes you request here.

Thanks for credo! It's a very important part of our CI process.


rrrene:
Hi, sorry for the slow response.

The test cases from this PR are great, but coupling checks together just because they are seemingly dealing with the same technical construct is a bad idea IMHO as it adds a lot of mental overhead for later bugfixes and extensions.

`PipeChainStart` concerns itself with the start of pipes.

`NestedFunctionCalls` concerns itself with nested function calls that could be refactored into a pipe. 

With this PR, `NestedFunctionCalls` starts using a magic "pipe helper" that provides a function to deduce if an AST node is a valid start for a chain. That makes sense in the context of `PipeChainStart`. How does this make sense in the context of `NestedFunctionCalls`? A "*valid* start" in which sense?

Stratus3D:
No worries at all, sorry for introducing the bugs in `NestedFunctionCalls` check in the first place.

The `valid_chain_start?/3` returns `true` for an expression that would not work in a pipeline, and thus only be suitable as the start of the pipeline.  For example, an arithmetic expression (`3 + 2`). And `false` for any expression that would work in a pipeline. For example, invocation of a function with one argument (`somefun("arg")`). Perhaps the function would be better named `expression_cannot_be_in_pipeline?/3` or something along those lines?

`NestedFunctionCalls` uses `valid_chain_start?/3` to determine if an expression that is not in a pipeline could be converted to one. Specifically, by default, if there are 2 nested function calls they should be converted to a pipeline. If `valid_chain_start?/2` doesn't return false twice we won't warn the user.

There are two ways I can address this:

* Rename `valid_chain_start?/3` to something like `cannot_be_in_pipeline?/3`. I introduced `PipeHelper` because I saw that pattern being used in some existing checks for code shared between checks.
* Revert the changes I've made to `PipeChainStart` and duplicate the code I need into `NestedFunctionCalls`, and rename the  `valid_chain_start?/3` to something like `cannot_be_in_pipeline?/3`.

Please let me know what you would like me to do. Happy to make any changes you request.

rrrene:
@Stratus3D Yeah, we should simply duplicate the seemingly similar parts of the code.

I tried to re-familiarize myself with this PR and I still don't understand what the changes do different now and which changes are for which of the listed bugs and which parts of PipeHelper can be thrown away in the future under which conditions.

In my humble opinion this has to do with the code de-duplication, which - it seems - was driven by "this code looks alike on a lexical level" and not "this code serves the same purpose in the same domain/context and will always change in sync and should therefore be unified to have a single source of truth".

Stratus3D:
Thanks @rrrene ! I have duplicated the code into a module inside the `Credo.Check.Readability.NestedFunctionCalls` module. Can you review again?

rrrene:
@Stratus3D LGTM! Thank you! :+1:

PR #1032, opened 2023-03-15
Add check to prefer explicit lists over ~w and ~W sigils

adamu:
This is a check I have wanted myself for a while. I thought I would submit it upstream in case it's useful to others.

I have added it as disabled by default/controversial.

### DIFF
--- a/lib/credo/check/readability/no_list_sigils.ex
+++ b/lib/credo/check/readability/no_list_sigils.ex
@@ -1,0 +1,63 @@
+defmodule Credo.Check.Readability.NoListSigils do
+  @moduledoc """
+  A check to suggest explicit lists over the ~w and ~W sigls.
+  """
+  use Credo.Check,
+    param_defaults: [],
+    explanations: [
+      check: ~S"""
+      Although the ~w and ~W sigils allow for brevity when writing code, code is
+      read many more times than it is written. List definitions are both relatively
+      brief and completely unambiguous; consider defining lists explicitly rather than
+      via the sigil.
+
+      Instead of
+
+          ~w'foo bar baz'
+          ~w/例（括弧）+追加 別言葉/a
+          ~W({"key":"value"} {"key2":"value2"})
+
+      Prefer:
+
+          ["foo", "bar", "baz"]
+          [:"例（括弧）+追加", :別言葉]
+          [~S({"key":"value"}), ~S({"key2":"value2"})]
+      """,
+      params: []
+    ]
+
+  def run(source_file, params \\ []) do
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:sigil_w, context, _args} = ast, issues, issue_meta) do
+    issue =
+      format_issue(
+        issue_meta,
+        message: ~S{Avoid ~w sigil: ~w(foo bar), prefer explicit lists: ["foo", "bar"]},
+        trigger: :sigil_w,
+        line_no: context[:line]
+      )
+
+    {ast, [issue | issues]}
+  end
+
+  defp traverse({:sigil_W, context, _args} = ast, issues, issue_meta) do
+    issue =
+      format_issue(
+        issue_meta,
+        message:
+          ~S<Avoid ~W sigil: ~W({"key":"value"}), prefer explicit lists: [~S({"key":"value"})]>,
+        trigger: :sigil_W,
+        line_no: context[:line]
+      )
+
+    {ast, [issue | issues]}
+  end
+
+  defp traverse(ast, issues, _issue_meta) do
+    {ast, issues}
+  end
+end
--- a/.credo.exs
+++ b/.credo.exs
@@ -178,6 +178,7 @@
           {Credo.Check.Readability.ImplTrue, []},
           {Credo.Check.Readability.MultiAlias, []},
           {Credo.Check.Readability.NestedFunctionCalls, []},
+          {Credo.Check.Readability.NoListSigils, []},
           {Credo.Check.Readability.OneArityFunctionInPipe, []},
           {Credo.Check.Readability.SeparateAliasRequire, []},
           {Credo.Check.Readability.SingleFunctionToBlockPipe, []},
--- a/test/credo/check/readability/no_list_sigils_test.exs
+++ b/test/credo/check/readability/no_list_sigils_test.exs
@@ -1,0 +1,41 @@
+defmodule Credo.Check.Readability.NoListSigilsTest do
+  use Credo.Test.Case, async: true
+
+  @described_check Credo.Check.Readability.NoListSigils
+
+  #
+  # cases NOT raising issues
+  #
+
+  test "it should NOT report when no ~w or ~W sigil is used" do
+    ~S"""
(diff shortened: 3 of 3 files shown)


lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+          ~w/例（括弧）+追加 別言葉/a
+          ~W({"key":"value"} {"key2":"value2"})
+
+      Prefer:
+
+          ["foo", "bar", "baz"]
+          [:"例（括弧）+追加", :別言葉]
+          [~S({"key":"value"}), ~S({"key2":"value2"})]
+      """,
+      params: []
```
rrrene, line 26: I think we need at least two params to be able to configure this check:

One to only apply this to atom lists, e.g. disallow `~w/atom1 atom2/a`.

One that makes it so we only trigger the check when `\` is used to include a "spaced element" in the list, e.g. `~w/file\ 1.docx file\ 2.docx/` would trigger, but `~w/strict --format json/` would not when this param is set to `true`.

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+
+          ~w'foo bar baz'
+          ~w/例（括弧）+追加 別言葉/a
+          ~W({"key":"value"} {"key2":"value2"})
+
+      Prefer:
+
+          ["foo", "bar", "baz"]
+          [:"例（括弧）+追加", :別言葉]
+          [~S({"key":"value"}), ~S({"key2":"value2"})]
```
rrrene, line 24:
```suggestion
      # preferred

          ["foo", "bar", "baz"]
          [:"例（括弧）+追加", :別言葉]
          [~S({"key":"value"}), ~S({"key2":"value2"})]
          
      # NOT preferred

          ~w'foo bar baz'
          ~w/例（括弧）+追加 別言葉/a
          ~W({"key":"value"} {"key2":"value2"})

      Like all `Readability` issues, this one is not a technical concern.
      But you can improve the odds of others reading and liking your code by making
      it easier to follow.
```

This should be consistent with other Readability checks (lead with the preferred variant and include the readability disclaimer) :+1:

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+  A check to suggest explicit lists over the ~w and ~W sigls.
+  """
+  use Credo.Check,
+    param_defaults: [],
+    explanations: [
+      check: ~S"""
+      Although the ~w and ~W sigils allow for brevity when writing code, code is
+      read many more times than it is written. List definitions are both relatively
+      brief and completely unambiguous; consider defining lists explicitly rather than
+      via the sigil.
```
rrrene, line 12:
```suggestion
      Code is read more times than it is written and although the ~w and ~W sigils 
      allow for brevity in code, explicit list definitions have the benefit that they are 
      both relatively brief and completely unambiguous.
```

I think that leading with the benefits of what NOT to do weakens the reasoning behind this check. Instead we should put the reading > writing argument first :+1:

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:sigil_w, context, _args} = ast, issues, issue_meta) do
+    issue =
+      format_issue(
+        issue_meta,
+        message: ~S{Avoid ~w sigil: ~w(foo bar), prefer explicit lists: ["foo", "bar"]},
+        trigger: :sigil_w,
```
rrrene, line 40: The `:trigger` has to be the piece of code that triggered the issue (and should be highlighted in e.g. an editor). That would be `"~w"` here.

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+    {ast, [issue | issues]}
+  end
+
+  defp traverse({:sigil_W, context, _args} = ast, issues, issue_meta) do
+    issue =
+      format_issue(
+        issue_meta,
+        message:
+          ~S<Avoid ~W sigil: ~W({"key":"value"}), prefer explicit lists: [~S({"key":"value"})]>,
+        trigger: :sigil_W,
```
rrrene, line 53: see above

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+    issue_meta = IssueMeta.for(source_file, params)
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+  end
+
+  defp traverse({:sigil_w, context, _args} = ast, issues, issue_meta) do
+    issue =
+      format_issue(
+        issue_meta,
+        message: ~S{Avoid ~w sigil: ~w(foo bar), prefer explicit lists: ["foo", "bar"]},
```
rrrene, line 39: I am irritated by the fact that the messages for `~w` and `~W` are different. 

What was the reason for this decision?

rrrene (changes requested):
Looking really good. Just some housekeeping items and a little feature request 😄👍

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
+defmodule Credo.Check.Readability.NoListSigils do
```
rrrene, line 1: Also, please rename to `ListSigils` :+1:

lib/credo/check/readability/no_list_sigils.ex
```diff
@@ -0,0 +1,63 @@
...
+  @moduledoc """
+  A check to suggest explicit lists over the ~w and ~W sigls.
+  """
```
rrrene, line 4:
```suggestion
```

Checks don't define module docs.

rrrene (commented):
Checking in: Any news?

rrrene:
Checking in: Any update?

adamu:
Hi @rrrene apologies for the delay responding. This PR was open for 7 months before it got any comments, and during that time I solved my own needs locally with changes similar to the contents of this PR. Coming back to this isn't a high priority for me right now so please feel free to make any necessary changes directly, close, or accept an alternative implementation or PR!

rrrene:
Fair enough, sorry for not responding within your timeframe 👍

lib/credo/check/readability/no_list_sigils.ex
adamu: cd9c39f7

lib/credo/check/readability/no_list_sigils.ex
adamu: I'm not sure, it was probably related to the fact that `~W` does not interpolate. Anyway I unified the message in c8ba7512.

lib/credo/check/readability/no_list_sigils.ex
adamu: It's been a while since I looked at this, but does the `\` behaviour do anything. In iex it seems to have no effect:

```elixir
iex(1)> ~w(foo\ bar baz)
["foo", "bar", "baz"]
```

What is the reason why you think we should have an option to only apply the rule to atom lists?

Issue #1033, opened 2023-03-21
MissedMetadataKeyInLoggerConfig breaks with `metadata: :all`


cpiemontese:
### Environment

* Credo version (`mix credo -v`): 1.7.0
* Erlang/Elixir version (`elixir -v`): Elixir 1.14.2 (compiled with Erlang/OTP 25)
* Operating system: Ubuntu 22.04.2 LTS

### What were you trying to do?

I'm having trouble updating to 1.7.0 from 1.6.7 in my company's project, so I created a new project using `mix new`, and modified the initial dummy file to try and reproduce the error.

```elixir
defmodule Tmp do
  require Logger

  @moduledoc """
  Documentation for `Tmp`.
  """

  @doc """
  Hello world.

  ## Examples

      iex> Tmp.hello()
      :world

  """
  def test do
    Logger.debug("Test", a: 1, b: 2)
  end
end
```

My `mix.exs` is

```elixir
defmodule Tmp.MixProject do
  use Mix.Project

  def project do
    [
      app: :tmp,
      version: "0.1.0",
      elixir: "~> 1.14",
      start_permanent: Mix.env() == :prod,
      deps: deps()
    ]
  end

  # Run "mix help compile.app" to learn about applications.
  def application do
    [
      extra_applications: [:logger]
    ]
  end

  # Run "mix help deps" to learn about dependencies.
  defp deps do
    [
      {:credo, "~> 1.7", only: [:dev, :test], runtime: false}
    ]
  end
end
```

And my `config.exs` and `dev.exs`

```elixir
# config.exs

import Config

config :logger, :console, metadata: :all
config :logger, level: :info

import_config "#{config_env()}.exs"

# dev.exs

import Config
config :logger, level: :debug
```

### Expected outcome

I ran `mix credo -a --strict` and expected a positive outcome.

### Actual outcome

I got

```bash
❯ mix credo -a --strict
Checking 3 source files ...
Error while running Elixir.Credo.Check.Warning.MissedMetadataKeyInLoggerConfig on lib/tmp.ex

12:52:50.663 [error] Task #PID<0.385.0> started from #PID<0.379.0> terminating
** (FunctionClauseError) no function clause matching in Keyword.drop/2
    (elixir 1.14.2) lib/keyword.ex:1223: Keyword.drop([a: 1, b: 2], :all)
[...]

PR #1035, opened 2023-03-22
Fixes `:all` Logger metadata handling from MissedMetadataKeyInLoggerConfig check

antedeguemon:
Hey folks! 🖖 

Credo 1.7 was released with the [MissedMetadataKeyInLoggerConfig](https://github.com/rrrene/credo/blob/master/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex) check. The check ensures all `Logger` calls have their metadata fields defined in the `configs.exs` - in other words, it ensures all metadata fields are really logged and no metadata field is ignored from the logs.

There is a small issue with the new check: 

- It expects the metadata keys from the config file to be a list. 
- But it may be the `:all` - which makes all metadata fields to be logged.
 
The issue is that when the value is `:all`, the check crashes. 😿  

### What this PR does?

Even though it doesn't make much sense to run `MissedMetadataKeyInLoggerConfig` check for the codebases where all metadata fields are already logged, the check is enabled by default Credo configs. 
 
So this pull request fixes the problem by skipping issues when `metadata_keys == :all`. 
   
### More details

[From Logger.Backends.Console documentation:](https://hexdocs.pm/logger/1.14.3/Logger.Backends.Console.html#module-options)

>:metadata - the metadata to be printed by $metadata. Defaults to an empty list (no metadata). Setting :metadata to :all prints all metadata. See the "Metadata" section for more information.

### DIFF
--- a/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
+++ b/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
@@ -121,6 +121,10 @@ defmodule Credo.Check.Warning.MissedMetadataKeyInLoggerConfig do
     nil
   end
 
+  defp issue_for_call(_logger_metadata, _meta, _issue_meta, :all) do
+    nil
+  end
+
   defp issue_for_call(logger_metadata, meta, issue_meta, metadata_keys) do
     if Keyword.keyword?(logger_metadata) do
       case Keyword.drop(logger_metadata, metadata_keys) do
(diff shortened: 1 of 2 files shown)


cpiemontese:
This should close #1033

rrrene:
This is a great find (we obviously missed `:all`). 

But why would we look at every single potential issue to skip issue creation if `:all` is set. 

Couldn't we skip the check altogether in that case?

rrrene:
@antedeguemon *ping*

antedeguemon:
Hey folks!

Sorry about the year-long delay on this pull request, life got in the middle.

I'm closing this since https://github.com/rrrene/credo/pull/1044 already fixed it. Kudos to @davorbadrov.

rrrene:
@antedeguemon No worries. We all have lifes outside open source, even outside the internet. ✌️ 

Thank you to everyone contributing to projects in their spare time. It is a real net positive ❤️

PR #1043, opened 2023-04-16
Dependency sorting check

Sgoettschkes:
This check ensures the dependencies in the mix file are sorted alphabetically.

ToDos:

- [ ] Read the dependencies from the project function instead of assuming the name of the private function
- [ ] Add tests
- [ ] Add docs
- [ ] Adjust documentation to include `mix.exs` into the credo config
- [ ] Discuss edge cases


rrrene:
Hi, I don't think this is a good idea. 

It doesn't help that this isn't a case of "look we implemented this in our company, maybe it's useful for others" ...

Sgoettschkes:
> Hi, I don't think this is a good idea.

Sure, no worries!

PR #1044, opened 2023-04-17
fix logger metadata check when metadata :all set

davorbadrov:
Hello, this is another take on this PR https://github.com/rrrene/credo/pull/1035 which fixes the `metadata: :all` behavior for [MissedMetadataKeyInLoggerConfig](https://github.com/rrrene/credo/blob/master/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex) check.

The issue with the check is that it doesn't properly support the `metadata: :all` configuration. This PR is different from the linked one because it skips running the check entirely if `metadata: :all` is set (since it makes no sense to run it), I've also added a new test that covers the scenario.

Please let me know if it needs to be tweaked before merging.

**What this PR does?**

This pull request prevents credo from crashing when `metadata_keys == :all` and when `MissedMetadataKeyInLoggerConfig` check runs.

**More details**

[From Logger.Backends.Console documentation:](https://hexdocs.pm/logger/1.14.3/Logger.Backends.Console.html#module-options)

>    :metadata - the metadata to be printed by $metadata. Defaults to an empty list (no metadata). Setting :metadata to :all prints all metadata. See the "Metadata" section for more information.

### DIFF
--- a/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
+++ b/lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
@@ -44,16 +44,14 @@ defmodule Credo.Check.Warning.MissedMetadataKeyInLoggerConfig do
   def run(%SourceFile{} = source_file, params) do
     issue_meta = IssueMeta.for(source_file, params)
 
-    case ignore_check?(issue_meta) do
-      true ->
-        []
+    if ignore_check?(issue_meta) do
+      []
+    else
+      state = {false, []}
 
-      false ->
-        state = {false, []}
+      {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)
 
-        {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)
-
-        issues
+      issues
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
```diff
@@ -43,11 +43,25 @@ defmodule Credo.Check.Warning.MissedMetadataKeyInLoggerConfig do
...
+        []
 
-    issues
+      false ->
+        state = {false, []}
+
+        {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)
+
+        issues
+    end
```
Nezteb, line 57:
```elixir
if ignore_check?(issue_meta) do
  []
else
  state = {false, []}

  {_, issues} = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta), state)

  issues
end
```
😄

Nezteb:
For what it's worth I pulled this locally, built `credo`, tested with it, and it works with the `metadata: :all` logger config. 😄

lib/credo/check/warning/missed_metadata_key_in_logger_config.ex
davorbadrov: True :)

Nezteb:
Also just to link to the original issue: https://github.com/rrrene/credo/issues/1033

cc @rrrene 😄

rrrene:
@davorbadrov @Nezteb Thx! 🧡

PR #1045, opened 2023-04-21
Update missed_metadata_key_in_logger_config.ex

BlueHatbRit:
It looks like the `/n` in the code sample causes ex_docs to add a line break and corrupt the code block. Removing `/n` fixes this.


rrrene:
Hi, thanks for highlighting this. The solution is to put two backslashes, because removing the `\n` changes the code example rather than fixing it :+1:

PR #1057, opened 2023-07-14
If reading from standard input, do not consult the filesystem

### LINKED ISSUE #1054: Analyzing a single file through stdin is slow when individual check configs use `:excluded` option
### Environment

* Credo version (`mix credo -v`):

```
1.7.0-ref.version-updates.7b967c3d+uncommittedchanges
```

```
Erlang/OTP 25 [erts-13.2.2.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]

Elixir 1.14.5 (compiled with Erlang/OTP 25)
```

* Operating system:

```
Distributor ID:	Pop
Description:	Pop!_OS 22.04 LTS
Release:	22.04
Codename:	jammy
```

I'm using the new `lexical_credo` plugin for the new `lexical` language server, which uses the `--read-from-stdin` credo flag when analyzing the single file you have open. The plugin expects analysis to complete in 50ms or less or it times out and doesn't display results.  This works fine most of the time (analysis on a single file from stdin is 10's of ms or less usually).

But: if you have a `.credo.exs` file with excluded files configured for a specific check, e.g.:

```
{Credo.Check.Consistency.MultiAliasImportRequireUse, files: %{excluded: ["lib/my_app_web.ex"]}},
```

...then the analysis takes a lot longer (~100-200ms).

### Expected outcome

I expect that analysis of a single file from stdin should take the same amount of time with or without the `excluded` option set on a specific check.

### Actual outcome

It takes too long and the plugin cannot display realtime results.

### See also:

https://github.com/lexical-lsp/lexical/pull/254

scohen:
Files coming in via standard input are not on the filesystem, and consulting includes and excludes makes an execution run hundreds of times slower than it should be.

This PR checks to see if the execution is reading from standard input and if so, returns an empty list to bypass the checks.

Fixes #1054

### DIFF
--- a/lib/credo/check/runner.ex
+++ b/lib/credo/check/runner.ex
@@ -42,12 +42,17 @@ defmodule Credo.Check.Runner do
     files_excluded = Params.files_excluded(params, check)
 
     found_relevant_files =
-      if files_included == [] and files_excluded == [] do
-        []
-      else
-        exec
-        |> Execution.working_dir()
-        |> Credo.Sources.find_in_dir(files_included, files_excluded)
+      cond do
+        files_included == [] and files_excluded == [] ->
+          []
+
+        exec.read_from_stdin ->
+          []
+
+        true ->
+          exec
+          |> Execution.working_dir()
+          |> Credo.Sources.find_in_dir(files_included, files_excluded)
       end
 
     source_files =


rrrene:
@scohen Thx! :+1:

PR #1064, opened 2023-08-10
Credo.Check.Warning.WrongTestFileExtension: Ensure that all test files end in `_test.exs`

adriankumpf:
This extends the `WrongTestFileExtension` check to ensure that all test files end in `_test.exs`.

Previously, only test files ending in `_test.ex` were reported. Now it also reports `.exs` files without the `_test` suffix.

### DIFF
--- a/lib/credo/check/warning/wrong_test_file_extension.ex
+++ b/lib/credo/check/warning/wrong_test_file_extension.ex
@@ -2,42 +2,58 @@ defmodule Credo.Check.Warning.WrongTestFileExtension do
   use Credo.Check,
     id: "EX5025",
     base_priority: :high,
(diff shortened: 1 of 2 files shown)


rrrene:
Thx for the effort you put into this PR.

Unfortunately, I am not a fan. The current implementation of the check covers a narrowly defined, easy mistake (naming a manually added file `*_test.ex` instead of the required `*_test.exs`). This is something that has happened to me a dozen times and does require no configuration.

The idea that every single file in `test/` *must* be named `*_test.exs` does not seem right to me. This is not a stupid idea, it is just not a good idea for a general rule in a linter IMHO (this rule change would break projects like Phoenix, Nx, Nerves & Credo itself, which would all have to configure exceptions that were not necessary before the change).

PR #1071, opened 2023-09-12
Cache scope priorities

sasa1977:
Currently, every call to `format_issue` recalculates the scope priorities. This significantly hurts performance when many issues are reported. 

This change adds another ETS table (following the present cache design) to lazily cache these prios. On a large project with almost 7000 issues, this brings the running time of `mix credo` from 26 seconds to 10 seconds. 

Obviously there's more room for improvement, but I think this gain is already significant, and the change is simple enough to be merged on its own. Time-permitting I'll investigate further, but I can't make any promises.

### DIFF
--- a/lib/credo/application.ex
+++ b/lib/credo/application.ex
@@ -8,6 +8,7 @@ defmodule Credo.Application do
     Credo.Service.SourceFileAST,
     Credo.Service.SourceFileLines,
     Credo.Service.SourceFileScopes,
+    Credo.Service.SourceFileScopePriorities,
     Credo.Service.SourceFileSource
   ]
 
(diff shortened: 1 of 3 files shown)


sasa1977:
Hey, it's been a while since I've created this PR, so I'm pinging to check if could I get some feedback on this?

rrrene:
@sasa1977 Crap, I remember looking at this and being excited to merge this ... sorry I missed this, it looks great!

Will publish an RC with some other patches shortly :+1:

rrrene:
@sasa1977 This is published as part of `v1.7.2-rc.2`.

PR #1076, opened 2023-10-02
show sarif as format option

grrrisu:
When I recently set up code scanning for Github, I noticed that the SARIF format was not documented everywhere. This PR extends the documentation in this regard.


rrrene:
@grrrisu Thx!

PR #1092, opened 2023-11-17
Upgrade CI (Test against OTP 26, 24.3 instead of 24, 1.16 rc)

PragTob:
Commits mostly stand by themselves and should include rational, I can spin off more PRs if wanted but they were _somehow_ related


PragTob:
The one failing 1.16.0-rc.0 is due to #1091 (which is partially why added this, as I wanted to see it fail in CI), so this is working as intended :grin: 

Thanks for all your work! :green_heart:

PR #1095, opened 2023-11-19
fix: MultiAliasImportRequireUse collection and false positive

### LINKED ISSUE #1023: False positive on MultiAliasImportRequireUse when using multiple modules in same file
### Precheck

   * I have not found any other issue about MultiAlias used in files with multiple modules

### Environment

* Credo version (`mix credo -v`): 1.6.7
* Erlang/Elixir version (`elixir -v`): 1.13.4-otp-25 and 1.12.3-otp-24
* Operating system: Ubuntu 22 and macOS Ventura

Example repository: https://github.com/ZeLarpMaster/CredoMultiAliasExample

TL;DR in a single file:

```elixir
defmodule CredoMultiAliasExample.Bar do
  @moduledoc "This module aliases another module"

  alias CredoMultiAliasExample.Foo
end

defmodule CredoMultiAliasExample.Baz do
  @moduledoc "This module aliases a different module under the same parent"

  alias CredoMultiAliasExample.Bar
end
```

Moving the 2nd `defmodule` into its own file fixes the issue even though that's effectively the same thing.

### Expected outcome

No issues found

### Actual outcome

>   Consistency
┃
┃ [C] ↗ Most of the time you are using the multi-alias/require/import/use syntax, but here you are using multiple single
┃       directives
┃       lib/credo_multi_alias_example.ex:16 #(CredoMultiAliasExample.Baz)

msutkowski:
Fixes #1023 

When you have files that contain multiple modules this rule will raise a false positive and force a user to disable the rule for the file. These changes solve that, but they _do not_ solve the issue where you have mixed usage in just a single file. We could do that, but the scope was increase quite a bit and most likely isn't super important.

### DIFF
--- a/test/credo/check/consistency/multi_alias_import_require_use_test.exs
+++ b/test/credo/check/consistency/multi_alias_import_require_use_test.exs
@@ -24,6 +24,32 @@ defmodule Credo.Check.Consistency.MultiAliasImportRequireUseTest do
     require Foo.Quux
   end
   """
+  @multi_module_same_file """
+  defmodule CredoMultiAliasExample.SetMultiAliasToSingles do
+    @moduledoc "This modules does many aliases to set the consistency to multi-alias"
+
+    alias Credo.CLI.{Command, Output}
+    alias Config.{Reader, Provider}
+  end
+
+  defmodule CredoMultiAliasExample.Foo do
+    @moduledoc "This module has a function"
+
+    def test, do: :ok
+  end
+
+  defmodule CredoMultiAliasExample.Bar do
+    @moduledoc "This module aliases another module"
+
+    alias CredoMultiAliasExample.Foo
+  end
+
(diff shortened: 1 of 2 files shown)


test/credo/check/consistency/multi_alias_import_require_use_test.exs
```diff
@@ -50,6 +83,13 @@ defmodule Credo.Check.Consistency.MultiAliasImportRequireUseTest do
     |> refute_issues()
   end
 
+  test "it should not report errors when the single syntax is used consistently" do
+    [@single]
+    |> to_source_files
+    |> run_check(@described_check)
+    |> refute_issues()
+  end
```
msutkowski, line 91: This is just moved up - it was in the wrong section.

rrrene:
@msutkowski Thx!

PR #1094, opened 2023-11-18
fix: NegatedIsNil memory leak

msutkowski:
Fixes #1056.

Prior to this, this example test would raise 9 issues:

```elixir
test "regression test for issue list duplication" do
  """
  defmodule CredoSampleModule do
    def hello(a) when not is_nil(a), do: a
    def foo1(x) when is_integer(x), do: x
    def foo2(x) when is_integer(x), do: x
    def foo3(x) when is_integer(x), do: x
  end
  """
  |> to_source_file()
  |> run_check(@described_check)
  |> assert_issue()
end
```

PR #1096, opened 2023-11-22
fix: negated_is_nil when nested in guards

pdm-jd:
If a custom guard uses `not is_nil(foo)`, the number of issues reported by `Credo.Check.Refactor.NegatedIsNil` increases significantly.

For instance, this would report three violations.

```elixir
defguard custom_guard(value) when not is_nil(value)

def bar(foo, baz) when not is_nil(foo) and custom_guard(baz) do
end
```

This change ensures only a maximum of one issue per part of the guard will be reported. See the added tests for examples.


pdm-jd:
Oof, https://github.com/rrrene/credo/pull/1094 beat me too it.

rrrene:
@pdm-jd Still glad you took the time :+1:

PR #1097, opened 2023-12-01
Metadata in logger for 1.15 Elixir

### LINKED ISSUE #1070: MissedMetadataKeyInLoggerConfig is incorrect
### Environment

* Credo version (`mix credo -v`): `1.7.0-ref.initial.292d461+uncommittedchanges`
* Erlang/Elixir version (`elixir -v`): Elixir `1.15.4`, Erlang `26.0.2`
* Operating system: Fedora 38

Using default credo config, just installed as a dependency.

### The issue

I have this in my code

```elixir
    Logger.metadata(job: job.id)
```

which produces this credo warning

```
┃       Logger metadata key job will be ignored in production
````

I have this in my `config/config.exs`

```elixir
config :logger, :default_formatter,
  format: "$time [$level] $metadata $message\n",
  metadata: [:job]
```

and my `config/prod.exs` is empty.

So I'd say the metadata should work fine in prod too. I tried running the app with `MIX_ENV=prod` and surely I can see the metadata in the log.

### Expected outcome

Credo shouldn't generate a warning in this case.

hissssst:
Fixes #1070

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,0 +1,3 @@
+import Config
+
+import_config "#{config_env()}.exs"
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,0 +1,5 @@
+import Config
+
+config :logger, :default_formatter, metadata: [:key_new]
+
+config :logger, :console, metadata: [:key_old]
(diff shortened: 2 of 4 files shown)


rrrene:
@hissssst Thank you for putting this together. As you can see in the checks, it unfortunately doesn't work in the greater scheme of things for Credo.

That being said, your PR inspired the commit referenced above, which should fix this for all supported versions of Elixir :+1:

You can try this by setting the Credo dep to

```elixir
{:credo, github: "rrrene/credo"}
```

Please report back if this works for your use-case 👍

PR #1103, opened 2024-01-04
refactoring opp: [F] ↗  is more efficient than

bradhanks:
Changed Enum.filter/2 |> Enum.count/1 to Enum.count/2 as recommended by credo.


rrrene:
Thx!

PR #1077, opened 2023-10-04
fixes: slow multicore performance on task runners

Eein:
closes #1058

This PR removes the unnecessary worker code that was lowering performance on multicore systems.

I've used async streams instead of map/each on async await and updated the runner code to use an async stream keeping the exec.max_concurrent_check_runs to keep the same expected behaviour and make the code more modern.

This should resolve slow runs on multicore machines :)

Feel free to add any suggestions.

PR #1111, opened 2024-01-12
Improve usage of Tasks

novaugust:
Similar to #1077, optimizes the use of streams of tasks in credo in a few more places.

in addition to swapping `Enum.map(&Task.async(...))` -> `Task.async_stream`, this pr also removes ordering from the streams (where possible) for more slight performance gains (h/t @milmazz)

### DIFF
--- a/lib/credo/check/design/duplicated_code.ex
+++ b/lib/credo/check/design/duplicated_code.ex
@@ -88,14 +88,11 @@ defmodule Credo.Check.Design.DuplicatedCode do
   end
 
   defp duplicate_nodes(source_files, mass_threshold) do
-    chunked_nodes =
+    nodes =
       source_files
       |> Enum.chunk_every(30)
       |> Task.async_stream(&calculate_hashes_for_chunk(&1, mass_threshold), timeout: :infinity, ordered: false)
-      |> Enum.map(fn {:ok, hashes} -> hashes end)
-
-    nodes =
-      Enum.reduce(chunked_nodes, %{}, fn current_hashes, existing_hashes ->
+      |> Enum.reduce(%{}, fn {:ok, current_hashes}, existing_hashes ->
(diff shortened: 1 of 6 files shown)


lib/credo/check/design/duplicated_code.ex
```diff
@@ -97,8 +91,8 @@ defmodule Credo.Check.Design.DuplicatedCode do
     chunked_nodes =
       source_files
       |> Enum.chunk_every(30)
```
novaugust, line 93: unsure if this chunk is still necessary - was it here just to make larger subproblems for each task?

rrrene:
Seems to improve performance :+1:

Can we have some real-world numbers regarding speed difference, like in #1077?

novaugust:
running against 3842 files, it's negligible but there.

here's credo master, with credo's default checks enabled:
```
> # credo master
> hyperfine 'mix credo' -i --warmup 1
Benchmark 1: mix credo
  Time (mean ± σ):     24.770 s ±  0.331 s    [User: 187.391 s, System: 13.010 s]
  Range (min … max):   24.340 s … 25.235 s    10 runs
```

and here's against this branch:
```
> # Task.async_stream refactor
> hyperfine 'mix credo' -i --warmup 1
Benchmark 1: mix credo
  Time (mean ± σ):     24.077 s ±  0.421 s    [User: 182.291 s, System: 10.692 s]
  Range (min … max):   23.674 s … 24.633 s    10 runs
```

still, negligible enough that i'd understand not merging.

looking at just DuplicatedCode, a single-run test took it from 17412ms to 15451ms, but running it in isolation probably isn't the best test.
the crux here is that the checks that are slow in my codebase aren't really affected by these changes, so i don't see much of a difference :)

rrrene:
I'm sure merging this is worth the performance benefit. My preliminary tests all show good results.

I was just looking for some real world data (and maybe a check that earlier OTP versions are not getting slower). :+1:

PR #1118, opened 2024-03-15
Fix comment regarding how `credo:disable-for-lines` works

StandardMikeBenza:
"preceding" refers to something before the statement.  The correct word is `following` here.


rrrene:
@StandardMikeBenza Ouch! 😳 Great catch, thx!

PR #1121, opened 2024-03-24
Ignore PredicateFunctionNames when is impl

### LINKED ISSUE #1108: Predicate callbacks from behavior are not ignored
### Precheck

- [X] For bugs, please do a quick search and make sure the bug has not yet been reported here

### Environment

* Credo version (`mix credo -v`): `1.7.3-ref.dependabot/hex/credo-1.7.3.433dfec`
* Erlang/Elixir version (`elixir -v`): 1.16.0 OTP 26.2.1
* Operating system: MacOS Ventura

[NervesSSH implements the Erlang `:ssh_server_key_api`](https://github.com/nerves-project/nerves_ssh/blob/main/lib/nerves_ssh/keys.ex#L13-L14) behavior which has [an `is_auth_key/3` callback definition](https://www.erlang.org/doc/man/ssh_server_key_api#Module:is_auth_key-3). The currently fails the predicate check despite it having `@impl :ssh_server_key_api`.

This might be an extreme edge case and maybe not need to be handled, but wanted to bring attention just in case

### Expected outcome

Passing when `is_*` function is an implementation of a behavior callback.

### Actual outcome

```
> nerves_ssh ?4 % mix credo -a --strict
Checking 11 source files ...

  Code Readability                                                                                                                                                                   
┃ 
┃ [R] ↗ Predicate function names should not start with 'is', and should end in a question mark.
┃       lib/nerves_ssh/keys.ex:14:7 #(NervesSSH.Keys.is_auth_key)
```

quangngd:
# Overview

- Closes #1108

# Impl details

If any `:def, :defp, :defmacro` is preceded with `:impl`, we ignore.

Code is mostly inspired from the impl of `Specs`

### DIFF
--- a/lib/credo/check/readability/predicate_function_names.ex
+++ b/lib/credo/check/readability/predicate_function_names.ex
@@ -108,10 +108,10 @@ defmodule Credo.Check.Readability.PredicateFunctionNames do
   end
 
   defp issues_for_definition(op, [{name, meta, args} | _], issues, issue_meta, impl_list) do
-    if {name, length(args)} not in impl_list do
-      issues_for_name(op, name, meta, issues, issue_meta)
-    else
+    if {name, length(args)} in impl_list do
       issues
+    else
+      issues_for_name(op, name, meta, issues, issue_meta)
     end
   end
 
--- a/lib/credo/check/readability/predicate_function_names.ex
+++ b/lib/credo/check/readability/predicate_function_names.ex
@@ -72,23 +72,29 @@ defmodule Credo.Check.Readability.PredicateFunctionNames do
   end
 
   defp do_find_impls_in_block({:@, _, [{:impl, _, [impl]}]}, acc) when impl != false do
-    [:impl | acc]
+    [:record_next_definition | acc]
   end
 
   # def when
-  defp do_find_impls_in_block({keyword, meta, [{:when, _, def_ast} | _]}, [:impl | impls])
+  defp do_find_impls_in_block({keyword, meta, [{:when, _, def_ast} | _]}, [
+         :record_next_definition | impls
+       ])
        when keyword in @def_ops do
     do_find_impls_in_block({keyword, meta, def_ast}, [:impl | impls])
   end
 
   # def 0 arity
-  defp do_find_impls_in_block({keyword, _meta, [{name, _, nil} | _]}, [:impl | impls])
+  defp do_find_impls_in_block({keyword, _meta, [{name, _, nil} | _]}, [
+         :record_next_definition | impls
+       ])
        when keyword in @def_ops do
     [{name, 0} | impls]
   end
 
   # def n arity
-  defp do_find_impls_in_block({keyword, _meta, [{name, _, args} | _]}, [:impl | impls])
+  defp do_find_impls_in_block({keyword, _meta, [{name, _, args} | _]}, [
+         :record_next_definition | impls
+       ])
        when keyword in @def_ops do
     [{name, length(args)} | impls]
   end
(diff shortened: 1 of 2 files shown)


rrrene:
I like this in principle and will add suggestions later this week.

test/credo/check/readability/predicate_function_names_test.exs
```diff
@@ -81,6 +81,34 @@ defmodule Credo.Check.Readability.PredicateFunctionNamesTest do
...
+      @impl Foo
+      def is_bar(a) do
+      end
+    end
+    """
+    |> to_source_file
+    |> run_check(@described_check)
+    |> refute_issues()
+  end
+
```
rrrene, line 111: The proposed change yields a couple of false negatives which we should iron out before I review this PR:

```suggestion

  test "it should report a violation with false negatives" do
    """
    defmodule FooImpl do
      def impl(false), do: false
      def impl(true), do: true

      def is_bar do
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end

  test "it should report a violation with false negatives /2" do
    """
    defmodule FooImpl do

      impl(true)

      def is_bar do
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end

  test "it should report a violation with false negatives /3" do
    """
    defmodule Foo do
      @impl is_bar(a)
    end

    defmodule FooImpl do
      def is_bar do
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end

  test "it should report a violation with false negatives /4" do
    """
    defmodule Foo do
      @impl true
    end

    defmodule FooImpl do
      def is_bar do
      end
    end
    """
    |> to_source_file
    |> run_check(@described_check)
    |> assert_issue()
  end
```

quangngd: Sure! I took another look and came up with some changes.

lib/credo/check/readability/predicate_function_names.ex
```diff
@@ -52,41 +52,87 @@ defmodule Credo.Check.Readability.PredicateFunctionNames do
...
+    {ast, impls}
+  end
+
+  defp find_impls_in_block(block_args) when is_list(block_args) do
+    block_args
+    |> Enum.reduce([], &do_find_impls_in_block/2)
+  end
+
+  defp do_find_impls_in_block({:@, _, [{:impl, _, [impl]}]}, acc) when impl != false do
+    [:impl | acc]
```
rrrene, line 75: Let's rename this atom to something more descriptive, like `:record_next_definition`, so that it tells what it does.

lib/credo/check/readability/predicate_function_names.ex
```diff
@@ -52,41 +52,87 @@ defmodule Credo.Check.Readability.PredicateFunctionNames do
   def run(%SourceFile{} = source_file, params) do
     issue_meta = IssueMeta.for(source_file, params)
 
-    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))
+    impl_list = Credo.Code.prewalk(source_file, &find_impls(&1, &2))
+
+    Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta, impl_list))
```
rrrene, line 57: This traverses the whole file for `@impl` even if there is no need.

I would love for this logic to be reversed:

```elixir
issue_candidates = Credo.Code.prewalk(source_file, &traverse(&1, &2, issue_meta))

if issue_candidates == [] do
  []
else
  impl_list = Credo.Code.prewalk(source_file, &find_impls(&1, &2))

  # and then reject the issue_candidates that have an `@impl`
end
```

WDYT?

rrrene (changes requested):
Two more nitpicks, than we should be ready to merge  :+1:

lib/credo/check/readability/predicate_function_names.ex
quangngd: updated

lib/credo/check/readability/predicate_function_names.ex
quangngd: refactored

rrrene:
@quangngd Thx! ✌️

PR #1125, opened 2024-04-23
Fix deprecation for 1.17

mhanberg:
1.17 deprecates calling remote functions without parentheses, which
means that calls like `Foo.Bar.run` and `foo = %{bar: Foo}; foo.bar.run`
should instead be formatted like
`Foo.Bar.run()` and `foo = %{bar: Foo}; foo.bar.run()`

This show up as runtime warnings.


rrrene:
I will merge this soon! Thx for bringing this up 👍

PR #1126, opened 2024-04-25
Fix incorrect columns for dbg and IO.inspect

mhanberg:
Previously, this was backfilling the column based on analyzing the
source code line, but the column is already available in the AST, so we
instead take it from there.

This allows IDE type tools to accurately create refactor "code actions".


mhanberg:
I think there was a formatting error on the master branch, went ahead and fixed it here, I can open a new PR if that is preferable tho.

rrrene:
@mhanberg Thx!

PR #1133, opened 2024-05-24
Fix the path for included files in umbrella apps

tomasz-tomczyk:
Hello! I ran into an issue where `Credo.Check.Refactor.PassAsyncInTestCases` would not flag issues when running `mix credo` from the root of an umbrella application, but `cd`-ing into `apps/app_web` and running it from there would correctly flag the missing `async` option.

We only have a single `.credo.exs` config at the root and I replicated the issue with the barebones config:

```elixir
%{
  configs: [
    %{
      name: "default",
      checks: %{
        enabled: [{Credo.Check.Refactor.PassAsyncInTestCases, []}]
      }
    }
  ]
}
```

Before this change, inside the `Credo.Sources.find_in_dir` the value for `included_patterns` would compile down to:

```
["/Users/tom/my_umbrella_app/test/**/*_test.exs"]
```

But that would skip everything that was under `my_umbrella_app/apps/**/test`

I see this was intentionally changed in https://github.com/rrrene/credo/commit/ac76775a37bceea6afe66c00d1647fe60c46d755 so maybe I'm wrong to submit this PR and would appreciate feedback! Changing this locally did make the check work from both the root and the individual umbrella app directories though.

### DIFF
--- a/lib/credo/check/design/skip_test_without_comment.ex
+++ b/lib/credo/check/design/skip_test_without_comment.ex
@@ -3,7 +3,7 @@ defmodule Credo.Check.Design.SkipTestWithoutComment do
     id: "EX2003",
     base_priority: :normal,
     param_defaults: [
-      files: %{included: ["test/**/*_test.exs"]}
+      files: %{included: ["**/*_test.exs"]}
     ],
     explanations: [
       check: """
--- a/lib/credo/check/refactor/pass_async_in_test_cases.ex
+++ b/lib/credo/check/refactor/pass_async_in_test_cases.ex
@@ -3,7 +3,7 @@ defmodule Credo.Check.Refactor.PassAsyncInTestCases do
     id: "EX4031",
     base_priority: :normal,
     param_defaults: [
-      files: %{included: ["test/**/*_test.exs"]}
+      files: %{included: ["**/*_test.exs"]}
     ],
     explanations: [
       check: """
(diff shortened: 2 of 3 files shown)


rrrene:
Thx for pointing this out. I will take a closer look soon. :+1:

tomasz-tomczyk:
Thanks for taking a look!

I've reproduced in a minimal example in this repo: https://github.com/tomasz-tomczyk/credo-async-example
- [.credo.exs](https://github.com/tomasz-tomczyk/credo-async-example/blob/main/.credo.exs) with the single rule
- [test without `async` option](https://github.com/tomasz-tomczyk/credo-async-example/blob/main/apps/example_web/test/example_web/controllers/error_json_test.exs#L2)
- [CI workflow](https://github.com/tomasz-tomczyk/credo-async-example/actions/runs/9266857865/job/25491992824) running `mix credo` from root (passing) and then the `_web` directory (failing)

rrrene:
@tomasz-tomczyk I had to change the way the files were addressed, because the check for wrong file extensions would otherwise yield issues for files like `lib/foo/foo_case_test.ex` which we avoided in the past.

Please check out the RC to see if this works for you: https://github.com/rrrene/credo/releases/tag/v1.7.7-rc.0 :+1:

tomasz-tomczyk:
Ah okay, got it. Thanks! It works for us yes 👍

PR #1140, opened 2024-06-13
Ignore FunctionNames when sigil is private

leobessa:
This pull request updates the `Credo.Check.Readability.FunctionNames` module to ignore function names when the sigil is private. It also includes test cases to ensure the expected behavior.


rrrene:
@leobessa Thx!

PR #1149, opened 2024-09-01
Change to not warn on calls to IO.inspect/3

jisaacstone:
It seems unlikely that an inspect call with a specific file handler passed in would be from a debugging session, so it makes sense to exclude these calls.

Thoughts?


rrrene:
@jisaacstone Sounds good to me 👍

PR #1150, opened 2024-09-03
fix typo

includeVitor:
I believe we meant to type `"umbrella"` instead of `"umrella"` maybe we need fix it from the https://github.com/rrrene/credo/releases page.


rrrene:
@includeVitor Thx for catching this 👍

PR #1157, opened 2024-11-01
Add ** operator

gilbertbw:
This stops e.g. the following code from showing a readability warning:

```
defmodule CredoTest do
  @moduledoc """
  An example of the snakecase test failing
  """
  def a ** b do
    nil
  end
end
```


rrrene:
@gilbertbw Thx! This is part of Credo `1.7.9` :tada:

Issue #1175, opened 2025-01-09
Duplicate suggestions 


eksperimental:
### Precheck

Here a repo that reproduces the issue:
https://github.com/eksperimental-debug/credo-issues

### Environment

* Credo version (`mix credo -v`): 1.7.11
* Erlang/Elixir version (`elixir -v`): Elixir 1.18.1 (compiled with Erlang/OTP 27)

### What were you trying to do?

`mix credo --script`

### Expected outcome

To have only one suggestion

0### Actual outcome

```
❯ mix credo --strict

  Consistency
┃
┃ [C] ↗ Unused variables should be named consistently. It seems your strategy is to give them meaningful names (eg. `_foo`) but `_` does not follow that convention.
┃       lib/foo_web/components/core_components.ex:4:31 #(FooWeb.CoreComponents.icon)
┃ [C] ↗ Unused variables should be named consistently. It seems your strategy is to give them meaningful names (eg. `_foo`) but `_` does not follow that convention.
┃       lib/foo_web/components/core_components.ex:4:31 #(FooWeb.CoreComponents.icon)
```

PR #1178, opened 2025-01-31
Fix duplicated suggestions with UnusedVariableNamesTest

quangngd:
Fixes https://github.com/rrrene/credo/issues/1175


rrrene:
@quangngd Thx for this!

PR #1179, opened 2025-02-13
Extend double boolean negation to include '! not' and 'not !'

PaulOstazeski:
Currently, Credo.Check.Refactor.DoubleBooleanNegation triggers on "!!foo" and "not not foo".

This extends it to trigger on "!not foo" and "not !foo" as well.


rrrene:
Hi, thx! Will review as soon as I can!

PR #1181, opened 2025-02-28
Add columns to FIXME and TODO tag checks

mhanberg:
This allows text editors to correctly render the diagnostic

Before: 
CleanShot 2025-02-28 at 14 16 15@2x

After:

CleanShot 2025-02-28 at 14 18 20@2x

### DIFF
--- a/lib/credo/check/design/tag_fixme.ex
+++ b/lib/credo/check/design/tag_fixme.ex
@@ -40,11 +40,12 @@ defmodule Credo.Check.Design.TagFIXME do
     |> Enum.map(&issue_for(issue_meta, &1))
   end
 
-  defp issue_for(issue_meta, {line_no, _line, trigger}) do
+  defp issue_for(issue_meta, {line_no, column, _line, trigger}) do
     format_issue(
       issue_meta,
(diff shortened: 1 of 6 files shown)


mhanberg:
I'll look into those test failures

rrrene:
This is looking good so far, thanks! :+1:

We just have to make sure that it also works for module attributes, because the checks also catch the "tag" (`TODO` or `FIXME`) in those:

```elixir
@doc """
Does something incredible. Returns `:ok` or `:error`.

TODO: add example
"""
```

Judging from the code, I get the impression that this is not considered yet?

rrrene:
Hi, after writing the above, I realized that all I wrote was technically true, but of no concern to this PR 😞 🙏 

I also realized that the column should have been there in the first place and fixed something in `Credo.Check` so that `TagTODO` and `TagFIXME` aren't the only checks benefiting from our exchange :+1:

Could you check if your problem is now solved on `master`?

mhanberg:
Sorry, this got lost in the shuffle. 

I'll check the master branch

mhanberg:
Seems to work on the master branch now 👍

PR #1182, opened 2025-03-02
Add check for piping into Logger functions

hkrutzer:
See https://x.com/elixirboy/status/1770336803280556486 and https://bsky.app/profile/hauleth.dev/post/3lhguoqmpos2i


rrrene:
Hi, thx for putting this together.

But please don't link several social media posts on different networks (which not everybody might have an account for) instead of writing a simple rationale. 🙂

It is also bad for reviewers to come back to something like this and then having to open multiple tabs again, just to re-familiarize themselves. 👍

hkrutzer:
Thanks @rrrene the description was not intended to be a rationale, but to argue the merit of the check. The rationale is in the description of the check. I also added it to the PR description 👍

rrrene:
Hi, thx again for this. Unfortunately, I am closing it as I really never got why this check should be included as a standard check in Credo (the 1-sentence-explanation did not make it any clearer).

hkrutzer:
Which single sentence are you referring to? The explanation I wrote is 3 sentences:

```
      Calls into Logger can be purged when `:compile_time_purge_matching` is enabled in
      the Logger configuration. This will remove the entire pipeline.
      Use |> tap(&Logger.info(&1)) instead.
```

Also, only with permissions on the project can someone reopen a PR.

Issue #1170, opened 2024-12-18
False positive within typespecs on NestedFunctionCalls code readability check 


Eiji7:
### Environment

* Credo version (`mix credo -v`):
```
$ mix credo -v
1.7.10-ref.(HEAD odłączone na origin/master).2e15cdd2
```
* Erlang/Elixir version (`elixir -v`):
```
$ elixir -v
Erlang/OTP 27 [erts-15.2] [source] [64-bit] [smp:32:32] [ds:32:32:10] [async-threads:1] [jit:ns]

Elixir 1.18.0-rc.0 (f00f759) (compiled with Erlang/OTP 27)
```
* Operating system:
> Gentoo Linux using `desktop/plasma` profile for version `23.0` and with kernel version `6.6.41`.

### What were you trying to do?

I have simply tried to use non-literal as `Keyword`'s value. In my case it's user-defined type with some argument.

```elixir
@type type_name :: Keyword.t(Some.remote(some_arg))
```

### Expected outcome

This check should not appear for all [Typespec attributes | Module @ Elixir documentation](https://hexdocs.pm/elixir/Module.html#module-typespec-attributes).

### Actual outcome

```
Checking (…) source files ...

  Code Readability                                                                                                               
┃ 
┃ [R] → Use a pipeline instead of nested function calls.
┃       (…)

Please report incorrect results: https://github.com/rrrene/credo/issues

Analysis took 0.1 seconds (0.04s to load, 0.1s running 102 checks on (…) files)
(…) mods/funs, found 1 code readability issue.

Use `mix credo explain` to explain issues, `mix credo --help` for options.
```

rrrene:
Thx for reporting! :+1:

rrrene:
Thanks for reporting this 😀  It is now fixed on `master`.

You can try this by setting the Credo dep to

```elixir
{:credo, github: "rrrene/credo"}
```

Please report back if your issue is solved! 👍

Eiji7:
@rrrene This fixes:
https://gitlab.com/ramurix-software/dev-joy/-/blob/v2.0.0/lib/dev_joy/api/item.ex?ref_type=tags#L105
but **not**:
https://gitlab.com/ramurix-software/dev-joy/-/blob/v2.0.0/lib/dev_joy/generator/part.ex#L49

Eiji7:
[...]

PR #1185, opened 2025-03-18
Fix false positive for NestedFunctionCalls

Eiji7:
Added a guard for all typespec attributes, closes: #1170


rrrene:
This looks great 🎉 

There is already a test for `@type`. Could you update the tests to include the added cases?

Eiji7:
@rrrene Done :-)

rrrene:
@Eiji7 Thx for this!

Issue #1164, opened 2024-11-27
Credo.Check.Readability.FunctionNames potential false positives for non-sigil functions beginning with "sigil_"


brettbeatty:
### Environment

* Credo version (`mix credo -v`):
    ```
    $ mix credo -v
    1.7.10-ref.nobranch.nocommit+uncommittedchanges
    ```
* Erlang/Elixir version (`elixir -v`):
    ```
    $ elixir -v
    Erlang/OTP 27 [erts-15.1.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit:ns]
    
    Elixir 1.17.3 (compiled with Erlang/OTP 27)
    ```
* Operating system: Ubuntu 24.04

### What were you trying to do?
I have a sigil with a supporting private function. I prefixed the private function name with the sigil name, and when I updated credo it flagged it for not being in snake case.

### Expected outcome
No errors

### Actual outcome
```
┃ [R] ↗ Function/macro/guard names should be written in snake_case.
┃       lib/my_app.ex:6:8 #(MyApp.sigil_z_do)
```

It seems to happen with any function beginning with "sigil_" that is not in fact a sigil.

```
┃ [R] ↗ Function/macro/guard names should be written in snake_case.
┃       lib/my_app.ex:10:7 #(MyApp.sigil_do_something_else)
```

It seems a little odd to name a non-sigil function with a "sigil_" prefix, but as far as I'm aware there's nothing wrong with it.

Seems related to #1146

rrrene:
Yeah, it seems like this issue and the referenced issue are at odds with each other 😰 

Any suggestions how we could compromise here?

rrrene:
This is now available in Credo `1.7.15`! 😄 🎉

PR #1186, opened 2025-03-25
Fix false positive sigil-prefixed function names

brettbeatty:
My take on a fix for #1164

For functions that aren't sigils but begin with "sigil_" it checks the names as it would any other function.


rrrene:
@brettbeatty Not sure why I did not merge this sooner. Sorry. And thx! 🎉

PR #1189, opened 2025-04-04
Move regexes out of module attributes for OTP 28 compatibility

TBK145:
Also fix the `config_comment_format` regex, the `-` needed to be escaped.


rrrene:
@TBK145 Thx for this! 👍

PR #1190, opened 2025-04-06
Link to the credo docs when app is not credo

danschultzer:
The docs didn't link correctly with Credo checks in other libraries because there's just a relative ref to the extra page. This PR will autolink the Credo extra pages references as per: https://hexdocs.pm/ex_doc/readme.html#auto-linking

Unfortunately it doesn't seem that ExDoc understands how to link `e:credo:EXTRA_PAGE.html` correctly when used in the same library. The link just becomes `credo/EXTRA_PAGE.html` instead of `EXTRA_PAGE.html` properly because there's nothing in the deps. So I'm only using the autolink when outside Credo.

### DIFF
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -339,7 +339,7 @@ defmodule Credo.Check do
         end
       end
 
-    module_doc = moduledoc(opts, __CALLER__.module)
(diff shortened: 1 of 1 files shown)


rrrene:
Thx! 👍

PR #1191, opened 2025-04-07
Use latest Elixir and OTP versions in CI

danschultzer:
Got format error in CI I didn't see using Elixir 1.18 in #1190. I'm updating everything in CI here to use the latest elixir/otp versions.


rrrene:
Thx!

PR #1197, opened 2025-06-08
docs: add homebrew installation note

chenrui333:
I have recently packaged credo for homebrew-core, would be good to get called out in here. 

- https://github.com/Homebrew/homebrew-core/pull/226005

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,14 @@ And run:
 
     $ mix credo
 
+### Install via `brew`
+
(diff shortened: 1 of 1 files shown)


rrrene:
Hi, thanks for mentioning this. 

The officially supported way to use Credo is as a Mix dependency, so unfortunately this is not something I will merge at this time. 

P.S.: I don't know enough about `brew` to judge the merged formula, but why is it using `prod`? why is the test hard-coding expected standard checks?

chenrui333:
I would say feel free to modify the formula if you dont see fit (elixir is not something I am familiar with)

rrrene:
👍

PR #1198, opened 2025-06-10
fix incorrect call with Inspect.Algebra.format/2 result

Eiji7:
According to docs the result of the `Inspect.Algebra.format/2` call could be a list or binary. Therefore `Enum` functions should not be used directly on it.


rrrene:
@Eiji7 Thx! :+1:

PR #1208, opened 2025-07-07
fix failing test on the reference in re_pattern value in Erlang 28.0

Eiji7:


rrrene:
Hi, it seems I fixed the same thing in more or less the same way, which naturally does not look good ... 

Sorry, this was my sleep deprived brain not keeping up 😞

Eiji7:
Yeah, this works for me too, thanks!

PR #1211, opened 2025-07-20
fix false positive inside guards for NestedFunctionCalls readability check

Eiji7:
Hi, last time I did it for all specification module attributes and now I found that we are missing the same rule for guards.


rrrene:
@Eiji7 sorry it took this long and thx for the PR!

PR #1216, opened 2025-08-28
Fix default :excluded_paths regex in ModuleDependencies check to exclude first-level test files

alexisruccius:
The `ModuleDependencies` check was incorrectly triggered on code under the `test`
folder with default settings.

Example:

```shell
mix credo test/supex_test.exs
... 
[F] → Module has too many dependencies: 11 (max is 10)
test/supex_test.exs:1:11 #(SupexTest)
```

Cause: `Path.dirname/1` does not return a trailing `/`, so first-level test files
(e.g. `test/foo_test.exs`) were not excluded
(`lib/credo/check/refactor/module_dependencies.ex:65`).

Fix: Adjusted the default `:excluded_paths` regex to also match first-level test
files.
Added a test to confirm files in `test/` are excluded.

PS: Thanks for this awesome project :)

### DIFF
--- a/lib/credo/check/refactor/module_dependencies.ex
+++ b/lib/credo/check/refactor/module_dependencies.ex
@@ -7,7 +7,7 @@ defmodule Credo.Check.Refactor.ModuleDependencies do
       max_deps: 10,
       dependency_namespaces: [],
       excluded_namespaces: [],
-      excluded_paths: [~r"/test/", ~r"^test/"]
+      excluded_paths: [~r"/test/", ~r"^test"]
     ],
     explanations: [
       check: """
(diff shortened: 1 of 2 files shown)


rrrene:
Hi, thx for the PR. 

I think this now excludes all files in directories starting with `test`. Is this intended?

alexisruccius:
Hi :)

Yes, that’s intentional — the updated regex now matches any path starting with `test`, ensuring first-level test files like `test/foo_test.exs` are also excluded.
I understood the default intent to be to exclude all test files, and this change aligns with that.
It still only applies to paths whose dirname begins with `test`, so non-test directories (e.g. `lib/testable/...`) remain unaffected.

PR #1217, opened 2025-08-30
Added rules file to help Ai coding agents to write better code

pcharbon70:
The [Usage-rules](https://hexdocs.pm/usage_rules/readme.html) tool is a library that help ai coding agents write better Elixir code by synchronizing the coding agent rules file with the different rules for the libraries and frameworks used in a project.

Since the goal of this library is to statiscally analyze the code of a project for some coding-style that has become the de-facto standard for the Elixir community it makes sense to help coding assistant to not make those mistakes in the first place.

The usage-rules.md file contains a condensed version of Credo's rules and will be synced to the coding agent rules of those adopting the usage-rules tool. They will still be able to edit those rules to ignore the ones that they prefer not to include.

For those adopting an Ai work-flow it will hopefully help prevent the tango between the agent code generation and having to fix the code generated after the credo checks fails.

### DIFF
--- a/usage-rules.md
+++ b/usage-rules.md
@@ -1,0 +1,261 @@
+# Credo Usage Rules for AI Code Generation
+
+Comprehensive rules covering all 108 Credo checks to ensure complete compliance when generating Elixir code.
+
+## CONSISTENCY RULES (EX1001-EX1008)
+
+### Exception Naming (EX1001)
+- **MUST** use consistent exception naming patterns across codebase
+- Choose either suffix pattern: `HTTPError`, `ParseError`, `ValidationError`  
+- OR prefix pattern: `ErrorHTTP`, `ErrorParse`, `ErrorValidation`
+- **NEVER** mix patterns: `HTTPError` + `ParseException` + `ValidationFailure`
+
+### Line Endings (EX1002)
+- **MUST** use consistent line endings throughout project
+- Typically Unix `\n` (not Windows `\r\n`)
+- **NEVER** mix line ending styles in same codebase
+
+### Multi-Alias/Import/Require/Use (EX1003)
+- **MUST** be consistent with single vs multi-alias syntax
+- Choose ONE style and stick to it:
+  - Multi: `alias Ecto.{Query, Schema, Changeset}`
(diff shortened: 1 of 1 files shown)


rrrene:
Thanks for putting this together, quite impressive that we have that many checks in Credo.

I don't think this makes sense in Credo, the package, unfortunately. 

Have you considered pushing this as its own project? We could link it in Credo's README :+1:

pcharbon70:
Sure thing, I could do that and thanks for takiing the time to answer and
for Credo itself.

zachdaniel:
> I don't think this makes sense in Credo, the package, unfortunately.

I agree and disagree. I think most of this file should be removed, and instead the usage-rules.md file should just explain broadly how to use credo, i.e the mix task and its basic interface, how to skip rules if necessary, basic stuff that an agent should know.

This file isn't a place to actually describe the conventions. That is what credo is for 😆

zachdaniel:
The idea with the file living in this project is so that it can change as the library changes etc.

pcharbon70:
I understand that the file is quite extensive compared to some written for
other libraries but in this case the idea behind it was to prevent the
coding agent from making the mistakes before Credo needed to fix it. There
are some recurring mistakes that agents make. Maybe this could be trimmed
down to those specific mistakes and eventually reduced even more as the
models become better at Elixir coding.

zachdaniel:
> the idea behind it was to prevent the
coding agent from making the mistakes before Credo needed to fix it.

Two things

1. context management is key. The benefit of a linter is that the agent doesn't *have* to have the rules loaded up into their context window.
2. credo is fully configurable to disable any of these checks.

The only tenable usage-rules.md file would be one that explains how to run credo, configure it, etc. IMO.

zachdaniel:
Of course it would be worth finding out from @rrrene if a trimmed down doc like the one I'm suggesting would even be welcome, but I do personally think something lightweight would be good 😄

rrrene:
Hi, I am not knowledgable enough in the "AI assisted coding" space at this moment in time. 

Maybe in the future this will be more wide spread and obvious 👍

PR #1219, opened 2025-09-19
Make MatchInCondition warn on matches that contain conditionals

s3cur3:
Hi René!

Not sure if this is something you'd want in mainline Credo, but I figured I'd open the PR to start a conversation. Maybe you'd want it, but wrapped in a config?

This makes MatchInCondition slightly more restrictive in response to a bug we ran into at work.

The problematic code started out looking like this:

```elixir
if user_id = post.user_id do
  do_stuff(user_id, post)
end
```

...but was then modified with the intent that we should `do_stuff` only if we both had a user ID _and_ the post was already published:

```elixir
if user_id = post.user_id && post.published? do
  do_stuff(user_id, post)
end
```

That is, of course, not what the code actually does, but neither I, the author, nor two other reviewers caught that `user_id` was going to get assigned the value `true` when we entered the block.

With this change, `MatchInCondition` will be persnickety about cases like these and require you to extract the match out of the conditional.

### DIFF
--- a/lib/credo/check/refactor/match_in_condition.ex
+++ b/lib/credo/check/refactor/match_in_condition.ex
@@ -32,6 +32,10 @@ defmodule Credo.Check.Refactor.MatchInCondition do
             do_something(contents)
           end
 
+          if contents = file.contents && allowed? do
+            do_something(contents)
+          end
+
       If you want to match for something and execute another block otherwise,
       consider using a `case` statement:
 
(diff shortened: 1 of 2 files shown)


rrrene:
I really like this. Not sure if this should even be its own check, since it could benefit from its own explanation why this is flagged. WDYT?

s3cur3:
Hrm... personally I was surprised that this wasn't flagged by the existing check based on the check's top-line explanation of "Pattern matching should only ever be used for simple assignments inside `if` and `unless` clauses." (At first I thought we must have not had the `MatchInCondition` rule enabled on our codebase at all.) I'd consider the assignments with conditionals to no longer be simple, in the sense that there's an implicit branch in there.

I'm happy to break it out into its own check if you feel strongly, though!

rrrene:
@s3cur3 Nah, you are right, merged it.

Last thing I am wondering is how far we want to take this, because there is probably an argument for disallowing any operator on the right side of the assignment. WDYT?

s3cur3:
Yeah, I'd support that for sure! Want a PR for that?

PR #1199, opened 2025-06-11
feat (ExpensiveEmptyEnumCheck): Catch checking length(enum) > 0

s3cur3:
Hi René! Thanks so much for your work on Credo. It's one of the first things I set up on a new project. ☺️ 

This PR adds a new condition to the `ExpensiveEmptyEnumCheck` for cases like:

```elixir
if length(my_enum) > 0 do
  "not empty"
else
  "empty"
end
```

PR #1226, opened 2025-11-13
feat(ExpensiveEmptyEnumCheck): add comparisons against 1

hauleth:
Comparison `length(xs) == 0` or `length(xs) > 0` is the same as checking for `length(xs) >= 1` or `1 <= length(xs)`. There we add that additional checks.


hauleth:
This expands on #1199 by adding more checks when comparing with one.

rrrene:
@hauleth Thx! Appreciated as always 😎

PR #1230, opened 2025-12-02
Parallelize ConfigCommentFinder to significantly reduce runtime on large projects

s3cur3:
On my project with ~3,500 source files running on my M4 Max MacBook Pro, I was seeing `ConfigCommentFinder` take about 14 seconds to run (roughly half of the overall ~30 second runtime of `mix credo`).

With this PR, rather than processing each file individually, Credo will parallelize this processing across all schedulers. With this change, the runtime for this step dropped to about 2.5 seconds on my machine, reducing the overall runtime of `mix credo` by about 40%.


rrrene:
Thx! 🎉

PR #1231, opened 2025-12-03
Fix ExpensiveEnumCheck suggestion when the trigger is within a guard

### LINKED ISSUE #1229: [1.7.14] `ExpensiveEmptyEnumCheck` now triggers for guards
Using Credo 1.7.14, having a guard with `length(list) > 0` will now trigger `ExpensiveEmptyEnumCheck`, suggesting to use `Enum.empty?/1` instead.

This does not work, since `Enum.empty?/1` is not supported in guard clauses.

s3cur3:
This PR makes the suggestions given by ExpensiveEnumCheck more specific to the context:

- When a `length/1` call is in a guard, it says "Using `length/1` is expensive, prefer comparing against the empty list."
- When it's in the body of a function, it says "Using `length/1` is expensive, prefer `Enum.empty?/1`.

I've also significantly beefed up the tests around guards to make sure we're covering all the desired cases.

Resolves #1229

### DIFF
--- a/lib/credo/check/warning/expensive_empty_enum_check.ex
+++ b/lib/credo/check/warning/expensive_empty_enum_check.ex
@@ -28,7 +28,7 @@ defmodule Credo.Check.Warning.ExpensiveEmptyEnumCheck do
   @doc false
   @impl true
   def run(%SourceFile{} = source_file, params) do
-    ctx = Context.build(source_file, params, __MODULE__) |> Map.put(:handled_guards, MapSet.new())
+    ctx = Context.build(source_file, params, __MODULE__)
     result = Credo.Code.prewalk(source_file, &walk/2, ctx)
     result.issues
   end
@@ -50,7 +50,10 @@ defmodule Credo.Check.Warning.ExpensiveEmptyEnumCheck do
 
   # Handles cases like `when is_list(enum) and not is_map(enum) and length(enum) == 0`
   defp walk({:when, _meta, [_, guard_expr]} = ast, ctx) do
-    {ast, find_length_comparisons_in_guard(guard_expr, ctx)}
+    case find_in_guard(guard_expr, ctx) do
+      [] -> {ast, ctx}
+      issues -> {nil, Enum.reduce(issues, ctx, fn issue, ctx -> put_issue(ctx, issue) end)}
+    end
   end
 
   for {pattern, trigger} <- @comparisons do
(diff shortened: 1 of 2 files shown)


rrrene:
Hi, I get this is just a draft right now, but it would be unfair not to make the following observation:

In essence #1229 talks about the message being wrong. I reverted the issue message and now it is "not wrong, but maybe sub-optimal".

I think it is fair to say that this PR got way too complicated way too fast, especially if we also want to offer specific suggestions in the message (this PR tries to hint at that, but for `length(list) > 0` we would have to suggest `list != []` instead of `list == []`. which this PR does [edit: or it did, anyhow, not the relevant point ^^]).

s3cur3:
@rrrene I'm not sure this is true:

> I reverted the issue message and now it is "not wrong, but maybe sub-optimal".

The message _is_ wrong if the thing being checked is not a list; `Enum.count(map) == 0`, for instance, would include list equality in the suggestions.

I tried to make the suggestion for the guards general enough that the user will need to make a judgement call about how to implement it (for instance, if the guard was checking `length(l) != 0`, you can't merely check `l != []`, you need to also confirm `is_list(l)`).

rrrene:
You are, of course, right.

The thing I was trying to say was: The message

> "#{trigger} is expensive. Prefer Enum.empty?/1 or list == []"

has not bothered anybody for 8 years. Then I messed up by removing the last part.

This PR would add a lot of complexity and I am not sure that is the right way to go.

s3cur3:
I guess I'm puzzled by the criticism about complexity... it's a net +43 lines of production code, rigorously tested, in a leaf module (so it doesn't make anything else in the codebase harder to maintain). The file has had something like 20-25 code changes in its 8 years of life (slightly hard to count due to merge commits and docs-only commits). I guess I don't see what the point of having something like a complexity budget is if not to improve the functionality. 🤷‍♂️

rrrene:
Maybe it's just a difference of opinion.

Just to avoid misunderstanding, I will demonstrate with code (might also be better since I am not a native speaker 🙏).

Let me push my take 👍

rrrene:
@s3cur3 I really liked the test suite additions and those put me in a position where I could experiment with a "simpler" approach. Thx for that! Very much appreciated 🎉

Now, my first thought was that we would not need a separate set of callbacks to walk the guard expression. We could keep track in the context whether or not we are in a guard: b9d17abab4e8599440be2e2e1839f8b3e19996f0

Then I realized that a great side-effect of the phrasing "compare with an empty list" is that it is correct whether we are in a guard or not: e78cb4e511e92879a9dd2a1f48bc0b5713d89b14

As I said in my last comment, this may come down to a preference in style. Thx very much for this exchange and all your contributions. Credo is going strong for 10 years because of community members like you 💛💚💙

s3cur3:
Fair enough! Thank you! 🥂

PR #1233, opened 2025-12-03
Vastly speed up ConfigCommentFinder

s3cur3:
On my codebase of ~3500 files, even after the parallelization in #1230, ConfigCommentFinder still took about 2.5 seconds. This is mostly attributable to `Credo.Code.Sigils.replace_with_spaces`. I experimented with making that function faster by building an iodata list rather than repeatedly concatenating binaries, and that was indeed much faster (about 100 milliseconds on my machine), but it turns out to be about 3x faster to use `Code.string_to_quoted_with_comments/1`, introduced in Elixir 1.13.

With this change, ConfigCommentFinder goes from 2.5 seconds to about 30 milliseconds on my codebase.

### DIFF
--- a/lib/credo/check/config_comment_finder.ex
+++ b/lib/credo/check/config_comment_finder.ex
@@ -31,13 +31,30 @@ defmodule Credo.Check.ConfigCommentFinder do
     source = SourceFile.source(source_file)
 
     if source =~ config_comment_format() do
-      source
-      |> Credo.Code.clean_charlists_strings_and_sigils()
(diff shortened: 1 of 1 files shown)


rrrene:
@s3cur3 This and #1232 are great! 🎉 

I did some minor changes. Would be amazing if you could run the result to see if it still performs as expected 🙏

s3cur3:
@rrrene Yep, I confirm the tip of `master` is still real fast. 😄 
image

PR #1241, opened 2026-01-09
Fix 1.20.x compatibility issue with block expression AST

Eiji7:
Hi, I have issue when testing `credo` with the latest `1.20.0-rc.0` release of `Elixir`. Looks like the `AST` for a block expression i.e. `{:__block__, _meta, _arguments}` does no longer contain empty list for `meta`, but now there is additional `:line` and `:column` keys added to it. Therefore I have simply replaced `[]` with `_meta`.

This issue have affected `Credo.Check.Readability.ModuleDoc` check to emmit false positives for all modules in my project and the mentioned tiny change fixes this issue.

The issue was originally reported on `Elixir` forum:
https://elixirforum.com/t/elixir-v1-20-0-rc-0-released-type-inference-of-all-constructs/73927

### DIFF
--- a/test/credo/code/module_test.exs
+++ b/test/credo/code/module_test.exs
@@ -9,6 +9,26 @@ defmodule Credo.Code.ModuleTest do
   # attribute
   #
 
+  test "1.20.x block AST compatibility" do
+    doc = "1.20.x compatible"
+
+    ast =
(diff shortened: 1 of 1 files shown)


rrrene:
@Eiji7 @crbelaus  This is now part of Credo [`1.7.16`](https://github.com/rrrene/credo/releases/tag/v1.7.16) 🎉

PR #1242, opened 2026-01-09
Credo.Code.Module.exception?/1 test case

Eiji7:
Simply added the test case and removed the comment about missing test.


rrrene:
@Eiji7 Thx! 🎉

PR #1243, opened 2026-01-09
Add `:force_comment_on_explicit_false` option to `PassAsyncInTestCasesTest`

### LINKED ISSUE #1240: Feature request : force adding a comment when using `async: false` in `PassAsyncInTestCases`
`Credo.Check.Refactor.PassAsyncInTestCases` is really useful to ensure people pass in `:async` explicitly. However, they sometime pass in `async: false` "by default", or for wrong reasons. Forcing to have a comment forces the person to a little more at the "why", and helps reviewers and future peoples running through the test file understanding why this `async: false` was here.

I'm not sure `Credo` checks can read comments since they use the AST (and comments are not included in the AST, or are they?) but another solution could be to simply "only allow `async: true`", which would force adding `# credo:disable-for-next-line Credo.Check.Refactor.PassAsyncInTestCases` the line before, which would, in a way, suggest to add a comment 😄 

Note that I'm willing to open a PR (not right at the moment, but in a near-ish future) if this proposal is accepted.

Gladear:
Add a parameter to `PassAsyncInTestCases` to force having a comment before a case using `async: false`.

Closes #1240

### DIFF
--- a/lib/credo/check/refactor/pass_async_in_test_cases.ex
+++ b/lib/credo/check/refactor/pass_async_in_test_cases.ex
@@ -42,10 +42,10 @@ defmodule Credo.Check.Refactor.PassAsyncInTestCases do
           {ast, ctx}
 
         {:ok, false} ->
-          if not ctx.params[:force_comment_on_explicit_false] or has_comment?(ctx, meta[:line]) do
-            {ast, ctx}
-          else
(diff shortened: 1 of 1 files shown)


rrrene:
@Gladear Thx! 🎉

rrrene:
@Gladear  This is now part of Credo [`1.7.16`](https://github.com/rrrene/credo/releases/tag/v1.7.16) 🎉

PR #1248, opened 2026-01-27
Fix error message when `PassAsyncInTestCases`'s `:force_comment_on_explicit_false` produces issues

s3cur3:
This improves the error messages produced when using the new `:force_comment_on_explicit_false` option in `PassAsyncInTestCases`.

Prior to this change, if you had:

```elixir
defmodule MyTest do
  use ExUnit.Case, async: false
  ...
end
```

...you would get the rather misleading Credo report:

```
┃ [F] → Pass an `:async` boolean option to `use` a test case module.
┃       lib/my_app.exs:2:3 #(MyTest)
```

With this change, you'll get the more precise error "Tests with `async: false` need a comment explaining why they can't be run asynchronously."


rrrene:
@s3cur3 Thanks! Completely missed this.

PR #1249, opened 2026-01-27
Explicitly support UnusedOperation checking all functions in a given module

s3cur3:
This expands the `UnusedOperation` check to (explicitly) support checking all functions in the module by omitting the list of functions. I say "explicitly" because, digging into the implementation, it turns out one could already have passed `nil` for the list of function names, but that was both undocumented and rather obtuse as a public API (naively, if I saw a config that said we'd warn about unused results from the functions `nil`, I would assume that alert never fired).

This updates the configuration for UnusedOperation to explicitly allow passing just a module, or a module plus a string to use as the message, in addition to the variants that pass an explicit list of functions.

### DIFF
--- a/lib/credo/check/warning/unused_operation.ex
+++ b/lib/credo/check/warning/unused_operation.ex
@@ -23,7 +23,7 @@ defmodule Credo.Check.Warning.UnusedOperation do
         modules:
           "The modules and functions that should trigger this check.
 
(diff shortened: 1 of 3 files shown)


rrrene:
@s3cur3 Thx for this! I changed the params format back while adding an `:all` option.

PR #1252, opened 2026-01-29
Add `ignore_comments` param to `MaxLineLength` check

Nezteb:
Long story short, a lot of projects I work with sometimes have very long comment lines and I want to ignore them in a opt-in fashion for this Credo check. 🙏

### DIFF
--- a/lib/credo/check/readability/max_line_length.ex
+++ b/lib/credo/check/readability/max_line_length.ex
@@ -7,6 +7,7 @@ defmodule Credo.Check.Readability.MaxLineLength do
       max_length: 120,
       ignore_definitions: true,
       ignore_heredocs: true,
+      ignore_comments: false,
       ignore_specs: false,
       ignore_sigils: true,
       ignore_strings: true,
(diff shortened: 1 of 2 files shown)


rrrene:
Hi, sorry for the delay. 

I will include this in Credo 1.8, just need a weekend or two more to get things into RC territory 👍 

On a side note, I want to say that this check is becoming the most configurable (is that a word?) while it should be obsolete when using the formatter, which is kind of ironic 👀

Nezteb:
> I will include this in Credo 1.8, just need a weekend or two more to get things into RC territory 👍

No rush at all! 🙏 

> I want to say that this check is becoming the most configurable (is that a word?) while it should be obsolete when using the formatter

Yeah some folks I've talked to prefer to disable this check altogether because of how Elixir's formatter treats line length config:
- https://github.com/elixir-lang/elixir/issues/7851#issuecomment-403830578
- https://github.com/elixir-lang/elixir/issues/9809#issuecomment-584546625
- https://github.com/elixir-lang/elixir/issues/11194#issuecomment-902587478
- https://github.com/elixir-lang/elixir/issues/13418#issuecomment-2004970977

Still, I opted for adding a configuration option. 😄

PR #1254, opened 2026-02-02
Add ModuleDoc `ignore_using` parameter

pierrelegall:
# Add `ignore_using` parameter to `ModuleDoc` check

Solve #1255.

## Summary

Adds an `:ignore_using` parameter to `Credo.Check.Readability.ModuleDoc` that allows users to suppress missing `@moduledoc` warnings for modules that `use` a specific module. This is useful for framework-generated modules (e.g., Phoenix controllers, LiveViews) where documentation is typically unnecessary.

### Modified files
- `lib/credo/check/readability/module_doc.ex` - Added `:ignore_using` param, updated `:ignore_names` to accept atoms and strings alongside regexes, updated documentation for both params
- `test/credo/check/readability/module_doc_test.exs` - Added 5 tests: `:ignore_names` matching (atom, string, regex, mixed list), `:ignore_using` matching (atom, string, regex, mixed list), submodule matching, non-matching, and empty param

## Parameter Details

| Parameter | Default | Description |
|-----------|---------|-------------|
| `ignore_names` | *(Phoenix conventions)* | List of modules to ignore based on their name. Accepts atoms, strings, and regexes. |
| `ignore_using` | `[]` | List of modules to ignore based on their `use` declarations. Accepts atoms, strings, and regexes. |

### Example Configuration

```elixir
# Ignore modules that use Phoenix.LiveView or any controller base module
{Credo.Check.Readability.ModuleDoc, [
  ignore_using: [Phoenix.LiveView, ~r/\.Web$/]
]}
```

### Matching Behavior

```elixir
# This module would be ignored with ignore_using: ["MyApp.Web"]
defmodule MyController do
  use MyApp.Web, :controller
end
```

- **String values** match via `String.contains?/2`
- **Atom values** are converted to their full name and matched as strings
- **Regex values** match via `String.match?/2`
- An empty list (the default) skips the check entirely for performance

## Why `ignore_using` over `ignore_names`

The existing `:ignore_names` parameter matches on module names, which means it's tied to naming conventions. This works for Phoenix defaults (e.g., `~r/\.\w+Controller$/`) but breaks down when:

- **Naming doesn't follow convention.** A module like `MyApp.Admin.Dashboard` that uses `Phoenix.LiveView` won't be caught by any name-based pattern. With `:ignore_using`, `ignore_using: [Phoenix.LiveView]` covers it regardless of how it's named.
- **Custom base modules are involved.** Projects often define `MyApp.Web` and `use` it across controllers, views, and live views. There's no single name pattern that captures all of these, but `ignore_using: ["MyApp.Web"]` handles them in one rule.
- **The ignore list becomes fragile.** Name-based regexes must be updated every time a new naming pattern appears in the project. `:ignore_using` targets the actual behavior (what a module is built on) rather than an indirect signal (what it's called), so it stays stable as the codebase evolves.

In short, `:ignore_names` filters by what modules are called; `:ignore_using` filters by what modules do.

## `ignore_names` Changes

The `:ignore_names` parameter previously only accepted regexes. It now also accepts atoms and strings, matching the same types as `:ignore_using`. Both params share the same `matches_any?/2` implementation. The documentation has been updated to reflect this.

## Test Plan

- [x] All 16 ModuleDoc tests pass
- [x] Full test suite passes (1670 tests, 0 failures)
- **`:ignore_names`**
  - [x] Atom matching works (`ignore_names: [CredoSampleModule]`)
  - [x] String matching works (`ignore_names: ["CredoSampleModule"]`)
  - [x] Regex matching works (`ignore_names: [~r/SampleModule$/]`)
  - [x] Mixed list matching works (`ignore_names: [CredoSampleModule, "MyApp.Web", ~r/Other/]`)
- **`:ignore_using`**
  - [x] Atom matching works (`ignore_using: [MyApp.Web]`)
  - [x] String matching works (`ignore_using: ["MyApp.Web"]`)
  - [x] Regex matching works (`ignore_using: [~r/\.Web$/]`)
  - [x] Mixed list matching works (`ignore_using: [GenServer, "MyApp.Web", ~r/Other/]`)
  - [x] Submodules are correctly matched
  - [x] Non-matching modules still report issues
  - [x] Empty default does not change existing behavior

### DIFF
--- a/lib/credo/check/readability/module_doc.ex
+++ b/lib/credo/check/readability/module_doc.ex
@@ -135,7 +135,7 @@ defmodule Credo.Check.Readability.ModuleDoc do
   end
 
   defp matches?(name, string_matcher) when is_binary(string_matcher) do
-    String.contains?(name, string_matcher)
+    name == string_matcher
   end
 
   defp matches?(name, regex_matcher) when is_struct(regex_matcher, Regex) do
--- a/test/credo/check/readability/module_doc_test.exs
+++ b/test/credo/check/readability/module_doc_test.exs
@@ -94,6 +94,17 @@ defmodule Credo.Check.Readability.ModuleDocTest do
     |> refute_issues()
   end
 
+  test "it should report modules when :ignore_names string does not exactly match" do
+    ~S'''
+    defmodule MyApp.Web do
+      def some_fun, do: :ok
+    end
+    '''
+    |> to_source_file()
+    |> run_check(@described_check, ignore_names: ["Web"])
+    |> assert_issue()
+  end
+
   test "it should NOT report modules when the :ignore_using param matches" do
     source_file =
       ~S'''
@@ -207,6 +218,17 @@ defmodule Credo.Check.Readability.ModuleDocTest do
     '''
     |> to_source_file
     |> run_check(@described_check, ignore_using: ["MyApp.Web"])
+    |> assert_issue()
+  end
+
+  test "it should report modules when :ignore_using string does not exactly match" do
+    ~S'''
+    defmodule CredoSampleModule do
+      use MyApp.Web, :controller
+    end
+    '''
+    |> to_source_file()
+    |> run_check(@described_check, ignore_using: ["Web"])
     |> assert_issue()
   end
 


lib/credo/check/readability/module_doc.ex
```diff
@@ -135,7 +135,7 @@ defmodule Credo.Check.Readability.ModuleDoc do
   end
 
   defp matches?(name, string_matcher) when is_binary(string_matcher) do
-    String.contains?(name, string_matcher)
+    name == string_matcher
```
pierrelegall: This is original expected behavior from my point of view.

This can be a breaking change, but I suppose users already expect exact matches and not "contains".

rrrene:
This PR is still open due to  force push. The PR was merged into `master`, just not the commits in *this* incarnation of the PR.

Closing for this reason. Thx again for this feature, I think it is a great addition! 🎉🙏

PR #1258, opened 2026-02-06
Add improved check for detecting misnamed test files

s3cur3:
A colleague and I recently spent more hours than I'd care to admit tracking down why a test file which failed when run individually did not fail the full test suite. Even once we determined conclusively that `mix test` didn't run the file at all, we couldn't figure out why.

The answer turned out to be a one-character typo in the file name: `_text.exs` versus `_test.exs`.

This check is essentially a broader version of the existing `Credo.Check.Warning.WrongTestFileExtension`. It warns on any mis-named test file which appears to represent a test case (it includes `use *.Case`). Beyond my poor colleague and I, I can imagine this helping an Elixir newbie who got the naming convention wrong and named their file something like `test/test_my_module.exs`.

I'm not sure whether this should fully replace `WrongTestFileExtension`. It _does_ make that check fully redundant, but keeping the same name would be slightly misleading.

### DIFF
--- a/lib/credo/check/warning/wrong_test_file_name.ex
+++ b/lib/credo/check/warning/wrong_test_file_name.ex
@@ -1,0 +1,72 @@
+defmodule Credo.Check.Warning.WrongTestFileName do
+  @moduledoc """
+  Ensures that `use ExUnit.Case` and related test cases are only used in test files.
+  """
+  use Credo.Check,
+    base_priority: :high,
+    category: :warning,
+    param_defaults: [
+      files: %{excluded: ["test/**/*_test.exs", "apps/**/test/**/*_test.exs"]}
+    ],
+    explanations: [
+      check: """
+      Ensures that `use ExUnit.Case` and related test cases are only used in files ending with `_test.exs`.
+
+      ExUnit.Case is designed for test modules and provides test-specific functionality
+      like database sandboxing and test macros. Using it in non-test files is likely a
+      mistake in the file name and will cause the tests in that module not to be run.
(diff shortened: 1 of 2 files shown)


rrrene:
This is interesting. I started integrating it and wondered why there are no tests. 

The reasoning that there are no tests because it is relying on Credo's `:files` param seems weird since it checks on the passed filename as the very first thing in `run/3` (and then ignored `_test.exs` which is also excluded via `:files`, but also ignores `_case.ex`, which is not excluded via `:files`).

I think this check would really benefit from a test suite that covers the typo described in this PR's description and that `use Test.Case` and `use TestCase` trigger this check.

The rest was easily fixed on my checkout 👍

s3cur3:
Re: no tests, I think you must have been looking at the shipping `WrongTestFileExtension`. There are indeed tests for this one ([link](https://github.com/rrrene/credo/pull/1258/changes#diff-a60cc0a4673cbd0dcb5979d62567e3f50f2898cf25328436698da9b3dc5bc971)). The logic for the included files is actually reversed in the newly added check: whereas the old one only _includes_ files named with a particular kind of typo (`_test.ex`), this one only _excludes_ test files named correctly (`_test.exs`).

rrrene:
@s3cur3 ouch, how embarrassing 😄 it's been a rather long week.

Will merge soon and deprecate `WrongTestFileExtension` in favor of this :+1:

s3cur3:
Haha, happens to all of us. 😄 Thank you!

PR #1259, opened 2026-02-07
Add check for forbidden functions

s3cur3:
This complements the existing `Credo.Check.Warning.ForbiddenModule`. In our codebase, there are a number of specific functions from otherwise mundane modules that we'd like to forbid. For instance, `:erlang.binary_to_term/{1,2}` is one we found being used in a surprisingly (scarily) large number of places.

This functionality could potentially be included as an expansion of `ForbiddenModule`, but it would actually make more sense to me to go the other way and allow ForbiddenFunction to be configured to warn on all functions in a given module. Not sure if that's worth the churn, though.

### DIFF
--- a/lib/credo/check/warning/forbidden_function.ex
+++ b/lib/credo/check/warning/forbidden_function.ex
@@ -1,0 +1,91 @@
+defmodule Credo.Check.Warning.ForbiddenFunction do
+  use Credo.Check,
+    base_priority: :high,
+    category: :warning,
+    param_defaults: [
+      functions: []
+    ],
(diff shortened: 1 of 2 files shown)


rrrene:
I like this, just need a bit of time to merge/integrate it 👍

s3cur3:
Looks like this was merged as c853d1ed17b01ff49c8e4b56daf57f79ac34f9be. Should I close this one? ☺️

rrrene:
Yeah, sorry. No idea why this wasn't closed automagically 🤫

PR #1260, opened 2026-02-09
feat: add UnusedFunctionParameterPattern check

paulswartz:
I think this type of code is incorrect, but I'm also open to this being a personal preference that doesn't belong here.


rrrene:
I like this, will need some time to review/integrate 👍

rrrene:
@paulswartz Thx again!

PR #1264, opened 2026-03-09
Do not report external functions wrapping a multi-clause anonymous function in `NestedFunctionCalls`

Wigny:


rrrene:
@Wigny Thx!

PR #1265, opened 2026-03-11
Fix infinite recursion on broken symlinks

s3cur3:
I had created a symlink from `CLAUDE.md` to `AGENTS.md` in my project, then later deleted `AGENTS.md`. This caused `recurse_path/1` to recurse infinitely on the `CLAUDE.md` symlink. While normally `File.regular?/1` and `File.dir?/1` treat symlinks invisibly, if the symlink is broken, both will return false.

With this fix, we stop recursing in such scenarios. (Of course, you may get an error somewhere else in your Credo checks about the broken symlink, but that's not really this function's job.)

### DIFF
--- a/lib/credo/sources.ex
+++ b/lib/credo/sources.ex
@@ -200,9 +200,16 @@ defmodule Credo.Sources do
           |> Path.wildcard()
 
         true ->
-          path
-          |> Path.wildcard()
-          |> Enum.flat_map(&recurse_path/1)
+          expanded = Path.wildcard(path)
+
+          if expanded == [path] do
+            # If the path not a regular file, nor a directory, nor does its
+            # wildcard expansion match any additional files, it is probably
+            # a broken symlink.
+            expanded
+          else
+            Enum.flat_map(expanded, &recurse_path/1)
+          end
       end
 
     Enum.map(paths, &Path.expand/1)


rrrene:
Interesting. 

When trying to reproduce this under Ubuntu by creating a dangling symlink for an `.exs` file, I get the expected result:

```
$ cd lib
$ ln -s non_existing.exs dangling_symlink.exs
$ cd -
$ mix credo

22:34:03.588 [error] Task #PID<0.440.0> started from #PID<0.95.0> terminating
** (File.Error) could not read file "/home/rrrene/projects/credo/master/lib/dangling_symlink.exs": no such file or directory
    (elixir 1.20.0-rc.3) lib/file.ex:435: File.read!/1
    (credo 1.8.0-dev) lib/credo/sources.ex:231: Credo.Sources.to_source_file/1
    (elixir 1.20.0-rc.3) lib/task/supervised.ex:105: Task.Supervised.invoke_mfa/2
    (elixir 1.20.0-rc.3) lib/task/supervised.ex:40: Task.Supervised.reply/4
Function: &:erlang.apply/2
    Args: [#Function<4.66508286/1 in Credo.Sources.to_source_file>, ["/home/rrrene/projects/credo/master/lib/dangling_symlink.exs"]]
** (EXIT from #PID<0.95.0>) an exception was raised:
    ** (File.Error) could not read file "/home/rrrene/projects/credo/master/lib/dangling_symlink.exs": no such file or directory
        (elixir 1.20.0-rc.3) lib/file.ex:435: File.read!/1
        (credo 1.8.0-dev) lib/credo/sources.ex:231: Credo.Sources.to_source_file/1
        (elixir 1.20.0-rc.3) lib/task/supervised.ex:105: Task.Supervised.invoke_mfa/2
        (elixir 1.20.0-rc.3) lib/task/supervised.ex:40: Task.Supervised.reply/4
```

Could you include more steps to reproduce? Which kind of system are you on?

s3cur3:
Ah, the Elixir file is taking the fast-path short circuit (which doesn't touch the disk at all)—the `non_wildcard_elixir_file_path?(path)` branch of that `cond`. I think to repro, it requires having the broken symlink be both not an Elixir file _and_ explicitly included in some check. Having the broken `CLAUDE.md` in my `lib` directory with only this check enabled is sufficient to repro for me: 

```
{Credo.Check.Consistency.LineEndings, files: %{included: ["lib/**/*"]}}
```

rrrene:
@s3cur3 Got it. Thx!

PR #1266, opened 2026-03-16
Fix multi-alias violation description

sl1depengwyn:
The following case
```elixir
defmodule MyApp.MyModule do
  alias Phoenix.{Controller, LiveView}
  alias MyApp.Core.{Account, User}
end
```
results in the following description:
```
The alias `` is not alphabetically ordered among its group.
```
Fixed so it looks like this
```
The alias Phoenix.{Controller, LiveView} is not alphabetically ordered among its group.
```


rrrene:
@sl1depengwyn Thx!

Issue #1268, opened 2026-03-17
WrongTestFileExtension docs say it's enabled by default, but it's missing from generated config


michallepicki:
Hello!

I saw a check I want to run called `Credo.Check.Warning.WrongTestFileExtension`, in [docs](https://hexdocs.pm/credo/1.7.17/Credo.Check.Warning.WrongTestFileExtension.html) it says it is enabled by default. But when I generated a config file with `mix credo.gen.config`, I couldn't find it listed. If it's documented as enabled by default, shouldn't it be included in the generated config file?

michallepicki:
I think the docs say it is enabled by default because 1.7.17 was published with old docs (see also https://github.com/rrrene/credo/issues/1267#issuecomment-4084584555 ), and it was disabled by default by https://github.com/rrrene/credo/commit/ab67e975e2f1ab7e3ebe0f1662789fcea6e443cc . But I don't get it, why was it disabled? I think both `WrongTestFilename` and `WrongTestFileExtension` checks are useful

rrrene:
Yes, you were right. Docs were old, just re-published them from the tag `v1.7.17` :+1:

PR #1271, opened 2026-03-18
Re-enable WrongTestFileExtension check by default

michallepicki:
I think it was disabled by mistake in https://github.com/rrrene/credo/commit/ab67e975e2f1ab7e3ebe0f1662789fcea6e443cc 

Closes https://github.com/rrrene/credo/issues/1268

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -165,7 +165,8 @@
           {Credo.Check.Warning.UnusedRegexOperation, []},
           {Credo.Check.Warning.UnusedStringOperation, []},
           {Credo.Check.Warning.UnusedTupleOperation, []},
-          {Credo.Check.Warning.WrongTestFilename, []}
+          {Credo.Check.Warning.WrongTestFilename, []},
+          {Credo.Check.Warning.WrongTestFileExtension, []}
         ],
         disabled: [
           #


rrrene:
Hi, `WrongTestFilename` should be a super-set of `WrongTestFileExtension` as it not only checks the extension.

We should probably document that :+1:

michallepicki:
> Hi, `WrongTestFilename` should be a super-set of `WrongTestFileExtension` as it not only checks the extension.

From what I saw it is not a super-set because `WrongTestFilename` only catches misnamed files that `use ExUnit.Case` directly, but it does not catch misnamed files that use a custom module created with `use ExUnit.CaseTemplate`

But when writing this comment I found that there's now an Elixir warning for such misnamed files?! So I think `WrongTestFileExtension` maybe can be deprecated in favor of this built-in Elixir warning
```
warning: the following files do not match any of the configured `:test_load_filters` / `:test_ignore_filters`:

test/hm_test.ex

This might indicate a typo in a test file name (for example, using "foo_tests.exs" instead of "foo_test.exs").

See the configuration for `:test_pattern` under `mix help test` for more information.
```

PR #1279, opened 2026-04-23
perf (AliasUsage): Greatly improve performance when `:if_nested_deeper_than` excludes lots of issues

s3cur3:
This takes the runtime of this check from ~30 seconds to ~2.6 seconds on our codebase at Jump, where we've set the `:if_nested_deeper_than` setting to 10, meaning a _ton_ of calls were previously getting picked up in the initial walk of the codebase, then being filtered based on slow string splitting.


rrrene:
🥳 thx!

PR #1280, opened 2026-04-23
perf (UnusedOperation): Only walk the full AST once

s3cur3:
Prior to this, if UnusedOperation was given a list of custom modules like this:

```
{Credo.Check.Warning.UnusedOperation,
  modules: [
    {DateTime, :all},
    {Keyword, :all},
    {List, :all},
    {Map, :all},
    {MapSet, :all},
    {Path, :all}
  ]}
```

...it would walk each file's AST once per item in the `:modules` list. This could be quite slow.

With this change, we first collect _all_ unused calls to _all_ the specified modules in a single pass over the AST, then filter from there to get down to just the function calls the user is interested in.

At Jump, we had 39 such `:modules` configured, so the total runtime of the check was over 30 seconds. With this change, it's down to 1.5-2 seconds, which is right on par with the runtime for `UnusedEnumOperation` and `UnusedFileOperation` (each of which specify just a single module to search for). The runtime for the particular `Unused*Operation` checks appears unchanged in my testing.

It's *conceivable* that this would actually be a performance _pessimization_, if you had tons and tons of calls to a module in your `:modules` config, and wanted to only filter down to one or two such calls. However, I'm having trouble imagining a scenario where a module is so heavily used across the _entire_ codebase that the filtering would take more time on average than the AST walking.

### DIFF
--- a/lib/credo/check/warning/unused_function_return_helper.ex
+++ b/lib/credo/check/warning/unused_function_return_helper.ex
@@ -39,7 +39,7 @@ defmodule Credo.Check.Warning.UnusedFunctionReturnHelper do
          required_mod_list,
          nil
        ) do
-    if mods == required_mod_list do
+    if mods == required_mod_list or mods in required_mod_list do
       {ast, [ast | acc]}
     else
       {ast, acc}
@@ -52,7 +52,7 @@ defmodule Credo.Check.Warning.UnusedFunctionReturnHelper do
          required_mod_list,
          restrict_fun_names
        ) do
-    if mods == required_mod_list and fun_name in restrict_fun_names do
+    if (mods == required_mod_list or mods in required_mod_list) and fun_name in restrict_fun_names do
       {ast, [ast | acc]}
     else
       {ast, acc}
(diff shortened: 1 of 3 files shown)


lib/credo/check/warning/unused_function_return_helper.ex
```diff
@@ -39,7 +39,7 @@ defmodule Credo.Check.Warning.UnusedFunctionReturnHelper do
          required_mod_list,
          nil
        ) do
-    if mods == required_mod_list do
+    if mods == required_mod_list or mods in required_mod_list do
```
s3cur3, line 42: I considered ensuring that `required_mod_list` was always wrapped a list (since `UnusedOperation` is the only client in the codebase), which would allow us to remove the `mods == required_mod_list` case here. However, since this is a public module, I was concerned about potentially breaking clients. (The module _is_ marked `@moduledoc false`, so it's definitely a contract violation if someone _was_ using this, though.)

rrrene:
🥳 thx!

PR #1282, opened 2026-04-27
Fix `--all` command line flag

s3cur3:
Commit 2e53ab5a broke the `--all` option to `mix credo`, causing the truncation message, like:

```
┃  ...  (17 more, use `--all` to show them)
```

to be shown even when `--all` was supplied. With this change, `--all` is restored to listing all issues.


rrrene:
🥳 thx!

PR #1284, opened 2026-05-12
feat: add allow_lists and allow_maps params to single pipe check

flavioescobar:
Add a couple new params to `Credo.Check.Readability.SinglePipe`:
  * `:allow_lists` - Makes the check not raise issues on single pipes when the first element of the pipeline is a list literal.
  * `:allow_maps` - Makes the check not raise issues on single pipes when the first element of the pipeline is a map or struct.

### Context

This check is already flagged as "controversial" because it's more opinionated. 

We would love to enable single pipe checks at the place I work. But in our opinion, it's fine to have maps, structs or list literals on single pipelines. Especially when those are large. Here's an example:

```elixir
%{
  first_name: "John",
  last_name: "Doe",
  email: "johndoe@example.com",
  address: %{
    line1: "222333 Peachtree Place",
    city: "Atlanta",
    state: "GA",
    zip_code: "30318"
  }
}
|> create_person(authorize?: false)
```

We think the above looks better than this:

```elixir
create_person(
  %{
    first_name: "John",
    last_name: "Doe",
    email: "johndoe@example.com",
    address: %{
      line1: "222333 Peachtree Place",
      city: "Atlanta",
      state: "GA",
      zip_code: "30318"
    }
  },
  authorize?: false
)
```

But again -- understanding this is an opinionated and controversial check, we're proposing an additional param to this check, so that users can opt in or out of this exception. By default, the check behaves as usual -- no single pipes at all, except on macros.

I'd appreciate a review on this and I'd be happy to discuss it further.

### DIFF
--- a/lib/credo/check/readability/single_pipe.ex
+++ b/lib/credo/check/readability/single_pipe.ex
@@ -3,7 +3,12 @@ defmodule Credo.Check.Readability.SinglePipe do
     id: "EX3023",
     base_priority: :high,
     tags: [:controversial],
-    param_defaults: [allow_0_arity_functions: false, allow_blocks: true],
+    param_defaults: [
+      allow_0_arity_functions: false,
+      allow_blocks: true,
+      allow_lists: false,
+      allow_maps: false
+    ],
     explanations: [
       check: """
       Pipes (`|>`) should only be used when piping data through multiple calls.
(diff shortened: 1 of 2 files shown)


rrrene:
Hi, I like this very much :+1: 

One nitpick: Can we make it two params? One for maps/structs and one for lists?

flavioescobar:
> Hi, I like this very much 👍
> 
> One nitpick: Can we make it two params? One for maps/structs and one for lists?

I'm glad you like it! I'll change it to two params and let you know.

flavioescobar:
@rrrene I've updated the PR as requested. Could you please take another look at it?

Thanks again for taking the time!

rrrene:
@flavioescobar Thx!! 🎉

PR #1285, opened 2026-05-15
Fix MissingCheckInConfig missing unused checks (#1278)

### LINKED ISSUE #1278: MissingCheckInConfig doesn't find unused checks, as it relies on the checks being loaded
I tried the check from current main after this comment: https://github.com/rrrene/credo/issues/1277#issuecomment-4246623827 - locked on 2d116684cd4b16b505031a8d84c6ae8ee48617bd atm.

But I found it not to work. After enabling it here are the runs:

```
vscode ➜ /workspace/server (main) $ mix credo
Checking 222 source files (this might take a while) ...

  Software Design                                                                                         
┃ 
┃ [D] ↗ Check `Credo.Check.Warning.UnusedOperation` missing in config: enable or disable it explicitly.
┃       .credo.exs #(config)

Analysis took 0.3 seconds (0.1s to load, 0.2s running 59 checks on 222 files)
2023 mods/funs, found 1 software design suggestion.

Showing priority issues: ↑ ↗ →  (use `mix credo explain` to explain issues, `mix credo --help` for options).
vscode ➜ /workspace/server (main) $ mix credo
Checking 222 source files (this might take a while) ...

Analysis took 0.3 seconds (0.09s to load, 0.2s running 59 checks on 222 files)
2023 mods/funs, found no issues.
```

I.e. it only found `UnusedOperation` - after I fixed that none other were found.

To the best of my understanding this is because the check relies on `:code.all_loaded()` - I believe as the unused checks are never referenced they are never loaded and hence are never caught by this. It's one of these cases where I'm sure it works in the test suite of `credo` - as the modules are loaded here. It does find `UnusedOperation` as it's a helper check used in other checks and hence loaded (I'd also probably exclude it from being warned about, as the check does nothing without configuration as best as I understand it).

I'm not sure how to fix this other to make sure that the check modules are loaded, but at this point one might also just define a static list of known "good" checks that should be in a complete config (wouldn't work with plugins then, unless one made them implement an interface) - but maybe there are also other smarter ways to do this.

For completeness here is our `.credo.exs` - I added the checks that we were missing but the check didn't catch at the end with a comment above them.

<details>
<summary> .credo.exs </summary>

```elixir
# This file contains the configuration for Credo and you are probably reading
# this after creating it with `mix credo.gen.config`.
#
# If you find anything wrong or unclear in this file, please report an
# issue on GitHub: https://github.com/rrrene/credo/issues
#
%{
  #
  # You can have as many configs as you like in the `configs:` field.
  configs: [
    %{
      #
      # Run any config using `mix credo -C <name>`. If no config name is given
      # "default" is used.
      #
      name: "default",
      #
      # These are the files included in the analysis:
      files: %{
        #
        # You can give explicit globs or simply directories.
        # In the latter case `**/*.{ex,exs}` will be used.
        #
        included: [
          "lib/",
          "src/",
          "test/",
          "web/",
          "apps/*/lib/",
          "apps/*/src/",
          "apps/*/test/",
          "apps/*/web/"
        ],
        excluded: [~r"/_build/", ~r"/deps/", ~r"/node_modules/"]
      },
      #
      # Load and configure plugins here:
      #
      plugins: [],
      #
      # If you create your own checks, you must specify the source files for
      # them here, so they can be loaded by Credo before running the analysis.
      #
      requires: [],
      #
      # If you want to enforce a style guide and need a more traditional linting
      # experience, you can change `strict` to `true` below:
      #
      strict: false,
      #
      # To modify the timeout for parsing files, change this value:
      #
      parse_timeout: 5000,
      #
      # If you want to use uncolored output by default, you can change `color`
      # to `false` below:
      #
      color: true,
      #
      # You can customize the parameters of any check by adding a second element
      # to the tuple.
      #
      # To disable a check put `false` as second element:
      #
      #     {Credo.Check.Design.DuplicatedCode, false}
      #
      checks: %{
        enabled: [
          #
          ## Consistency Checks
          #
          {Credo.Check.Consistency.ExceptionNames, []},
          {Credo.Check.Consistency.LineEndings, []},
          {Credo.Check.Consistency.ParameterPatternMatching, []},
          {Credo.Check.Consistency.SpaceAroundOperators, []},
          {Credo.Check.Consistency.SpaceInParentheses, []},
          {Credo.Check.Consistency.TabsOrSpaces, []},

          #
          ## Design Checks
          #
          # You can customize the priority of any check
          # Priority values are: `low, normal, high, higher`
          #
          {Credo.Check.Design.AliasUsage,
           [priority: :low, if_nested_deeper_than: 2, if_called_more_often_than: 0]},
          {Credo.Check.Design.TagFIXME, []},
          # You can also customize the exit_status of each check.
          # If you don't want TODO comments to cause `mix credo` to fail, just
          # set this value to 0 (zero).
          #
          {Credo.Check.Design.TagTODO, [exit_status: 0]},

          #
          ## Readability Checks
          #
          {Credo.Check.Readability.AliasOrder, []},
          {Credo.Check.Readability.FunctionNames, []},
          {Credo.Check.Readability.LargeNumbers, []},
          {Credo.Check.Readability.MaxLineLength, [priority: :low, max_length: 120]},
          {Credo.Check.Readability.ModuleAttributeNames, []},
          {Credo.Check.Readability.ModuleDoc, []},
          {Credo.Check.Readability.ModuleNames, []},
          {Credo.Check.Readability.ParenthesesInCondition, []},
          {Credo.Check.Readability.ParenthesesOnZeroArityDefs, []},
          {Credo.Check.Readability.PipeIntoAnonymousFunctions, []},
          {Credo.Check.Readability.PredicateFunctionNames, []},
          {Credo.Check.Readability.PreferImplicitTry, []},
          {Credo.Check.Readability.RedundantBlankLines, []},
          {Credo.Check.Readability.Semicolons, []},
          {Credo.Check.Readability.SpaceAfterCommas, []},
          {Credo.Check.Readability.StringSigils, []},
          {Credo.Check.Readability.TrailingBlankLine, []},
          {Credo.Check.Readability.TrailingWhiteSpace, []},
          {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
          {Credo.Check.Readability.VariableNames, []},
          {Credo.Check.Readability.WithSingleClause, []},

          #
          ## Refactoring Opportunities
          #
          {Credo.Check.Refactor.Apply, []},
          {Credo.Check.Refactor.CondStatements, []},
          {Credo.Check.Refactor.CyclomaticComplexity, []},
          {Credo.Check.Refactor.FilterCount, []},
          {Credo.Check.Refactor.FilterFilter, []},
          {Credo.Check.Refactor.FunctionArity, []},
          {Credo.Check.Refactor.LongQuoteBlocks, []},
          {Credo.Check.Refactor.MapJoin, []},
          {Credo.Check.Refactor.MatchInCondition, []},
          {Credo.Check.Refactor.NegatedConditionsInUnless, []},
          {Credo.Check.Refactor.NegatedConditionsWithElse, []},
          {Credo.Check.Refactor.Nesting, []},
          {Credo.Check.Refactor.PassAsyncInTestCases, []},
          {Credo.Check.Refactor.RedundantWithClauseResult, []},
          {Credo.Check.Refactor.RejectReject, []},
          {Credo.Check.Refactor.UnlessWithElse, []},
          {Credo.Check.Refactor.UtcNowTruncate, []},
          {Credo.Check.Refactor.WithClauses, []},

          #
          ## Warnings
          #
          {Credo.Check.Warning.ApplicationConfigInModuleAttribute, []},
          {Credo.Check.Warning.BoolOperationOnSameValues, []},
          {Credo.Check.Warning.Dbg, []},
          {Credo.Check.Warning.ExpensiveEmptyEnumCheck, []},
          {Credo.Check.Warning.IExPry, []},
          {Credo.Check.Warning.IoInspect, []},
          {Credo.Check.Warning.MissedMetadataKeyInLoggerConfig, []},
          {Credo.Check.Warning.MixEnv, []},
          {Credo.Check.Warning.OperationOnSameValues, []},
          {Credo.Check.Warning.OperationWithConstantResult, []},
          {Credo.Check.Warning.RaiseInsideRescue, []},
          {Credo.Check.Warning.SpecWithStruct, []},
          {Credo.Check.Warning.UnsafeExec, []},
          {Credo.Check.Warning.UnusedEnumOperation, []},
          {Credo.Check.Warning.UnusedFileOperation, []},
          {Credo.Check.Warning.UnusedKeywordOperation, []},
          {Credo.Check.Warning.UnusedListOperation, []},
          {Credo.Check.Warning.UnusedPathOperation, []},
          {Credo.Check.Warning.UnusedRegexOperation, []},
          {Credo.Check.Warning.UnusedStringOperation, []},
          {Credo.Check.Warning.UnusedTupleOperation, []},
          {Credo.Check.Warning.WrongTestFileExtension, []},

          #
          ## Meta
          #
          # Check that all checks are actually configured!
          {Credo.Check.Design.MissingCheckInConfig, []}
        ],
        disabled: [
          #
          # Controversial and experimental checks (opt-in, just move the check to `:enabled`
          #   and be sure to use `mix credo --strict` to see low priority checks)
          #
          {Credo.Check.Consistency.MultiAliasImportRequireUse, []},
          {Credo.Check.Consistency.UnusedVariableNames, []},
          {Credo.Check.Design.DuplicatedCode, []},
          {Credo.Check.Design.SkipTestWithoutComment, []},
          {Credo.Check.Readability.AliasAs, []},
          {Credo.Check.Readability.BlockPipe, []},
          {Credo.Check.Readability.ImplTrue, []},
          {Credo.Check.Readability.MultiAlias, []},
          {Credo.Check.Readability.NestedFunctionCalls, []},
          {Credo.Check.Readability.OneArityFunctionInPipe, []},
          {Credo.Check.Readability.OnePipePerLine, []},
          {Credo.Check.Readability.SeparateAliasRequire, []},
          {Credo.Check.Readability.SingleFunctionToBlockPipe, []},
          {Credo.Check.Readability.SinglePipe, []},
          {Credo.Check.Readability.Specs, []},
          {Credo.Check.Readability.StrictModuleLayout, []},
          {Credo.Check.Readability.WithCustomTaggedTuple, []},
          {Credo.Check.Refactor.ABCSize, []},
          {Credo.Check.Refactor.AppendSingleItem, []},
          {Credo.Check.Refactor.DoubleBooleanNegation, []},
          {Credo.Check.Refactor.FilterReject, []},
          {Credo.Check.Refactor.IoPuts, []},
          {Credo.Check.Refactor.MapMap, []},
          {Credo.Check.Refactor.ModuleDependencies, []},
          {Credo.Check.Refactor.NegatedIsNil, []},
          {Credo.Check.Refactor.PipeChainStart, []},
          {Credo.Check.Refactor.RejectFilter, []},
          {Credo.Check.Refactor.VariableRebinding, []},
          {Credo.Check.Warning.LazyLogging, []},
          {Credo.Check.Warning.LeakyEnvironment, []},
          {Credo.Check.Warning.MapGetUnsafePass, []},
          {Credo.Check.Warning.UnsafeToAtom, []},

          # This check was found by MissingCheckInConfig because it happens to be loaded
          # as a shared helper by other enabled checks (e.g. UnusedEnumOperation)
          {Credo.Check.Warning.UnusedOperation, []},

          #
          # Checksnot found by `MissingCheckInConfig` despite being
          # present in this version of Credo. They are missed because all_loaded_checks/0
          # relies on :code.all_loaded/0, which only returns modules already loaded into
          # the VM — checks absent from the config are never loaded, so they are invisible.
          #
          {Credo.Check.Readability.PreferUnquotedAtoms, []},
          {Credo.Check.Refactor.CaseTrivialMatches, []},
          {Credo.Check.Refactor.CondInsteadOfIfElse, []},
          {Credo.Check.Refactor.MapInto, []},
          {Credo.Check.Refactor.PerceivedComplexity, []},
          {Credo.Check.Warning.StructFieldAmount, []},
          {Credo.Check.Warning.UnusedMapOperation, []},
          {Credo.Check.Warning.WrongTestFilename, []}

          #
          # Custom checks can be created using `mix credo.gen.check`.
          #
        ]
      }
    }
  ]
}
```

</details>

Thanks as usual for all your work :)

dl-alexandre:
Closes #1278.

`Credo.Check.all_loaded_checks/0` was sourcing its result via `Credo.Code.loaded_modules_implementing/1`, which iterates `:code.all_loaded/0`. Because nothing references an unused check module, the BEAM never loads it — so it never appeared in the result, and `MissingCheckInConfig` silently failed to flag it. The bug is hidden in credo's own test suite, where every check module gets loaded incidentally; the reporter (@PragTob) diagnosed this correctly in the issue.

This sources credo's own checks from the curated `standard_checks/0` list instead. Two side benefits beyond fixing the original bug:

1. **`UnusedOperation` is no longer flagged as missing.** It's not in `standard_checks/0` (it's a generic check that does nothing without `modules: [...]` configuration), so it falls out naturally. That matches @rrrene's note on the issue ("I'd also probably exclude it from being warned about, as the check does nothing without configuration").

2. **Plugin checks are now discovered without depending on load order.** A new private `external_check_modules/0` walks `Application.loaded_applications/0` and reads each app's module list from `:application.get_key/2`, then filters to modules implementing the `Credo.Check` behaviour. That doesn't depend on `:code.all_loaded/0` either, so plugin checks show up whether or not anything has referenced them yet.

The only caller of `Credo.Check.all_loaded_checks/0` in `lib/` is `MissingCheckInConfig`, so the contract change is well-scoped.

Added a small unit test asserting that `all_loaded_checks/0` returns a superset of `standard_checks/0` regardless of load order, and that the result has no duplicates. CI will verify the full matrix.

### DIFF
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -914,17 +914,35 @@ defmodule Credo.Check do
         app != :credo,
         {:ok, modules} <- [:application.get_key(app, :modules)],
         module <- modules,
-        Code.ensure_loaded?(module),
         implements_credo_check?(module),
         do: module
   end
 
   defp implements_credo_check?(module) do
-    function_exported?(module, :module_info, 1) and
-      module.module_info(:attributes)
-      |> Keyword.get_values(:behaviour)
-      |> List.flatten()
-      |> Enum.member?(__MODULE__)
+    case beam_attributes(module) do
+      {:ok, attributes} ->
+        attributes
+        |> Keyword.get_values(:behaviour)
+        |> List.flatten()
+        |> Enum.member?(__MODULE__)
+
+      :error ->
+        false
+    end
+  end
+
+  defp beam_attributes(module) do
+    with path when is_list(path) <- :code.which(module),
+         {:ok, {^module, [attributes: attributes]}} <- :beam_lib.chunks(path, [:attributes]) do
+      {:ok, attributes}
+    else
+      _ ->
+        if function_exported?(module, :module_info, 1) do
+          {:ok, module.module_info(:attributes)}
+        else
+          :error
+        end
+    end
   end
 
   @doc false
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -914,35 +914,17 @@ defmodule Credo.Check do
         app != :credo,
         {:ok, modules} <- [:application.get_key(app, :modules)],
         module <- modules,
+        Code.ensure_loaded?(module),
         implements_credo_check?(module),
         do: module
   end
 
   defp implements_credo_check?(module) do
-    case beam_attributes(module) do
-      {:ok, attributes} ->
-        attributes
-        |> Keyword.get_values(:behaviour)
-        |> List.flatten()
-        |> Enum.member?(__MODULE__)
-
-      :error ->
-        false
-    end
-  end
-
-  defp beam_attributes(module) do
-    with path when is_list(path) <- :code.which(module),
-         {:ok, {^module, [attributes: attributes]}} <- :beam_lib.chunks(path, [:attributes]) do
-      {:ok, attributes}
-    else
-      _ ->
-        if function_exported?(module, :module_info, 1) do
-          {:ok, module.module_info(:attributes)}
-        else
-          :error
-        end
-    end
+    function_exported?(module, :module_info, 1) and
+      module.module_info(:attributes)
+      |> Keyword.get_values(:behaviour)
+      |> List.flatten()
+      |> Enum.member?(__MODULE__)
   end
 
   @doc false
--- a/lib/credo/check.ex
+++ b/lib/credo/check.ex
@@ -896,6 +896,38 @@ defmodule Credo.Check do
   end
 
   @doc false
+  def all_loaded_checks do
+    # Source credo's own checks from the curated `standard_checks/0` list
+    # rather than `:code.all_loaded/0` — until something references an
+    # unused check module the BEAM hasn't loaded it, so the all-loaded set
+    # silently misses checks that nothing has yet pulled in. That's the
+    # root of `MissingCheckInConfig` failing to flag unused checks (#1278).
+    #
+    # We supplement with check modules from any *other* loaded application
+    # (plugins, custom check libraries) by reading their module list from
+    # `:application.get_key/2`, which doesn't depend on prior code loading.
+    Enum.uniq(standard_checks() ++ external_check_modules())
+  end
+
+  defp external_check_modules do
+    for {app, _, _} <- Application.loaded_applications(),
+        app != :credo,
+        {:ok, modules} <- [:application.get_key(app, :modules)],
+        module <- modules,
+        Code.ensure_loaded?(module),
+        implements_credo_check?(module),
+        do: module
+  end
+
+  defp implements_credo_check?(module) do
+    function_exported?(module, :module_info, 1) and
+      module.module_info(:attributes)
+      |> Keyword.get_values(:behaviour)
+      |> List.flatten()
+      |> Enum.member?(__MODULE__)
+  end
+
+  @doc false
   # TODO: decide whether this should live here or in `Credo.Execution`.
   def mentioned_checks(exec) do
     exec
--- a/test/credo/check_all_loaded_checks_test.exs
+++ b/test/credo/check_all_loaded_checks_test.exs
@@ -1,0 +1,25 @@
+defmodule Credo.Check.AllLoadedChecksTest do
+  use ExUnit.Case, async: true
+
+  # `all_loaded_checks/0` used to source its result from `:code.all_loaded/0`,
+  # so any check module nothing had yet referenced was silently missing —
+  # which is exactly what made `MissingCheckInConfig` fail to flag unused
+  # checks (#1278). The fix sources credo's own checks from the curated
+  # `standard_checks/0` list and supplements with check modules discovered
+  # via `:application.get_key/2` on every other loaded application, so the
+  # result is independent of which modules happen to be loaded.
+
+  test "returns every standard credo check regardless of load order" do
+    all = Credo.Check.all_loaded_checks()
+    missing = Credo.Check.standard_checks() -- all
+
+    assert missing == [],
+           "expected all standard checks to be included in `all_loaded_checks/0`, " <>
+             "but these were missing: #{inspect(missing)}"
+  end
+
+  test "result contains no duplicates" do
+    all = Credo.Check.all_loaded_checks()
+    assert all == Enum.uniq(all)
+  end
+end
--- a/test/credo/check_all_loaded_checks_test.exs
+++ b/test/credo/check_all_loaded_checks_test.exs
@@ -1,5 +1,5 @@
 defmodule Credo.Check.AllLoadedChecksTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
 
   # `all_loaded_checks/0` used to source its result from `:code.all_loaded/0`,
   # so any check module nothing had yet referenced was silently missing —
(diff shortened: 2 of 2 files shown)


rrrene:
Hi, thx for trying to solve this.

But isn't this approach (going `apps -> modules -> Code.ensure_loaded?(module)`)  just `:code.all_loaded/0` with extra steps?

dl-alexandre:
Hi René, thanks for the question.

The key distinction is the *source* of the candidate modules, not the loading step.

`:code.all_loaded/0` returns only modules already resident in the VM. Because unused checks are never referenced, the BEAM never loads them — so they are absent from the result, which is exactly why `MissingCheckInConfig` missed them (#1278).

`:application.get_key(app, :modules)` reads the `.app` manifest of each loaded application, which declares every module the app ships with, regardless of whether anything has loaded it yet. `Code.ensure_loaded?/1` is then used only as a filter to verify the module exists and implements `Credo.Check`.

For Credo's own checks we skip dynamic discovery entirely and use the static `standard_checks/0` list from `.credo.exs`. The manifest walk is solely for discovering plugin / external checks.

If you'd prefer to avoid any loading side-effects for plugins, we could switch that path to `:beam_lib.chunks/2` to inspect the `:behaviour` attribute directly from the `.beam` file on disk.

Does that address the concern?

rrrene:
We can't load the chunks because we can not assume the project is compiled. 

Could you explain in your own words what the new test from your last commit actually tests?

dl-alexandre:
You're right, the `:beam_lib.chunks/2` follow-up was the wrong direction because it assumes compiled BEAM files are available. I reverted that commit.

The test I added there only proved that a compiled module listed in an `.app` manifest could be discovered without loading it. That does not cover the uncompiled project case Credo has to support, so it was not good evidence for this PR.

I think the core fix here is still the static `standard_checks/0` path for Credo's own checks, which addresses #1278 without depending on load order. The external/plugin discovery part is the questionable bit, and I'm happy to scope that down if you'd prefer this PR to only cover Credo's built-in checks.

rrrene:
Sorry for not checking in sooner.

Didn't you just re-iterate something you said before? What's the meaningful difference?

dl-alexandre:
Hello!

After a couple reviews, I think this change is obsolete. It was suggested to throw in a regression test but that could be more noise.. closing

PR #1292, opened 2026-06-03
Enable UnsafeToAtom by default and frame it as a security check

joshprice:
Inspired by the excellent EEF security write up https://erlef.org/blog/security/atom-exhaustion I wondered why this was disabled by default.

I've moved `Credo.Check.Warning.UnsafeToAtom` from the disabled (opt-in) section of `.credo.exs` into the enabled section, drop its `:controversial` tag, and rewrite the explanation to explain the security risk.

Converting unknown or external input into atoms is a security risk. Atoms are never garbage-collected and a node's atom table is capped (1_048_576 by default), so any code path that turns attacker-controlled input into atoms can exhaust the table and crash the whole VM. That is a denial-of-service vulnerability, and in a web-facing application it is a potential CVE.

Leaving this off by default and tagging it `:controversial` sends the wrong message for a security check. The explanation now links to the EEF Security WG write-up.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -157,6 +157,7 @@
           {Credo.Check.Warning.SpecWithStruct, []},
           {Credo.Check.Warning.StructFieldAmount, []},
           {Credo.Check.Warning.UnsafeExec, []},
+          {Credo.Check.Warning.UnsafeToAtom, []},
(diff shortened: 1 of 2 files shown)


rrrene:
I think the reasoning is :+1:

PR #1293, opened 2026-06-03
chore: Clean stale deps

nedimtokic:
## Summary

Removes unused dependencies from `mix.lock` by running `mix deps.clean --unlock --unused`.

## Motivation

`hackney` was present in `mix.lock` despite not being part of the active dependency tree. Since `hackney` has a known security vulnerability, retaining it in the lock file poses an unnecessary risk — even unused locked dependencies can be picked up by tools that scan `mix.lock` for vulnerable packages.

## Changes

- Ran `mix deps.clean --unlock --unused` to remove all stale entries from `mix.lock` that are no longer referenced in the dependency tree.


rrrene:
@nedimtokic Thx! 🎉

PR #1295, opened 2026-06-17
Ensure we aren't adding extra newlines to failure output

synic:
Use `Bunt.write` instead of `Bunt.puts` for failure text output so we don't end up with extra newlines

Before:

image

After:

image

### DIFF
--- a/lib/credo/cli/command/diff/output/default.ex
+++ b/lib/credo/cli/command/diff/output/default.ex
@@ -328,18 +328,10 @@ defmodule Credo.CLI.Command.Diff.Output.Default do
       " ",
       first_line
     ]
-    |> UI.puts()
+    |> UI.write_line()
 
     other_lines
     |> Enum.each(&print_issue_message(&1, issue, outer_color, message_color))
-  end
-
-  defp print_issue_message(
-         "",
-         _issue,
-         _outer_color,
-         _message_color
-       ) do
   end
 
   defp print_issue_message(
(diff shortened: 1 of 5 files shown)


rrrene:
Hi, would be helpful if you could state why we need this?

Do you want to use multiline issue messages?

synic:
@rrrene yes, we use multiline for a bunch of them (like you see in the example). Without the fix, every line is rewritten with an extra newline due to `Bunt.puts` adding it's own newline.

rrrene:
Got it. I am not necessarily against allowing multi-line issue messages, it just did not come up so far.

Out of curiosity: why would you want to print a multi-paragraph explanation as a message, when there is also the `:explanations` one can define for a check?

rrrene:
Checking in. Any updates on this?

s3cur3:
The example in the screenshot comes from our fork of [ExcellentMigrations](https://github.com/Artur-Sulej/excellent_migrations). That single Credo check (`ExcellentMigrations.CredoCheck.MigrationsSafety`) checks for a variety of issues with Ecto migrations, but the explanation we provide is specific to each particular issue. (For instance, the explanation for a warning about removing a column is different from a warning about changing a column type.) We _could_ split this up into 20+ separate checks, but at the cost of running 20x slower.

rrrene:
Makes sense. Could you check if your issue persists on `master`? 

1ad66e2a should have fixed this w/o introducing a new write function to the `UI` module.

s3cur3:
Yep, I confirm it works without @synic's change on the tip of `master`! 🙏

PR #1296, opened 2026-06-18
Add PreferDateTimeShift check

### LINKED ISSUE #1291: Suggestion: Prefer DateTime/NaiveDateTime/Date/Time shift over add
👋 

Before PRing the check (already implemented it on our code base) I wanted to check if people would even want it:

All of the aforementioned modules have a notice on their `add/4` variants to prefer `shift/4`, i.e [for DateTime](https://hexdocs.pm/elixir/DateTime.html#add/4):

> Prefer [shift/2](https://hexdocs.pm/elixir/DateTime.html#shift/2)
> Prefer [shift/2](https://hexdocs.pm/elixir/DateTime.html#shift/2) over [add/3](https://hexdocs.pm/elixir/DateTime.html#add/3), as it offers a more ergonomic API.
>
> [add/3](https://hexdocs.pm/elixir/DateTime.html#add/3) provides a lower-level API which only supports fixed units such as :hour and :second, but not :month (as the exact length of a month depends on the current month). [add/3](https://hexdocs.pm/elixir/DateTime.html#add/3) always considers the unit to be computed according to the [Calendar.ISO](https://hexdocs.pm/elixir/Calendar.ISO.html).

So a simple check that checks these was helpful imo (as it cut down on my PR comments).

Thanks as always for all your work!

PragTob:
👋 

Fixes #1291 - explanation in there and in the check, but basically docs say to prefer it so we prefer it!

This is the first check I've ever PR'ed to credo (I believe 😅 ) so happy for feedback. Put priority to low and disabled it by default, as I've never met anyone else who cared about this but would also be happy to up it. 😁

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -209,6 +209,7 @@
           {Credo.Check.Refactor.NegatedIsNil, []},
           {Credo.Check.Refactor.PassAsyncInTestCases, []},
           {Credo.Check.Refactor.PipeChainStart, []},
+          {Credo.Check.Refactor.PreferDateTimeShift, []},
           {Credo.Check.Refactor.RejectFilter, []},
           {Credo.Check.Refactor.VariableRebinding, []},
           {Credo.Check.Warning.LazyLogging, []},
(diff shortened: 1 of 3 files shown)


PragTob:
The 2 CI failures are optional/non applicable as best as I can tell :)

rrrene:
@PragTob Thx!! :tada:

PR #1304, opened 2026-08-04
fix(alias_usage): flag unaliased struct literals

### LINKED ISSUE #1303: `Credo.Check.Design.AliasUsage` doesn't warn on un-aliased struct modules
Given the following code:

```elixir
defmodule Foo do
  def run do
    %Foo.Bar.Baz{..}
  end
end
```

I'd expect `Credo.Check.Design.AliasUsage` to warn on `Foo.Bar.Baz`, as it could be aliased:

```elixir
defmodule Foo do
  alias Foo.Bar.Baz

  def run do
    %Baz{..}
  end
end
```

I don't mind whether this is configurable or enabled by default, but off the top of my head I can't think of any places where I'd prefer the un-aliased form.

superhawk610:
`%Foo.Bar.Baz{}` was silently ignored because the `:%` AST node didn't match any `find_issues` clause. Add a clause for struct literals so they're treated as module references.

Closes #1303.

### DIFF
--- a/lib/credo/check/design/alias_usage.ex
+++ b/lib/credo/check/design/alias_usage.ex
@@ -119,6 +119,12 @@ defmodule Credo.Check.Design.AliasUsage do
     do_find_issues(ast, mod_list, meta, ctx)
   end
 
+  # Struct literals: %Foo.Bar.Baz{} — the module is a reference, not a function call
+  defp find_issues({:%, _, [{:__aliases__, meta, mod_list}, _]} = ast, ctx)
+       when is_list(mod_list) do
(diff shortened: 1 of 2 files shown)


rrrene:
Thx for this. You are right that is missing 🙏  

Have you thought about adding a param for this? To give people the option to de-activate this "new" behaviour?

superhawk610:
@rrrene great idea! I added `include_struct_literals`, enabled by default.

rrrene:
@superhawk610 Thx for this! 🎉

PR #1306, opened 2026-08-10
Add NestedWith check

easink:
Reports a `with` that does not start at the top of a function body: one buried inside an `if`, `unless`, `case`, `cond`, `for`, `receive`, or another `with`. Such a `with` splits a chain of pattern matches across two levels of control flow, which hides its failure paths and makes the `else` of either block ambiguous to a reader.

`def`, `defp`, `defmacro`, `defmacrop`, `fn` and `defmodule` open a scope, so a `with` at the top of one of their bodies is never reported, however deeply the scope itself is nested. `defmodule` counts because a module is routinely wrapped in a compile-time `if
Code.ensure_loaded?(Dep)`, which is conditional compilation rather than control flow around the `with`. `try` is neither scope nor block and is passed through: wrapping a chain in `try/rescue` does not branch it, and a `try` has no `else` to confuse with the one belonging to the `with`.

The traversal is an explicit recursion rather than `Credo.Code.prewalk/3` because the flag it carries has to be reset per branch, and reset entirely on the way into a scope.

Opt-in for now, so it lands in `:disabled`.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -207,6 +207,7 @@
           {Credo.Check.Refactor.MapMap, []},
           {Credo.Check.Refactor.ModuleDependencies, []},
           {Credo.Check.Refactor.NegatedIsNil, []},
+          {Credo.Check.Refactor.NestedWith, []},
           {Credo.Check.Refactor.PassAsyncInTestCases, []},
           {Credo.Check.Refactor.PipeChainStart, []},
           {Credo.Check.Refactor.PreferDateTimeShift, []},
--- a/lib/credo/check/refactor/nested_with.ex
+++ b/lib/credo/check/refactor/nested_with.ex
@@ -1,0 +1,158 @@
+defmodule Credo.Check.Refactor.NestedWith do
+  use Credo.Check,
+    id: "EX4035",
+    base_priority: :high,
+    explanations: [
+      check: ~S"""
+      A `with` should start at the top of a function body. One buried inside another
+      block - an `if`, a `case`, or another `with` - splits a chain of pattern matches
+      across two levels of control flow, which hides its failure paths and makes the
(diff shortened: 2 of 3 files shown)


rrrene:
Hi, thx for putting this together. Unfortunately, I do not see this as a standard check in its current form. 

Could you describe why this is a meaningful refactoring beyond a readability preference for "do never nest `with`"?

easink:
It's that "with" is supposed to give a chain of matches exactly one failure-handling. Nesting it inside "if" / "unless" / "case" / "cond" / another "with" adds a second, independent one, so figuring out how this can fail means finding two else's and working out which failures each owns.

It's also not just taste, the outer condition almost always hoists cleanly into a leading "with" clause, collapsing two "else"s. 

There's precedent for this in Credo itself: "Refactor.WithClauses" is already a standard, high-priority check that polices "with"s shape for a related reason (keeping the chain legible), and the project's own style guide already discourages nesting "if" / "unless" / "case" more than once.

Happy to give some more examples if that would help make the case.

rrrene:
I think there is value in providing teams a check that allows them to say:

- "We don't want ifs nested in case statements anymore"
- "We don't want with nested in with"
- "We don't want ifs nested in ifs"

and so on.

It could be a general check that has your use-case as its default config:

```elixir
{Credo.Check.Refactor.NestedSomething, disallow_nested_in: %{
        with: %{
          any_of: [:if, :unless, :case, :cond, :for, :receive, :with]
        }
      }}
```

I understand that you might be invested in this being a `with` specific check. 

I am just saying this might benefit more people and teams if it offers more flexibility 😄

easink:
I like the generic approach and renamed it to ForbiddenNesting.

PR #1311, opened 2026-08-29
fix: avoid arbitrary consistency warnings on tied conventions

### LINKED ISSUE #1174: Incorrect `Exception modules should be named consistently`
### Environment

* Credo version (`mix credo -v`): 1.7.11
* Erlang/Elixir version (`elixir -v`): 1.18.0
* Operating system: Mac 15.1.1

```
defmodule NervesHubWeb.NotFoundError do
  defexception message: "not found", plug_status: 404
end

defmodule NervesHubWeb.UnauthorizedError do
  defexception message: "forbidden", plug_status: 401
end
```

### Expected outcome

No warnings.

### Actual outcome

```
  Consistency
┃
┃ [C] ↗ Exception modules should be named consistently. It seems your strategy is to prefix them with `Unauthorized`, but
┃       `NervesHubWeb.NotFoundError` does not follow that convention.
┃       lib/nerves_hub_web/errors.ex:5:11 #(NervesHubWeb.NotFoundError)
```

mvanhorn:
Change the shared collector's convention inference so an unforced analysis produces issues only when exactly one match has the highest frequency; tied leaders represent an ambiguous convention and should yield no issues. Preserve the existing `force` path so checks with an explicitly configured convention continue reporting deviations even when observed counts are tied or uniform. `Credo.Check.Consistency.ExceptionNames` reports `NervesHubWeb.NotFoundError` as inconsistent when the sample contains two names sharing the `Unauthorized` prefix and two names sharing the `Error` suffix. The shared collector currently sorts frequency entries and chooses one entry with `Enum.max_by/2`, so equally prevalent conventions are resolved by term ordering rather than by evidence of a dominant project style.

A shared-collector input with one strictly most-frequent match still formats issues for source files containing less-frequent alternatives; A shared-collector input whose highest frequency is shared by two matches produces no issues instead of selecting a convention by term ordering.

Fixes #1174

### DIFF
--- a/lib/credo/check/consistency/collector.ex
+++ b/lib/credo/check/consistency/collector.ex
@@ -183,13 +183,14 @@ defmodule Credo.Check.Consistency.Collector do
   end
 
   defp most_frequent_match(frequencies, supress_issues_for_single_match?, nil) do
-    {value, frequency_of_match} = frequencies |> Enum.sort() |> Enum.max_by(&elem(&1, 1))
+    {value, frequency_of_match} = Enum.max_by(frequencies, &elem(&1, 1))
     single_match? = frequency_of_match == 1
+    tied_match? = Enum.count(frequencies, &(elem(&1, 1) == frequency_of_match)) > 1
 
-    if single_match? && supress_issues_for_single_match? do
-      :__only_single_match__
-    else
-      value
+    cond do
+      tied_match? -> :__no_most_frequent_match__
+      single_match? && supress_issues_for_single_match? -> :__only_single_match__
+      true -> value
     end
   end
 
@@ -202,6 +203,10 @@ defmodule Credo.Check.Consistency.Collector do
   end
 
   defp source_files_with_issues(_frequencies_per_file, :__only_single_match__) do
+    []
+  end
+
+  defp source_files_with_issues(_frequencies_per_file, :__no_most_frequent_match__) do
     []
   end
 
--- a/test/credo/check/consistency/collector_test.exs
+++ b/test/credo/check/consistency/collector_test.exs
@@ -1,0 +1,46 @@
+defmodule Credo.Check.Consistency.CollectorTest do
+  use Credo.Test.Case
+
+  alias Credo.Check.Consistency.Collector
+  alias Credo.SourceFile
+
+  defmodule TestCollector do
+    def collect_matches(source_file, _params) do
+      case SourceFile.source(source_file) do
+        ":mostly_foo" -> %{foo: 2}
+        ":only_bar" -> %{bar: 1}
+        ":mostly_bar" -> %{bar: 2}
+      end
+    end
+  end
+
+  test "reports files containing matches other than the unique most frequent match" do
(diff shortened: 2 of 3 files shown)


rrrene:
Thx! I like that the analysis now also yields a `:no_most_frequent_thing` as a result. 

But how is this useful in the context of this project in a practical sense?

With this change, if half my lines are indented using tabs and the other half using spaces, I won't notice.

mvanhorn:
Fair point, and the tabs-vs-spaces case shows where this lands wrong. The change fixes #1174, where a tie was resolved by term ordering and the check then reported files as deviating from a convention nobody actually chose. But going silent on a tie throws away the other half of the signal: a project split down the middle has no convention, and that is worth telling the user. What I'd propose is a single project-level issue for the tie case, something like 'no dominant convention found for X, seen N of A and N of B', instead of the per-file 'expected A' issues that pointed at an arbitrary side. That keeps the tie visible without asserting a winner. The force path is unchanged either way. If you'd rather keep the current per-file behavior and only fix the ordering, I can do that instead. Which shape do you prefer?

rrrene:
> What I'd propose is a single project-level issue for the tie case

There is no concept like that in this project. Are you proposing to add this? How would you go about it? Can you answer this for each affected check/file in this repo sequentially? And could you elaborate on why

> The force path is unchanged either way.

is true in those cases?

It is really important to talk about this before hand, I think :+1:

mvanhorn:
Fair questions, and taking them in order.

**Am I proposing to add it?** No — I withdraw that. I floated it before checking what it would cost, and you are right that the concept does not exist. Every consistency issue is built through `format_issue(issue_meta, ...)`, and `issue_meta` is bound to a source file, so a project-level issue needs either a synthetic location or a new issue kind that the CLI, the JSON output and every consumer would have to learn. That is a much larger change than #1174 justifies, and not one I should smuggle in through a bug fix.

**Per affected check.** There is only one path to describe, which is the part that makes this tractable. All eight consistency checks reach the same `most_frequent_match/3` clause in `Collector`. None of them decides ties for itself, so whatever the tie rule is, it is uniform across `exception_names`, `line_endings`, `multi_alias_import_require_use`, `parameter_pattern_matching`, `space_around_operators`, `space_in_parentheses`, `tabs_or_spaces` and `unused_variable_names`. The only per-check variation is `supress_issues_for_single_match?`, passed `true` by `exception_names` and `space_in_parentheses`, and that is orthogonal to the tie.

**Why the force path is unchanged.** It is a different function clause. My tie handling lives in `most_frequent_match(frequencies, supress, nil)`, which matches only when `params[:force]` is nil. When `:force` is set, `most_frequent_match(_frequencies, _supress, forced_value)` matches instead and returns the forced value — the tie branch is unreachable by pattern, not by a runtime guard I added.

So that leaves the narrow choice you implied: keep the per-file issues and only make the tie deterministic instead of term-ordered, or stay silent on a tie as the PR currently does. I would rather you pick, since it is your call what a tie should mean.

PR #1314, opened 2026-09-14
perf (tokenization): Speed up token-based checks by caching file tokenization

s3cur3:
Many checks might need to tokenize the same file (`SpaceAroundOperators`, `ParenthesesInCondition`, `PreferUnquotedAtoms`, etc.). Previously, each check did the tokenization of each file itself, resulting in duplicate work.

This adds two new ETS-backed caches: one for the Credo tokens (the method produced by `CredoTokenizer`), and another for the Elixir tokens (via `:elixir_tokenizer`). This allows us to pay just once per method per file for the tokenization.

With just the Consistency and Readability checks enabled on our ~750k LoC codebase at Jump:

Before:
> Analysis took 20.3 seconds (3.4s to load, 16.8s running 29 checks on 11832 files)

After:
> Analysis took 16.7 seconds (3.0s to load, 13.7s running 29 checks on 11832 files)


lib/credo/service/ets_table_helper.ex
```diff
@@ -28,27 +48,25 @@ defmodule Credo.Service.ETSTableHelper do
       end
 
       def put(source_file, value) do
-        GenServer.call(__MODULE__, {:put, source_file.hash, value}, @timeout)
+        true = :ets.insert(@table_name, {source_file.hash, value})
```
s3cur3, line 51: This is a change in behavior compared to the previous implementation, but I couldn't find any uses of the ETSTableHelper where it seemed like only inserting from the GenServer process mattered for correctness. (In the worst case, two processes concurrently insert the same data for the source file, so it doesn't matter which write wins.) 

The motivation here is simply performance: when serializing the writes behind the single GenServer, _all_ concurrently-checked files had to wait on each other. In my tests, that led to each file's tokenization taking about 98 ms, compared to around 60 ms to just run the tokenization without any caching at all. By allowing any process to write to the ETS table, we drop down to about 30 ms per file.

lib/credo/check.ex
```diff
@@ -800,39 +800,21 @@ defmodule Credo.Check do
...
-      case SourceFileScopePriorities.get(source_file) do
-        {:ok, value} ->
-          value
-
-        :notfound ->
-          result = Priority.scope_priorities(source_file)
-          SourceFileScopePriorities.put(source_file, result)
-          result
-      end
+    scope_prio_map = SourceFileScopePriorities.get_or_compute(source_file, &Priority.scope_priorities/1)
```
s3cur3, line 817: Unrelated to the performance changes here, but since I was introducing `get_or_compute/2` anyway, I figured I should update all the callsites that were previously doing the same thing.

rrrene:
@s3cur3 Thx for this!

I am wondering if we really need GenServers for this at all, especially with the concurrent writes. 

This implementation is almost 10 years old and I honestly never critically reflected on the GenServer based reads either ... maybe they are bottle-necking us as well ...

